PR #22, opened 2016-10-24
installer/windows: link to downloads page

tessr:
Per a suggestion from @danrobinson, point folks to our downloads page.


tessr:
PTAL

PR #24, opened 2016-10-24
readme: add download link

danrobinson:
Adds a download link similar to the one suggested in #22, but in the top-level readme, which right now only describes how to build the project from source.


danrobinson:
PTAL

PR #26, opened 2016-10-24
protocol/validation: add blockchain and asset ID checking to issuance validation

bobg:

### DIFF
--- a/core/asset/asset.go
+++ b/core/asset/asset.go
@@ -26,21 +26,19 @@ import (
 
 const maxAssetCache = 100
 
-func NewRegistry(db pg.DB, chain *protocol.Chain, initialBlockHash bc.Hash) *Registry {
+func NewRegistry(db pg.DB, chain *protocol.Chain) *Registry {
 	return &Registry{
-		db:               db,
-		chain:            chain,
-		initialBlockHash: initialBlockHash,
-		cache:            lru.New(maxAssetCache),
+		db:    db,
+		chain: chain,
+		cache: lru.New(maxAssetCache),
 	}
 }
 
 // Registry tracks and stores all known assets on a blockchain.
 type Registry struct {
-	db               pg.DB
-	chain            *protocol.Chain
-	indexer          Saver
-	initialBlockHash bc.Hash
+	db      pg.DB
+	chain   *protocol.Chain
+	indexer Saver
 
 	cacheMu sync.Mutex
 	cache   *lru.Cache
@@ -87,8 +85,8 @@ func (reg *Registry) Define(ctx context.Context, xpubs []string, quorum int, def
 	asset := &Asset{
 		Definition:       definition,
 		IssuanceProgram:  issuanceProgram,
-		InitialBlockHash: reg.initialBlockHash,
-		AssetID:          bc.ComputeAssetID(issuanceProgram, reg.initialBlockHash, 1),
+		InitialBlockHash: reg.chain.InitialBlockHash,
+		AssetID:          bc.ComputeAssetID(issuanceProgram, reg.chain.InitialBlockHash, 1),
 		Signer:           assetSigner,
 		Tags:             tags,
 	}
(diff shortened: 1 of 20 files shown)


protocol/validation/block.go
```diff
@@ -32,7 +32,7 @@ var (
 // See $CHAIN/protocol/doc/spec/validation.md#accept-block.
 // It evaluates the prevBlock's consensus program,
 // then calls ValidateBlock.
-func ValidateBlockForAccept(ctx context.Context, snapshot *state.Snapshot, prevBlock, block *bc.Block, validateTx func(*bc.Tx) error) error {
+func ValidateBlockForAccept(ctx context.Context, snapshot *state.Snapshot, prevBlock, block *bc.Block, initialBlockHash bc.Hash, validateTx func(*bc.Tx) error) error {
```
kr: What do you think about putting the initial block hash in state.Snapshot? (Not in the serialized snapshot data, just the internal representation.)

I don't know how that would shake out exactly, but it might be a not-entirely-unreasonable place for it, since the state is tied to the initial block hash anyway.

bobg:
PTAL

protocol/protocol.go
```diff
@@ -143,7 +145,15 @@ func NewChain(ctx context.Context, store Store, pool Pool, heights <-chan uint64
...
-	// Note that c.height.n may still be zero here.
+	if c.state.height >= 1 {
+		b, err := store.GetBlock(ctx, 1)
+		if err != nil {
+			return nil, errors.Wrap(err, "retrieving initial block")
+		}
+		c.InitialBlockHash = b.Hash()
+	}
+
+	// Note that c.state.height may still be zero here.
```
jbowens: If `c.state.height` is 0, will `c.InitialBlockHash` get set on the first block committed?

protocol/validation/block.go
jbowens: Maybe if we put the initial block hash in the `state.Snapshot`, we can avoid putting it on the `protocol.Chain` (and changing the interface of `asset.NewRegistry`)?

core/asset/asset.go
```diff
@@ -87,8 +85,8 @@ func (reg *Registry) Define(ctx context.Context, xpubs []string, quorum int, def
 	asset := &Asset{
 		Definition:       definition,
 		IssuanceProgram:  issuanceProgram,
-		InitialBlockHash: reg.initialBlockHash,
-		AssetID:          bc.ComputeAssetID(issuanceProgram, reg.initialBlockHash, 1),
+		InitialBlockHash: reg.chain.InitialBlockHash,
```
jbowens: If we read `reg.chain.InitialBlockHash` like this, then I think we can't updated `reg.chain.InitialBlockHash` on the first block committed because concurrent read/writes are possible. Maybe we should pass in the blockchain ID to `protocol.NewChain` explicitly.

protocol/protocol.go
kr: I was thinking of suggesting that we make the caller provide the initial block hash. (I think it's handy there, because it's part of core.Config right?) I didn't think it was important enough to stir the pot, but if the zero-blocks state will be an issue that might be a good way to fix it.

core/asset/asset.go
kr: >  Maybe we should pass in the blockchain ID to `protocol.NewChain` explicitly.

+1 I was thinking the same thing for different reasons

bobg:
OK, I like the feedback on this PR and will revise it.

However, I also like how the change to `asset.NewRegistry` makes it match `account.NewManager` and `query.NewIndexer` and would like to keep that.

jbowens:
> However, I also like how the change to `asset.NewRegistry` makes it match `account.NewManager` and `query.NewIndexer` and would like to keep that.

Fine by me, but it seems like their signatures will probably diverge with time.

bobg:
Closing in favor of https://github.com/chain/chain/pull/58

PR #35, opened 2016-10-24
cmd/ed25519: ed25519 cmd

bobg:
As described in https://gist.github.com/kr/2977d492d6afc428e31f49dee13dc3a7

### DIFF
--- a/cmd/ed25519/ed25519.go
+++ b/cmd/ed25519/ed25519.go
@@ -3,7 +3,6 @@ package main
 import (
 	"encoding/hex"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 
@@ -18,7 +17,7 @@ func main() {
 	case "gen":
 		_, prv, err := ed25519.GenerateKey(nil)
 		must(err)
-		os.Stdout.Write([]byte(prv))
+		os.Stdout.Write(prv)
 
 	case "pub":
 		prv, err := ioutil.ReadAll(os.Stdin)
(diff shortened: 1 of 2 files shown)


bobg:
PTAL

kr:
:eyes:

cmd/ed25519/ed25519.go
```diff
@@ -0,0 +1,105 @@
...
+
+func main() {
+	if len(os.Args) < 2 {
+		usage()
+	}
+	switch os.Args[1] {
+	case "gen":
+		_, prv, err := ed25519.GenerateKey(nil)
+		must(err)
+		os.Stdout.Write([]byte(prv))
```
kr: nit: the `[]byte(...)` conversion is unnecessary

cmd/ed25519/ed25519.go
```diff
@@ -0,0 +1,105 @@
...
+			output = "BAD"
+		}
+		if silent {
+			w = os.Stderr
+		} else {
+			w = os.Stdout
+		}
+		fmt.Fprintln(w, output)
+		if silent && !ok {
+			os.Exit(1)
```
kr: Any reason not to set the exit status in the not-silent mode as well?

cmd/ed25519/ed25519.go
```diff
@@ -0,0 +1,105 @@
...
+			output string
+			w      io.Writer
+		)
+		if ok {
+			output = "OK"
+		} else {
+			output = "BAD"
+		}
+		if silent {
+			w = os.Stderr
```
kr: I feel like maybe silent mode should not print anything (to either output stream). I can imagine trying to use this in a shell script and having to provide both `-s` and also redirect stderr to /dev/null. What do you think?

bobg: I agree, but the gist says otherwise. Switched to be silent in silent mode.

cmd/ed25519/ed25519.go
bobg: Yeah, in non-silent mode it's not failure successfully to detect a bad signature. Using this in a script and parsing its output, you want it to exit with status 0 to not affect the flow of the script.

bobg:
PTAL

cmd/ed25519/ed25519.go
bobg: Shrug, OK.

kr:
LGTM

bobg:
rebased

bobg:
rerebased

PR #37, opened 2016-10-24
cmd/sha3: sha3 cmd

bobg:
As described in https://gist.github.com/kr/2977d492d6afc428e31f49dee13dc3a7

### DIFF
--- a/cmd/sha3/sha3.go
+++ b/cmd/sha3/sha3.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	"flag"
+	"fmt"
+	"hash"
 	"io"
 	"os"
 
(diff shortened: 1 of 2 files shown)


bobg:
PTAL

cmd/sha3/sha3.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"hash"
+	"io"
+	"os"
+
+	"golang.org/x/crypto/sha3"
+)
+
+// Usage:
+//   sha3 <bytes >hash
```
tessr: spacing typo

bobg:
rebased

cmd/sha3/sha3.go
bobg: It's moot now because I removed that comment (it's redundant with doc.go), but what was wrong with the spacing? That it was a 2-space indent instead of 4?

tessr: s/<bytes >hash/<bytes> hash

cmd/sha3/sha3.go
```diff
@@ -11,7 +14,23 @@ import (
 //   sha3 <bytes >hash
 
 func main() {
-	h := sha3.New256()
+	size := flag.Int("n", 256, "size in bits of the desired hash: 224, 256 (default), 384, or 512")
```
kr: nit: the built-in help output prints the default value, so this ends up as

```
size in bits of the desired hash: 224, 256 (default), 384, or 512 (default 256)
```

PR #38, opened 2016-10-24
cmd/hex: new command

kr:


kr:
PTAL

bobg:
Please add the optional `-e` command-line switch we discussed (the explicit antonym of `-d`).

Also, in the category of composability, WDYT of outputting a single line by default?

kr:
> Please add the optional `-e` command-line switch we discussed (the explicit antonym of `-d`).

Oh yeah, will do.

> Also, in the category of composability, WDYT of outputting a single line by default?

Hah, yeah I picked a default length that was big enough for a sha3-256 hash but not big enough to fit our larger structures (tx, block) without breaking lines. I guess this would mean a default of MaxInt. Or should it also leave out the trailing newline (so a default of 0)?

But it is sort of nice to have line breaks for larger items, instead of one gigantic line. Maybe we should consider updating our other tooling (e.g. the base transaction field in a template) to strip out all whitespace before decoding hex??? I don't know how I feel about that. It doesn't strike me as a great alternative.

kr:
PTAL

jbowens:
lgtm

PR #39, opened 2016-10-24
core/signers: take explicit db handle

jbowens:
Instead of pulling the db handle out of the context.Context, take an
explicit pg.DB as a parameter.

If we want, we could go farther and have the signers package keep its
own db handle, but the signers abstraction is a little leaky in a few
places. Both the account and asset packages do JOINs on the signers SQL
table.


kr:
:eyes:

kr:
LGTM

PR #40, opened 2016-10-24
crypto/ed25519/chainkd: remove ed25519 convenience constructor

bobg:
Also fixes an erroneous error in `CHECKSIG`. (Wrong-length pubkey should return false, not error out.)


bobg:
PTAL

kr:
:eyes:

kr:
LGTM

bobg:
rebased

PR #44, opened 2016-10-24
core/rpc: rename package from net/rpc

jbowens:
Our `chain/net/rpc` package isn't really parallel to the standard
library `net/rpc` package and is pretty Chain Core specific. This
commit moves it into `chain/core`.


jbowens:
PTAL

PR #45, opened 2016-10-24
docs: adjust alignement of counters for mobile

Zaryafaraj:


Zaryafaraj:
@oleganza PTAL

docs/css/style.css
```diff
@@ -1262,14 +1262,22 @@ video {
 
 
 /* Media Queries */
-@media (max-width: 992px) {
+@media (max-device-width: 992px) {
```
oleganza: Are you sure this won't break on desktop? Previous rule was applying when browser window is less than 992px wide.

PR #47, opened 2016-10-24
installer/windows: rename chain-core.exe to cored.exe

tessr:
`cored` gets compiled as `cored.exe`. Let's let `cored`s be `cored`s. (Honestly it's just annoying to rename it every time.) 

This PR also fixes some wonky whitespace in the README.


jbowens:
lgtm

PR #48, opened 2016-10-24
core/accesstoken: encapsulate storage

jbowens:
Remove the dependency on the Context's database in the
`core/accesstoken` package.


erykwalder:
LGTM

PR #49, opened 2016-10-25
cmd/slashland: add slack /land command

erykwalder:
This package runs an http server to listen connect to slack to be able
to receive and respond to /land commands. This will land pull requests
on the chain and chainprv github repositories.

### DIFF
--- a/cmd/slashland/main.go
+++ b/cmd/slashland/main.go
@@ -185,7 +185,7 @@ func land(req *landReq) {
 	cmd = dirCmd(landdir, "git", "filter-branch", "-f", "--env-filter", `
 		export GIT_COMMITTER_NAME=$GIT_AUTHOR_NAME
 		export GIT_COMMITTER_EMAIL=$GIT_AUTHOR_EMAIL
-	`, "--", "origin/main..", req.ref)
+	`, "--", req.ref, "^origin/main")
 	cmd.Stderr = os.Stderr
 	err = cmd.Run()
 	if err != nil {
--- a/cmd/slashland/main.go
+++ b/cmd/slashland/main.go
@@ -179,7 +179,7 @@ func land(req *landReq) {
 	cmd = dirCmd(landdir, "git", "filter-branch", "-f", "--env-filter", `
 		export GIT_COMMITTER_NAME=$GIT_AUTHOR_NAME
 		export GIT_COMMITTER_EMAIL=$GIT_AUTHOR_EMAIL
-	`, "--", "origin/main", ".."+req.ref)
+	`, "--", "origin/main..", req.ref)
 	cmd.Stderr = os.Stderr
 	err = cmd.Run()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


erykwalder:
PTAL

cmd/slashland/main.go
```diff
@@ -0,0 +1,354 @@
...
+		err = cmd.Run()
+		sayf("<@%s|%s> failed to land %s: branch has conflicts",
+			req.userID,
+			req.userName,
+			req.ref,
+		)
+		return
+	}
+
+	cmd = dirCmd(landdir, "git", "filter-branch", "--env-filter", `
```
erykwalder: For the actual filter, does that need to contain single quotes, as it would run on the command line?

jbowens:
lgtm, but I don't know the answer to your question

erykwalder:
Updated, PTAL
main differences:
- write a netrc file for authenticating to github
- set git name/email for rebasing
- fix filter-branch command (do we even need this if we're committing as chainbot?)

erykwalder:
I know this has been approved but can I get another look?

erykwalder:
PTAL

kr:
:eyes:

kr:
> do we even need [filter-branch] if we're committing as chainbot?

Chainbot pushes the branch after rebasing, so if we didn't fix the committer then it would show up as Ubuntu or whatever. You're right that github then replaces the committer again when we use the merge api, but as long as the branch is unmerged and rebased by chainbot, it would show up as ubuntu. That caused some confusion while we were juggling git repos around so it's probably worth the effort to preserve the committer? Or maybe we're abusing git and it really should say chainbot as the committer since the commit was generated by chainbot. Maybe we should just write a .gitconfig on chainbot's host so it shows up as `Chainbot <chainbot@chain.com>` instead of `Ubuntu <ubuntu@whatever.ec2.internal>`.

erykwalder:
> Chainbot pushes the branch after rebasing, so if we didn't fix the committer then it would show up as Ubuntu or whatever. You're right that github then replaces the committer again when we use the merge api, but as long as the branch is unmerged and rebased by chainbot, it would show up as ubuntu. That caused some confusion while we were juggling git repos around so it's probably worth the effort to preserve the committer? Or maybe we're abusing git and it really should say chainbot as the committer since the commit was generated by chainbot. Maybe we should just write a .gitconfig on chainbot's host so it shows up as Chainbot <chainbot@chain.com> instead of Ubuntu <ubuntu@whatever.ec2.internal>.

We're now setting the git config as `chainbot <ops@chain.com>`, so even the rebase shows up as that.

kr:
> We're now setting the git config as `chainbot <ops@chain.com>`

Oh, well in that case yeah let's skip the filter branch entirely I guess. (We can always bring it back if some other problem crops up in the future.)

erykwalder:
Filter branch removed, will land soon

PR #50, opened 2016-10-25
docs: add tracking for download links

Zaryafaraj:


Zaryafaraj:
@devongundry PTAL

docs/core/get-started/install.partial.html
```diff
@@ -4,16 +4,18 @@
         <h2>Choose operating system</h2>
         <ul id="download-options" class="tab">
           <li class="download-option" id="download-option-mac">
-            <a href="javascript:void(0)" class="tablinks active" onclick="selectOSTab(event.currentTarget, 'mac')">
+            <a href="javascript:void(0)" class="tablinks active" onclick="trackOption('mac');selectOSTab(event.currentTarget, 'mac')">
```
Zaryafaraj: we track other things so I wanted to keep that event separate. what are the downsides of having them be separate?

devongundry: ok makes sense. so the philosophy is always to include tracking as it's own inline onclick function call.

PR #51, opened 2016-10-25
database/pg: remove context convenience funcs

jbowens:
Delete `pg.Query`, `pg.QueryRow` and `pg.Exec` and update all call sites to either use an explicit db handle in scope or use `pg.FromContext(ctx)`.


kr:
:eyes:

kr:
LGTM

PR #53, opened 2016-10-25
database/pg: remove Begin

jbowens:
Remove pg.Begin. Call sites must explicitly check that a DB handle
implements Beginner.


jbowens:
PTAL. I'm not actually sure what the original use of this code was, but I assume it was for using pgtest.NewTx in tests that tested code already using SQL transactions?

database/pg/context.go
```diff
@@ -53,13 +53,12 @@ var dbKey key
 // associated database.
 //
 // Note: if a transaction is already pending in the passed-in context,
-// this function does not create a new one but returns the existing
-// one.
+// this function panics.
 func Begin(ctx context.Context) (Committer, context.Context, error) {
```
kr: Are we at the point yet where we can delete this function entirely? It is one of the "context convenience functions".

kr:
LGTM

database/pg/context.go
jbowens: > Are we at the point yet where we can delete this function entirely? It is one of the "context convenience functions".

Good call. There are there call sites, all in `core/account/utxodb`. I'll remove it in this PR.

kr:
> I assume it was for using pgtest.NewTx in tests that tested code already using SQL transactions?

Yeah, exactly.

core/account/utxodb/reserve.go
```diff
@@ -57,15 +58,23 @@ type (
 	}
 )
 
+func beginTxn(ctx context.Context) (pg.Tx, error) {
+	db := pg.FromContext(ctx)
+	beginner, ok := db.(pg.Beginner)
```
kr: I think pg.Beginner isn't the interface we need here. A real sql.DB has a different signature. See the deleted pg.begin:

```
type beginner interface {		
	Begin(context.Context) (*sql.Tx, error)		
}		
switch d := db.(type) {		
case beginner: // e.g. *sql.DB		
	return d.Begin(ctx)		
case Beginner: // e.g. pgtest.noCommitDB		
	return d.Begin(ctx)		
}
```

kr: I think maybe we should type assert to the concrete type `*sql.DB` inline at the call sites.

PR #54, opened 2016-10-25
core: remove errors from /health response

kr:
We provide errors in the /info response, which is
authenticated. But /health is not authenticated, and we
don't need to serve the errors in both places. Our
alerting on these errors should provide credentials.


kr:
PTAL

jbowens:
In the future, it might be cool to report some basic health information from this endpoint. Maybe just a single value that's either `HEALTHY` or `DEGRADED`

kr:
rebased

PR #55, opened 2016-10-25
bin/upload-docs: set content type on extensionless files

jeffomatic:
We strip file extensions when rendering static files in order to
provide for pretty, extension-free URLs. Unfortunately, this
removes S3's ability to deduce the MIME type from the extension.

This commit adds a post-sync step that iterates over every
extensionless file in the generated docs, and updates the content
type to text/html. For now, we're not concerned about false
positives.

One other option we explored was having md2html render non-index
leaf pages as BASENAME/index.html, but this caused relative URLs
in the content to be off by one level of depth. Changing relative
URLs to root-prefixed URLs (/foo/bar) was a deep change, and broke
the ability to browse the Markdown links from GitHub.

### DIFF
--- a/bin/upload-docs
+++ b/bin/upload-docs
@@ -21,11 +21,9 @@ echo
 echo "Uploading docs...."
 aws s3 sync --delete $generate_dest_path $bucket_url/docs
 
-# Set text/html MIME types for all files without extensions
-cd $generate_dest_path
-
 echo
 echo "Setting content type for extensionless files, this will take a while..."
+cd $generate_dest_path
 find . -type f ! -name "*.*" |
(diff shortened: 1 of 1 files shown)


jeffomatic:
PTAL @ryandotsmith . Tested this on staging. This, combined with a [public-read bucket policy](http://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html), will give us some peace of mind when publishing docs to the website.

jeffomatic:
PTAL: need to get this across the line so we can upload docs to the website without manual fix-ups.

dominic:
LGTM

PR #56, opened 2016-10-25
cmd/cored: fix auth on unconfigured core

jbowens:
Populate the core.Handler with an accesstoken.CredentialStore when
launching an unconfigured Core. Unconfigured routes still need access
token authentication.

Bug was introduced in #48.


jbowens:
PTAL

kr:
LGTM

PR #58, opened 2016-10-25
protocol/validation: add blockchain and asset ID checking to issuance validation

bobg:

Supersedes https://github.com/chain/chain/pull/26

Updates some logic to handle nil Snapshots.


bobg:
PTAL

core/account/builder_test.go
```diff
@@ -22,11 +22,19 @@ import (
...
 	var (
-		_, db    = pgtest.NewDB(t, pgtest.SchemaPath)
-		ctx      = pg.NewContext(context.Background(), db)
-		c        = prottest.NewChain(t)
+		_, db = pgtest.NewDB(t, pgtest.SchemaPath)
+		ctx   = pg.NewContext(context.Background(), db)
+		c     = prottest.NewChain(t)
+	)
+
+	b1, err := c.GetBlock(ctx, 1)
```
kr: Would it be reasonable for prottest to provide a function to do this lookup and Fatal the test if it's not found? This snippet here crops up in enough places now it would probably pay for itself pretty easily.

core/generator/generator.go
```diff
@@ -75,7 +75,16 @@ func Generate(
 		log.Fatal(ctx, log.KeyError, err)
 	}
 	if b != nil && (g.latestBlock == nil || b.Height == g.latestBlock.Height+1) {
-		s := state.Copy(g.latestSnapshot)
+		var s *state.Snapshot
+		if g.latestSnapshot == nil {
```
kr: I'm not 100% following why these nil snapshot changes should happen together with plumbing and checking the initial block hash in validation.

bobg:
rebased

core/generator/generator.go
jbowens: > so maybe Chain.Recover can just error out in the empty-blockchain case

I think we need to support that case for booting a new participant node w/o a snapshot. Can we pass the blockchain ID into `Recover` instead?

> I also made CommitBlock synchronously store a snapshot for the initial block (because I was getting non-deterministic failures in TestGetAndAddBlockSignatures)

Hrm, can we make any changes to TestGetAndAddBlockSignatures to make it more reliable? It'd be nice to avoid special cases like this because it makes the invariants more complicated.

bobg: > Can we pass the blockchain ID into Recover instead?

Looking at the code, the best way to do that seems to be by storing it in the `Chain` object as in the earlier version of this PR. I'm fine with that, but maybe you're not?

jbowens: What about passing it explicitly as an argument? We have a blockchain ID in cored when we launchConfiguredCore. That will call `leader.Run`, which in turn calls either `fetch.Fetch` or `generator.Generate`. Those are the two functions that call `Recover`.

bobg: That starts to make some functions look like parameter salad, but I gave it a try - WDYT?

Note that `Chain.ValidateBlockForSig` still needs to query the store for block 1 in order to instantiate a snapshot - or else have the initial block hash plumbed into it, too - which could be done easily if `Chain` included it as a member.

core/txdb/snapshot.go
```diff
@@ -40,10 +40,12 @@ func DecodeSnapshot(data []byte) (*state.Snapshot, error) {
 		issuances[hash] = issuance.ExpiryMs
 	}
 
-	return &state.Snapshot{
+	result := &state.Snapshot{
 		Tree:      tree,
 		Issuances: issuances,
-	}, nil
+	}
+	copy(result.InitialBlockHash[:], storedSnapshot.InitialBlockHash)
```
jbowens: Do we need it on the serialized snapshot proto? For backwards compatibility, I think we're going to need to lookup the initial block hash because any existing snapshots won't have it populated.

kr:
FWIW I it's fine by me putting the initial block hash as a field on the protocol.Chain (possibly in addition to in the state.Snapshot). Especially since we initialize it in the NewChain call so it's guaranteed to be available.

```
func NewChain(..., b1Hash bc.Hash, ...) *Chain

func (c *Chain) B1Hash() bc.Hash
```

How does that sound? (Perhaps with a better name than "B1hash" :p)

core/txdb/snapshot.go
jbowens: We'll be ok as long as we don't deploy until the next testnet reset.

jbowens:
If we populate the initial block hash when calling NewChain(...) that sgtm too.

kr:
>  Especially since we supply it as an argument to the NewChain call

Oh, I guess we don't actually do that yet. Hah!

bobg:
I think part of the difficulty we're having is that the abstraction boundary between `Chain` and `Snapshot` is a little arbitrary. Both represent aspects of the current state of the blockchain.

kr:
Making the blockchain ID a parameter of protocol.NewChain might be a nice leaf item we could pick off and do first. Any reason we wouldn't want to to that anyway? A protocol.Chain is absolutely tied to a single initial block hash, even if it doesn't have any blocks yet. I think this might helpfully reduce the design space (and afford more opportunities for convenient safety checks in the future).

bobg:
> Making the blockchain ID a parameter of protocol.NewChain might be a nice leaf item we could pick off and do first

Great call, will do. Un-PTAL this PR for now.

bobg:
Closed in favor of https://github.com/chain/chain/pull/74

PR #60, opened 2016-10-25
core/account/utxodb: encapsulate storage

jbowens:
Unfortunately, this means a few tests need to use pgtest.NewDB
instead of pgtest.NewTx to satisfy the account.NewManager signature.

### DIFF
--- a/core/account/accounts.go
+++ b/core/account/accounts.go
@@ -52,9 +52,8 @@ func (m *Manager) IndexAccounts(indexer Saver) {
 	m.chain.AddBlockCallback(m.indexAccountUTXOs)
 }
 
-// ExpireReservations removes reservations that have expired on the provided
-// period until the context is cancelled. It blocks until the context is
-// cancelled.
+// ExpireReservations removes reservations that have expired periodically.
+// It blocks until the context is canceled.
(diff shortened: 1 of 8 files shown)


jbowens:
PTAL

bobg:
:eyes:

core/account/accounts.go
```diff
@@ -3,15 +3,18 @@ package account
 
 import (
 	"context"
-	"database/sql"
+	stdsql "database/sql"
```
bobg: Another case of "which import to alias?" Just calling it out for now, don't need to fix it in this PR, but for the record we should standardize on something, and my strong preference is to let standard packages be standard and to call chain/database/sql `chainsql`.

core/account/accounts.go
```diff
@@ -47,6 +52,13 @@ func (m *Manager) IndexAccounts(indexer Saver) {
 	m.chain.AddBlockCallback(m.indexAccountUTXOs)
 }
 
+// ExpireReservations removes reservations that have expired on the provided
```
bobg: Comment could use a little editing.

bobg:
LGTM, see comments

PR #63, opened 2016-10-25
vendor: remove grpc package and its dependencies

erykwalder:
This package is not needed until GRPC is closer on the timeline. Rather than just reverting the commit that added them, this selectively removes the dependencies that are not in use.


tessr:
I'm confused that you had to add x/context back before removing it again (especially if you weren't reverting a commit), but the end result looks good.

erykwalder:
Ha, that is what I did, but that message is for the final commit

tessr:
Lol, caught you red-handed

PR #67, opened 2016-10-26
errors: add a field for arbitrary data

kr:
We plan to use this to report more error detail to RPC clients in a structured format. The Detail functions provide only unstructured text.


jbowens:
lgtm

PR #70, opened 2016-10-26
cmd/testbot: fetch pull request instead of branch ref

boymanjor:
This fixes a bug that didn't allow testbot to run
against pull requests from forked repos.

### DIFF
--- a/cmd/testbot/main.go
+++ b/cmd/testbot/main.go
@@ -74,39 +74,21 @@ func prHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	log.Println("pr ref:", req.PR.Head.Ref)
+	log.Println("pr sha:", req.PR.Head.Sha)
 	log.Println("pr action:", req.Action)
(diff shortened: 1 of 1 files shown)


kr:
:eyes:

boymanjor:
PTAL

boymanjor:
@kr is it fine to just fetch the pr references and use the commit hashes?

kr:
> is it fine to just fetch the pr references and use the commit hashes?

Yeah that should work great.

kr:
:eyes:

cmd/testbot/main.go
```diff
@@ -35,10 +35,12 @@ var (
 type pullRequest struct {
 	Action string
 	PR     struct {
-		Head struct {
+		Number string
+		Head   struct {
 			Ref string
 			Sha string
```
kr: (nit: out of scope for this PR, but this should be named `SHA` rather than `Sha`)

kr:
LGTM

PR #71, opened 2016-10-26
docs: rename block generator to block proposer throughout Chain Core

danrobinson:
"Proposer" is a more conventional name (it dates back to Paxos), and more accurately describes the role that the "generator" plays in the blockchain. (The other common terms, "leader" and "primary", would be misleading.)


oleganza:
I believe we'll need a DB migration code too.

devongundry:
We'll also need to add a changelog document to the references section in the Chain Core docs describing this change

devongundry:
I think we should also start thinking of the role of proposing blocks separately from the role of being the connection into the network, even though they are currently performed by the same core.

Practically, this means that when you connect your core to an existing network, you would provide a network URL instead of a block proposer URL

danrobinson:
@devongundry I definitely agree, but seems like the change is a bit complex and would require logic as well as renaming—I think signers currently use one variable to represent both the proposer's URL and the network URL. Maybe a future PR (but still before this is actually released)?

devongundry:
I'm just talking about a renaming issue. I want to call it network URL, not block proposer url

kr:
> I want to call it network URL, not block proposer url

Is the idea here that you might connect to the network via some other participant other than the block proposer, and you send transactions to and get blocks from the intermediary node? That seems pretty good and reasonable.

The words for the closest equivalent to this in IP are "gateway" and "router". Just as another point of reference.

devongundry:
I think that's part of it. Certainly something we may support on the future.

But right now, what I'm trying to accomplish is to avoid introducing `block_proposer_url' and `block_proposer_network_token` to network participants. I.e. the configure endpoint and the configure form on the dashboard.

`gateway_url` is nice.

PR #72, opened 2016-10-26
protocol: add initial block hash field to the Chain object

bobg:
Resurrects some code from https://github.com/chain/chain/pull/26. Related to https://github.com/chain/chainprv/issues/2404.

### DIFF
--- a/protocol/block_test.go
+++ b/protocol/block_test.go
@@ -35,7 +35,7 @@ func TestGetBlock(t *testing.T) {
 	}
 
 	for _, test := range cases {
-		c, err := NewChain(ctx, test.store, emptyPool, nil)
+		c, err := NewChain(ctx, b1.Hash(), test.store, emptyPool, nil)
 		if err != nil {
 			testutil.FatalErr(t, err)
 		}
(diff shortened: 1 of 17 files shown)


bobg:
PTAL

jbowens:
lgtm

protocol/block_test.go
```diff
@@ -221,16 +221,19 @@ func TestValidateBlockForSig(t *testing.T) {
 // It commits b1 before returning.
 func newTestChain(tb testing.TB, ts time.Time) (c *Chain, b1 *bc.Block) {
 	ctx := context.Background()
-	c, err := NewChain(ctx, memstore.New(), mempool.New(), nil)
+
+	var err error
+
+	b1, err = NewInitialBlock(nil, 0, ts)
```
kr: nit: we could use `:=` here

protocol/protocol.go
```diff
@@ -103,16 +103,18 @@ type Pool interface {
 // validation logic from package validation to decide what
 // objects can be safely stored.
 type Chain struct {
+	InitialBlockHash  bc.Hash
```
kr: We might consider making this a method to enforce that it's readonly.

kr:
(I was hitting the approve button anyway, so I my comments above are not a big deal.)

PR #73, opened 2016-10-26
core: use specific error code for non-unique alias

jbowens:
Use a specific Chain error code for errors that try to create an object
with an alias that already exists. Previously, this error was bundled
under the generic bad request error code. It's useful to distinguish so
that clients can write get-or-create functions.

I'm not sure what the appropriate error namespace for this is.


bobg:
:eyes:

bobg:
LGTM

PR #74, opened 2016-10-26
protocol/validation: add blockchain ID checking to issuance validation

bobg:

Supersedes https://github.com/chain/chain/pull/58

Back roughly where we started in https://github.com/chain/chain/pull/26

### DIFF
--- a/protocol/validation/tx.go
+++ b/protocol/validation/tx.go
@@ -31,7 +31,7 @@ var (
 // to the pool.
 //
 // ConfirmTx must not mutate the snapshot or the block.
-func ConfirmTx(snapshot *state.Snapshot, block *bc.Block, tx *bc.Tx) error {
+func ConfirmTx(snapshot *state.Snapshot, initialBlockHash bc.Hash, block *bc.Block, tx *bc.Tx) error {
 	if block.Version == 1 && tx.Version != 1 {
 		return errors.WithDetailf(ErrBadTx, "unknown transaction version %d for block version 1", tx.Version)
 	}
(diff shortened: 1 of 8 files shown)


bobg:
PTAL

kr:
:eyes:

protocol/validation/tx.go
```diff
@@ -48,6 +48,13 @@ func ConfirmTx(snapshot *state.Snapshot, block *bc.Block, tx *bc.Tx) error {
 			if txin.AssetVersion != 1 {
 				continue
 			}
+			if ii.InitialBlock != initialBlockHash {
+				return errors.WithDetail(ErrBadTx, "issuance is for different blockchain")
+			}
+			assetID := bc.ComputeAssetID(ii.IssuanceProgram, initialBlockHash, ii.VMVersion)
```
jbowens: super tiny nit: should we use `ii.InitialBlock` instead of `initialBlockHash`? When we support joining chains, I think this would need to use `ii.InitiialBlock` and the above check would check that `ii.InitialBlock` is in a set of acceptable initial block hashes

jbowens:
oops, might want to wait for @kr to finish too

protocol/validation/tx.go
bobg: Hm, doing that reveals how redundant the assetID check is (because `ii.AssetID()` simply does `ComputeAssetID` of the very same values). I've _removed_ the assetID check. Please verify this still LGTY.

jbowens: Hrmm, the exact words from the spec are:
> Compute asset ID from the initial block ID, asset version 1, and the VM version and issuance program declared in the witness. If the resulting asset ID is not equal to the declared asset ID in the issuance commitment, halt and return false.

I think maybe we need to check that here: https://github.com/chain/chain/blob/main/protocol/bc/txinput.go#L181

Currently, we just read the asset ID from the commitment and disregard it.

jbowens:
lgtm

bobg:
rebased

PR #77, opened 2016-10-26
docs/internal: release versioning

oleganza:
This is a draft proposal, a request for comments.


docs/internal/release-versions.md
```diff
@@ -0,0 +1,50 @@
...
+
+We need a consistent, yet simple, versioning scheme that allows both individual updates as well as grouped updates: from major feature releases to smaller, maintenance releases.
+
+We probably do not want to formally commit to specific semantics of the version numbers, such as [SemVer](http://semver.org), but we need to keep all numbers in order, so it’s easier navigate them.
+
+## Proposal
+
+Use the following version format:
+
+    <major>.<minor>.<build>
```
tessr: Do you think it's a problem to have versioning that _looks like_ semver but _isn't_ semver?

oleganza: I'm not sure semver stole the dot-separated versioning from all software in the world.

oleganza: @tessr pushed extra commit on rationale for dated build version (credits go to @kr)

tessr: I actually like this a lot and it makes sense to me--just wanted to ask the question.

oleganza: Yup, the question is fair.

oleganza: @tessr Since some may expect not to experience breaking changes from 1.5 to 1.6, we should try to keep it that way. And if we need to break something without major version bump, then we'll make sure we have proper migrations or deprecation notices, or translation layers in place. Probably the doc should expand on that to make it more clear.

docs/internal/release-versions.md
```diff
@@ -0,0 +1,129 @@
...
+## Examples
+
+Suppose, we have the following versions currently shipping:
+
+* Chain Core 1.0
+* Java SDK 1.0
+* Chain Core for Mac 1.0
+* Chain Core Windows Installer 1.0
+
+### Case 1: a bugfix in the SDK
```
kr: I'm not sure it makes sense to distinguish between a bugfix and a new feature in this place. What if we add a new feature to the sdk?

kr: Or is it just an example. But still it looks like this is meant to contrast with the example below where we  fix a bug vs add a feature to the API.

oleganza: Added Case 4b and Case 5 to address your questions. https://github.com/chain/chain/pull/77/commits/04ea644bb86bb5268a61722beb06805b063411d9

docs/internal/release-versions.md
```diff
@@ -0,0 +1,167 @@
...
+
+* 1.0
+* 1.0.20161024
+* 1.7.20170119
+* 2.3.20170601
+
+
+## Build version
+
+The reason to use dates for **build** versions is to avoid frequent collisions across different software packages that may convey non-existant correlation. E.g. if at some point Java SDK and Mac app both have version `1.3.2`, some may be confused by the identical version.
```
jeffomatic: Do you see this as a significant problem? There is a nice simplicity advantage to using an incremented integer for the build/patch version, and it allows for the not-unheard-of possibility of updating more than once in a day.

docs/internal/release-versions.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Build version
+
+The reason to use dates for **build** versions is to avoid frequent collisions across different software packages that may convey non-existant correlation. E.g. if at some point Java SDK and Mac app both have version `1.3.2`, some may be confused by the identical version. 
+
+On the other hand, dependent releases usually happen on the same day and therefore most likely will have the identical versions, indicating the correlation truthfully. E.g. if Chain Core adds a minor fix in the API, all SDKs may be updated with the same new build number: `1.0.161205`.
+
+## Compatibility
+
+We should strive to keep expectation that minor version updates do not break backward compatibility. E.g. if we need to introduce a completely new SDK, we should bump the major version, or rename the SDK. If we need to retire/improve all APIs, we should add a compatibility layer and diagnostic messages to the SDK without simply removing them.
```
jeffomatic: I read this as "if any user code will break with a given change, upgrade the major version".

Taking stock of that rule, plus the breadth of the software (spanning the client API, network API, and all SDKs), it seems as if Chain Core will undergo relatively many major version changes, at least compared to other most other software. Sampling at random, I will note that Postgres is on major version 9 after 20 years of development, and that Ruby on Rails is on version 5 after over a decade. Given that we have lots of ideas for Chain Core's evolution, it wouldn't surprise me if we were on version 50 this time next year.

I think this is _totally fine_, but since it is a little unconventional, it just feels worth pointing out. I sort of assume we will get used to upgrading the major version fairly casually.

PR #81, opened 2016-10-27
cmd/benchcore: add -with flag

jbowens:
Add a -with flag so that we can upload test data on the client EC2
instance for the java program to consume.


jbowens:
PTAL

jbowens:
Should -with take multiple files separated by ListSeparator?

jbowens:
> Should -with take multiple files separated by ListSeparator?

I guess it won't be hard to follow up with that when there's a need.

PR #85, opened 2016-10-27
core/txbuilder: return error info for all actions

kr:
It will be useful for clients to see all available error messages at once, so they can fix them all without having to make multiple requests.


kr:
PTAL @jbowens

jbowens:
lgtm

kr:
PTAL

jbowens:
looks even gooder to me

PR #87, opened 2016-10-27
protocol/state: put b1 hash in Snapshot

kr:


protocol/state/snapshot.go
```diff
@@ -13,6 +13,7 @@ type PriorIssuances map[bc.Hash]uint64
 // Snapshot encompasses a snapshot of entire blockchain state. It
 // consists of a patricia state tree and the issuances memory.
 type Snapshot struct {
+	B1Hash    bc.Hash
```
jbowens: Are we calling it B1Hash now?

kr: No, sorry, this pull request isn't really ready for review.

In fact… it doesn't really look like a good idea IMO. It's more code and doesn't really bring much benefit.

PR #90, opened 2016-10-28
sdk/java: support additional error details

jbowens:
In 411d336 we changed the response of the build endpoint. Errors when
building actions now result in a CH706 error. Additional error details
per-action are included on a nested `data` key. This adds support for
this in the SDK by adding a BuildException class and parameterizing
BatchResponse.

### DIFF
--- a/docs/core/examples/java/BatchOperations.java
+++ b/docs/core/examples/java/BatchOperations.java
@@ -30,7 +30,7 @@ class BatchOperations {
     // endsnippet
 
     // snippet asset-create-batch
-    BatchResponse<Asset> assetBatch = Asset.createBatch(client, assetBuilders);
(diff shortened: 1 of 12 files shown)


jbowens:
PTAL @kr

kr:
:eyes:

jeffomatic:
This is great, although I think we can minimize the impact of this code change if we:

1. Selectively instantiate `BuildException` in the low-level `post` method by detecting `CH706`, rather than through type hinting via generics. That seems like a nice use case for machine-readable error codes.
2. Add to the documentation a recommendation that the user either explicitly catch `BuildException` (when in non-batch mode), or perform an `instanceof BuildException` check when iterating over batch errors.

jbowens:
!PTAL for now, going with less-invasive #91 to unblock other work

PR #91, opened 2016-10-28
sdk/java: add BuildException for build errors

jbowens:
Unmarshal build errors into BuildException instead of APIException.
Clients will need to use instanceof and cast build errors to
BuildException if using the batch endpoint.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/BatchResponse.java
+++ b/sdk/java/src/main/java/com/chain/http/BatchResponse.java
@@ -24,7 +24,7 @@ public class BatchResponse<T> {
   /**
    * This constructor is used when deserializing a response from an API call.
    */
-  public BatchResponse(Response response, Gson serializer, Type tClass)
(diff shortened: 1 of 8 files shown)


jbowens:
PTAL

sdk/java/src/main/java/com/chain/http/BatchResponse.java
```diff
@@ -34,7 +34,7 @@ public BatchResponse(Response response, Gson serializer, Type tClass)
         JsonElement elem = root.get(i);
 
         // Test for interleaved errors
-        APIException err = serializer.fromJson(elem, APIException.class);
+        APIException err = serializer.fromJson(elem, eClass);
```
jeffomatic: My sense is that we could probably reduce the surface area of this change by peeking at the code in the response and check for CH706, but since we'd like to get this out soon this is fine for now.

PR #94, opened 2016-10-28
core/txbuilder: roll back builder side-effects

kr:
If one action succeeds with some side effects (most
notably, reserving some UTXOs) and another action fails,
we should roll back the side effects if possible so that
the overall operation has no effect.

Note that this is not a guaranteed operation, so any
side effects need to be designed to be safe not to roll
back. Reservations are: they will eventually expire on
their own. Meanwhile, it helps performance to eagerly
cancel them upon failure rather than waiting for the
deadline to pass.

### DIFF
--- a/core/account/builder.go
+++ b/core/account/builder.go
@@ -57,10 +57,12 @@ func (a *spendAction) Build(ctx context.Context, maxTime time.Time) (*txbuilder.
 		ClientToken: a.ClientToken,
 	}
 	utxodbSources := []utxodb.Source{utxodbSource}
-	reserved, change, err := a.accounts.utxoDB.Reserve(ctx, utxodbSources, maxTime)
+	// TODO(kr): make utxodb.Reserve tkae a single Source not a slice
+	rids, reserved, change, err := a.accounts.utxoDB.Reserve(ctx, utxodbSources, maxTime)
 	if err != nil {
 		return nil, errors.Wrap(err, "reserving utxos")
 	}
+	rid := rids[0] // len(rids)==len(utxodbSources)
 
 	var (
 		txins      []*bc.TxInput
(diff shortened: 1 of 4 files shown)


kr:
PTAL

core/account/utxodb/reserve.go
```diff
@@ -141,28 +140,31 @@ func (res *Reserver) ReserveUTXO(ctx context.Context, txHash bc.Hash, pos uint32
...
 	}
 
 	defer func() {
 		if err != nil {
-			dbtx.Exec(ctx, "SELECT cancel_reservations($1)", pq.Int64Array(reservationIDs)) // ignore errors
+			var a pq.Int64Array
+			for _, rid := range reservationIDs {
+				a = append(a, int64(rid))
+			}
+			dbtx.Exec(ctx, "SELECT cancel_reservations($1)", a) // ignore errors
```
jbowens: I think `pq.Int64Array` needs a pointer receiver.

core/account/builder.go
```diff
@@ -56,10 +57,12 @@ func (a *spendAction) Build(ctx context.Context, maxTime time.Time) (*txbuilder.
 		ClientToken: a.ClientToken,
 	}
 	utxodbSources := []utxodb.Source{utxodbSource}
-	reserved, change, err := a.accounts.utxoDB.Reserve(ctx, utxodbSources, maxTime)
+	// TODO(kr): make utxodb.Reserve tkae a single Source not a slice
```
jbowens: nit: s/tkae/take/

jbowens:
lgtm modulo the pointer receiver thing

core/account/utxodb/reserve.go
jbowens: jk

kr: Resolved offline: this is providing a value, not scanning.

PR #95, opened 2016-10-28
cmd/varint: new command for encoding/decoding varints

tessr:


cmd/varint/flag.go
```diff
@@ -0,0 +1,31 @@
+package main
```
tessr: This whole file is borrowed from cmd/hex. Should we move this stuff somewhere where both commands can use it?

tessr:
PTAL

bobg:
:eyes:

cmd/varint/flag.go
bobg: My suggestion: Let's round out our CLI tools suite and then see what makes best sense to factor out into some sort of CLI-tool library.

cmd/varint/varint.go
```diff
@@ -0,0 +1,85 @@
...
+
+var signed bool
+
+func main() {
+	log.SetFlags(0)
+	log.SetPrefix("varint: ")
+
+	flag.BoolVar(&signed, "s", false, "signed")
+	flagNotBoolVar(&signed, "u", true, "unsigned (default)")
+	flag.Parse()
```
bobg: Will `flag.Parse` and `flag.Args` do the right thing when the first non-flag argument starts with a minus sign (i.e., a negative number to encode)?

I don't know if it does or not, but it's worth checking.

bobg:
LGTM but before committing please check about non-flag args that look like flags (as mentioned in my comment above).

cmd/varint/flag.go
kr: Sure, like chain/flag? (Or open to other ideas.)

cmd/varint/doc.go
```diff
@@ -0,0 +1,28 @@
...
+
+Examples:
+
+Obtain the decimal value of the hex-encoded varint 0101:
+
+      printf 0101 | hex -d | varint -u
+
+Obtain the unsigned varint value of the decimal number 1234:
+
+      varint -u 1234 | hex -e
```
kr: nit: I would leave off flags for default behavior in general (so both flags here and -u on line 18)

cmd/varint/varint.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		return
+	}
+
+	// encode from args
+	if len(args) != 1 {
+		errorf("invalid argument count %d; varint must read from stdin or take 1 argument", len(args))
+	}
+
+	val, err := strconv.ParseInt(args[0], 10, 64)
```
kr: We might need to use ParseUint for the unsigned case. For example if the command is `varint -u 9223372036854775808` then ParseInt will return an error. (Not a big deal because we probably won't be exercising those edge cases in practice, but it'd be best to handle it correctly, heh.)

cmd/varint/varint.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		var nbytes int
+		if signed {
+			var n int64
+			n, nbytes = binary.Varint(b)
+			if nbytes <= 0 {
+				errorf("could not parse varint")
+			}
+			_, err = os.Stdout.Write([]byte(strconv.FormatInt(n, 10)))
```
kr: This could also be just `fmt.Println(n)`.

cmd/varint/varint.go
```diff
@@ -0,0 +1,85 @@
...
+			if err != nil {
+				errorf("could not write to stdout: %s", err)
+			}
+		} else {
+			var n uint64
+			n, nbytes = binary.Uvarint(b)
+			if nbytes <= 0 {
+				errorf("could not parse uvarint")
+			}
+			_, err = os.Stdout.Write([]byte(strconv.FormatUint(n, 10)))
```
kr: Same here, could be `fmt.Println(n)`.

cmd/varint/flag.go
tessr: @bobg: I like that take on it. For now, I'll leave this file in here.

tessr:
Removed flags (:tada:), and made other fixes. PTAL

kr:
:eyes:

kr:
LGTM. Nice.

PR #96, opened 2016-10-28
core/account/utxodb: fix Reserve defer func

jbowens:
We don't define any PL/pgSQL `cancel_reservations` function, there's
only `cancel_reservation`. Also, rolling back is already covered by the
deferred SQL transaction rollback, so this call is redundant.

### DIFF
--- a/core/account/utxodb/reserve.go
+++ b/core/account/utxodb/reserve.go
@@ -6,8 +6,6 @@ import (
 	stdsql "database/sql"
 	"time"
 
-	"github.com/lib/pq"
-
 	"chain/database/pg"
 	"chain/database/sql"
 	"chain/errors"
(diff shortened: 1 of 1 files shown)


jbowens:
PTAL

bobg:
LGTM

Is it worth writing a plpgsql `cancel_reservations`? Perhaps not.

jbowens:
Should I just get rid of the defer? It seems redundant with the defer rollback above.

bobg:
> Should I just get rid of the defer? It seems redundant with the defer rollback above.

Oh, totally. Good call.

However, if an error occurs at a higher level in the callstack, it would be ideal to cancel any reservations then, which means returning a cancelable reservation object from this function. Not sure if it's really worth the effort, but certainly worth mentioning in a comment.

bobg:
> which means returning a cancelable reservation object from this function

Oh we already do (the reservation ID).  Carry on...

PR #97, opened 2016-10-28
core/account/utxodb: take single Source in Reserve

jbowens:
Change the Reserve signature to take a single utxodb.Source instead
of a slice of utxodb.Sources. We only ever provide a single element
slice to the sources parameter.


bobg:
:eyes:

bobg:
LGTM

PR #98, opened 2016-10-28
encoding/json: treat null as zero-value duration

jeffomatic:
When deserializing API request objects, it's useful to treat null
the same as undefined. This is especially convenient for
high-level languages, where it is common to assign values to
key-value objects that may or may not be null.


jeffomatic:
PTAL

erykwalder:
LGTM

PR #99, opened 2016-10-28
core/account/utxodb: rename to DBReserver

jbowens:
Rename to DBReserver to reflect that reservation happens within the DB.
Also, move the periodic logic of ExpireReservations out of utxodb and
into the account package.

This is preparation for adding an in-memory reserver that can be used
when a Core has a single process.


jbowens:
PTAL

bobg:
LGTM

PR #100, opened 2016-10-28
core/account: always lock the account_utxos table

bobg:
We were writing to the `account_utxos` table during indexing without locking it, but that table requires locking. It probably led to deadlock in a recent test run.


bobg:
PTAL

kr:
:eyes:

kr:
LGTM

bobg:
rebased

PR #101, opened 2016-10-28
core: set errorInfo 'temporary' field for CH706

jbowens:
The CH706 error encapsulates multiple errors. It should be marked as
temporary if all of its contained errors are also temporary. This is
relevant because CH761 can be returned by actions and is temporary.


kr:
LGTM

PR #103, opened 2016-10-28
core/account/utxodb: add in-memory reserver

jbowens:
Replace the PL/pgSQL UTXO reserver implementation with
an in-memory implementation. For now, non-leader processes
forward all build calls to the leader process for reservation.

Account UTXOs are still stored in the `account_utxos`
PostgreSQL table. Only the reservations themselves are
held in-memory.


jbowens:
PTAL

Any suggestions on how to adapt TestConflictingTxsInPool now that we can't cancel reservations by just modifying the DB directly? I could add a function to cancel by outpoint, but it seems like a shame to export it just so that a test can use it.

bobg:
:eyes:

core/account/utxodb/reserve.go
```diff
@@ -0,0 +1,365 @@
...
+	ClientToken *string `json:"client_token"`
+}
+
+// Reservation describes a reservation of a set of UTXOs belonging
+// to a particular account. Reservations are immutable.
+type Reservation struct {
+	ID          uint64
+	AccountID   string
+	UTXOs       []*UTXO
+	Change      []Change
```
bobg: See https://github.com/chain/chain/pull/138. You can simplify `Change` to be a single `uint64` (among other improvements).

core/account/utxodb/reserve.go
```diff
@@ -0,0 +1,365 @@
...
+
+	untypedRes, err := re.idempotency.Once(*source.ClientToken, func() (interface{}, error) {
+		return re.reserve(ctx, source, exp)
+	})
+	return untypedRes.(*Reservation), err
+}
+
+func (re *Reserver) reserve(ctx context.Context, source Source, exp time.Time) (res *Reservation, err error) {
+	// Find the set of UTXOs that match this source.
+	utxos, err := re.findMatchingUTXOs(ctx, source)
```
bobg: I'm a little concerned that `findMatchingUTXOs` will often produce far too many results.

Have you considered bundling up the rest of this function as a callback that can be passed down into the db row-scanning loop, so that that loop can exit early? It would mean turning some of the logic inside out, which would be a shame, because this is generally structured very well. Maybe stream matching utxos over a channel and tie it off when the request is satisfied?

bobg:
> Any suggestions on how to adapt TestConflictingTxsInPool

You can build `secondTemplate` "manually" rather than by calling `Build`. That way it can spend the same utxo but have a different output.

In fact you should be able to just copy `firstTemplate` and make the smallest possible change to the tx (e.g. in some refdata) to give it a new hash.

kr:
> A request: before submitting this change, can we land #138 and #123? Then if we ever want to go back to the db reserver code it would be in its best state.

+1 on #138
-1 on #123

We'll still have the code in the PR's zombie branch if we ever need it, but I think we should branch predict that we won't need it and so not spend the effort now to merge the change. (I am not sure merging it would save us much or any effort later anyway. Once the code diverges, it won't be a simple git revert.)

core/account/utxodb/reserve.go
kr: Yeah, at the very least we can keep this idea around as a potential optimization. Should we try it the clean way first?

core/account/utxodb/reserve.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+	}
+
+	if reserved+unavailable < src.Amount {
+		// Even if everything was available, this account wouldn't have
+		// enough to satisfy the request.
+		return nil, 0, ErrInsufficient
+	}
+	if reserved < src.Amount {
+		// The account has enough for the request, but it's all tied up in
```
kr: s/it's all/some is/

core/txbuilder/finalize_test.go
```diff
@@ -33,69 +31,9 @@ import (
...
-	dumpState(ctx, t, db)
-	b := prottest.MakeBlock(t, info.Chain)
-	<-info.pinStore.WaitForPin(account.PinName, info.Chain.Height())
-
-	dumpState(ctx, t, db)
-	if len(b.Transactions) != 1 {
-		t.Errorf("got block.Transactions = %#v\n, want exactly one tx", b.Transactions)
-	}
+	// TODO(jackson): Figure out a proper interface to be able
+	// to cancel the reservation.
```
kr: Can this be done now? It looks like there is a cancel method.

jbowens: There's no reservation ID propagated through the template. Yeah, I think it should be done before landing the PR, but I wanted to get thoughts on how to accomplish it.

jbowens:
> You can build secondTemplate "manually" rather than by calling Build. That way it can spend the same utxo but have a different output.
>
> In fact you should be able to just copy firstTemplate and make the smallest possible change to the tx (e.g. in some refdata) to give it a new hash.

How does that interact with the txsighash constraint? Do I need to modify the signing instructions as well?

bobg:
> How does that interact with the txsighash constraint?

Just tried it. If it's not too ugly for you, all you need to do is, after assigning `secondTemplate`, add these lines:

```
	secondTemplate.SigningInstructions[0].WitnessComponents[0].(*SignatureWitness).Program = nil
	secondTemplate.SigningInstructions[0].WitnessComponents[0].(*SignatureWitness).Sigs = nil
```

jbowens:
> Just tried it. If it's not too ugly for you, all you need to do is, after assigning secondTemplate, add these lines:

Ah, great! Thank you!

jbowens:
Updated:
* Use `sync/idempotency` for ensuring idempotency of reservations
* Replace 'accountReserver' with 'sourceReserver' (keyed on account ID & asset ID)
* Use singleflight to prevent duplicate `account_utxos` queries.
* Remove explicit `account_utxos` table-level locks
* Adapt `TestConflictTxsInPool`

PTAL

bobg:
LGTM

Excited!

PR #104, opened 2016-10-29
cmd/benchcore: fix multiple instance types

kr:
Not all AMIs can be used on all instances, so this change adds a lookup table. Some instance types also require us to specify an EBS volume to get more than a few GB of storage.


kr:
PTAL

jbowens:
lgtm

PR #105, opened 2016-10-29
core: up max request size to 1mb

jbowens:
If UTXOs are fragmented, sign transaction requests (and maybe submit requests too) can exceed the current 100kb request body limit.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -79,7 +79,7 @@ type RequestLimit struct {
 }
 
 func maxBytes(h http.Handler) http.Handler {
-	const maxReqSize = 1e5 // 100kB
+	const maxReqSize = 1e7 // 10mb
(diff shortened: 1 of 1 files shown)


oleganza:
As a tribute to Satoshi, we should use either 1mb (block size limit) or 32 mb (p2p network message size limit).

jbowens:
PTAL

core/api.go
```diff
@@ -79,7 +79,7 @@ type RequestLimit struct {
 }
 
 func maxBytes(h http.Handler) http.Handler {
-	const maxReqSize = 1e5 // 100kB
+	const maxReqSize = 1e6 // 1mb
```
kr: nit: "MB" should be in caps

PR #108, opened 2016-10-29
cmd/testnet-reset: update testnet info after reset

boymanjor:
This commit updates the BLOCKCHAIN_ID config var
in the chain-testnet-info app after the network is reset.

### DIFF
--- a/cmd/testnet-reset/main.go
+++ b/cmd/testnet-reset/main.go
@@ -3,8 +3,11 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"chain/core/rpc"
--- a/cmd/testnet-reset/main.go
+++ b/cmd/testnet-reset/main.go
@@ -3,7 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
-	"io/ioutil"
+	"io"
 	"log"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


kr:
:eyes:

cmd/testnet-reset/main.go
```diff
@@ -94,6 +98,22 @@ func main() {
...
+	method := "PATCH"
+	url := "https://api.heroku.com/apps/chain-testnet-info/config-vars"
+	r := strings.NewReader(`{"BLOCKCHAIN_ID":"` + resp.BlockchainID + `"}`)
+	client := http.Client{}
+	req, err := http.NewRequest(method, url, r)
+	must(err)
+	req.Header.Add("Accept", "application/vnd.heroku+json; version=3")
+	req.Header.Add("Content-type", "application/json")
+	req.SetBasicAuth(os.Getenv("AUTH_USER"), os.Getenv("AUTH_TOKEN"))
+	resp, err := client.Do(req)
```
kr: If we're not setting anything on the http.Client, should we be using http.DefaultClient here instead?

cmd/testnet-reset/main.go
```diff
@@ -94,6 +98,22 @@ func main() {
 		"generator_url":          gen.BaseURL,
 		"generator_access_token": genCore.netTok,
 	}, nil))
+
+	method := "PATCH"
+	url := "https://api.heroku.com/apps/chain-testnet-info/config-vars"
+	r := strings.NewReader(`{"BLOCKCHAIN_ID":"` + resp.BlockchainID + `"}`)
+	client := http.Client{}
+	req, err := http.NewRequest(method, url, r)
```
kr: nit: I would inline method and url string here

cmd/testnet-reset/main.go
```diff
@@ -94,6 +98,22 @@ func main() {
...
+	must(err)
+	req.Header.Add("Accept", "application/vnd.heroku+json; version=3")
+	req.Header.Add("Content-type", "application/json")
+	req.SetBasicAuth(os.Getenv("AUTH_USER"), os.Getenv("AUTH_TOKEN"))
+	resp, err := client.Do(req)
+	must(err)
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	must(err)
+	os.Stdout.Write(body)
```
kr: nit: if you want, you could use io.Copy here instead of ReadAll and Write

kr:
LGTM with some small suggestions

PR #112, opened 2016-10-31
sdk/java: set default timeouts to 30 seconds

jbowens:
By default, OkHttp sets 10-second timeouts:
https://github.com/square/okhttp/blob/okhttp_25/okhttp/src/main/java/com/squareup/okhttp/OkHttpClient.java#L190


kr:
LGTM

PR #113, opened 2016-10-31
protocol: WaitForBlock funcs return chans

erykwalder:
This allows callers to select over the call without needing to introduce
a helper function.

### DIFF
--- a/protocol/protocol.go
+++ b/protocol/protocol.go
@@ -208,35 +208,39 @@ func (c *Chain) AddBlockCallback(f BlockCallback) {
 	c.blockCallbacks = append(c.blockCallbacks, f)
 }
 
-// WaitForBlockSoon waits for the block at the given height,
+// WaitForBlockSoon returns a channel that
+// waits for the block at the given height,
 // but it is an error to wait for a block far in the future.
-// WaitForBlockSoon will timeout if the context times out.
 // To wait unconditionally, the caller should use WaitForBlock.
-func (c *Chain) WaitForBlockSoon(ctx context.Context, height uint64) error {
-	const slop = 3
-	if height > c.Height()+slop {
-		return ErrTheDistantFuture
-	}
+func (c *Chain) WaitForBlockSoon(height uint64) chan error {
+	ch := make(chan error)
 
-	done := make(chan struct{}, 1)
 	go func() {
-		c.WaitForBlock(height)
-		done <- struct{}{}
+		const slop = 3
+		if height > c.Height()+slop {
+			ch <- ErrTheDistantFuture
+			return
+		}
+
+		<-c.WaitForBlock(height)
+		ch <- nil
 	}()
 
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-done:
-		return nil
-	}
+	return ch
 }
 
-// WaitForBlock waits for the block at the given height.
-func (c *Chain) WaitForBlock(height uint64) {
-	c.state.cond.L.Lock()
-	defer c.state.cond.L.Unlock()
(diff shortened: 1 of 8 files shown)


erykwalder:
PTAL

I'm not feeling super strong about pulling out the context.

kr:
So it's about 10 lines more code after this change?

erykwalder:
If we pull out the context, yeah

kr:
Is it better if we didn't pull out the context?

erykwalder:
Yes, I'll undo those parts of the change

erykwalder:
Updated, PTAL

protocol/protocol.go
```diff
@@ -208,35 +208,44 @@ func (c *Chain) AddBlockCallback(f BlockCallback) {
...
-func (c *Chain) WaitForBlock(height uint64) {
-	c.state.cond.L.Lock()
-	defer c.state.cond.L.Unlock()
-	for c.state.height < height {
-		c.state.cond.Wait()
-	}
+// WaitForBlock returns a channel that
+// waits for the block at the given height.
+func (c *Chain) WaitForBlock(height uint64) chan struct{} {
+	ch := make(chan struct{})
```
kr: This needs to be buffered so we don't leak a goroutine in the case the caller selects a different channel.

protocol/protocol.go
```diff
@@ -208,35 +208,44 @@ func (c *Chain) AddBlockCallback(f BlockCallback) {
...
 // but it is an error to wait for a block far in the future.
 // WaitForBlockSoon will timeout if the context times out.
 // To wait unconditionally, the caller should use WaitForBlock.
-func (c *Chain) WaitForBlockSoon(ctx context.Context, height uint64) error {
-	const slop = 3
-	if height > c.Height()+slop {
-		return ErrTheDistantFuture
-	}
+func (c *Chain) WaitForBlockSoon(ctx context.Context, height uint64) chan error {
+	ch := make(chan error)
```
kr: This needs to be buffered so we don't leak a goroutine in the case the caller selects a different channel.

protocol/protocol.go
```diff
@@ -208,35 +208,44 @@ func (c *Chain) AddBlockCallback(f BlockCallback) {
...
+// waits for the block at the given height,
 // but it is an error to wait for a block far in the future.
 // WaitForBlockSoon will timeout if the context times out.
 // To wait unconditionally, the caller should use WaitForBlock.
-func (c *Chain) WaitForBlockSoon(ctx context.Context, height uint64) error {
-	const slop = 3
-	if height > c.Height()+slop {
-		return ErrTheDistantFuture
-	}
+func (c *Chain) WaitForBlockSoon(ctx context.Context, height uint64) chan error {
```
kr: Can we specify the direction on the return type? `<-chan error`

protocol/protocol.go
```diff
@@ -208,35 +208,44 @@ func (c *Chain) AddBlockCallback(f BlockCallback) {
...
-// WaitForBlock waits for the block at the given height.
-func (c *Chain) WaitForBlock(height uint64) {
-	c.state.cond.L.Lock()
-	defer c.state.cond.L.Unlock()
-	for c.state.height < height {
-		c.state.cond.Wait()
-	}
+// WaitForBlock returns a channel that
+// waits for the block at the given height.
+func (c *Chain) WaitForBlock(height uint64) chan struct{} {
```
kr: Can we specify the direction on the return type? `<-chan struct{}`

erykwalder:
Buffered and directed. PTAL

kr:
LGTM

PR #114, opened 2016-10-31
core/account: elevate locking mode to avoid deadlock

bobg:


bobg:
rebased

jbowens:
lgtm, can we merge this to avoid deadlocks in the near term? we can work on fixing performance after

bobg:
rebased

PR #115, opened 2016-10-31
core: expose missing asset/account IDs in actions

dominic:
Replace "not found" errors for blank asset and account IDs with errors indicating
that enough information to locate an object was not provided.

### DIFF
--- a/core/account/builder.go
+++ b/core/account/builder.go
@@ -43,6 +43,14 @@ type spendAction struct {
 }
 
 func (a *spendAction) Build(ctx context.Context, maxTime time.Time) (*txbuilder.BuildResult, error) {
+	if a.AccountID == "" {
+		return nil, txbuilder.ErrActionMissingAccountID
+	}
+
+	if a.AssetID == (bc.AssetID{}) {
+		return nil, txbuilder.ErrActionMissingAssetID
+	}
+
 	acct, err := a.accounts.findByID(ctx, a.AccountID)
 	if err != nil {
 		return nil, errors.Wrap(err, "get account info")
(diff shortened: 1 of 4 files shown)


dominic:
PTAL

jeffomatic:
Probably need one for `spentUTXOAction#Build` as well

jeffomatic:
@kr Good question. There may be some value-add we can perform on the client side (in particular dashboard) if we have codified action error codes?

kr:
Sure, sounds good. (Don't let me block this PR.)

dominic:
@kr Yeah, I'd like to be able to address the missing field individually. I think an alternative would be creating a new `missing` error, and use the detail message as a comma separate list of required field names that need to be provided.

dominic:
@kr An alternative alternative is to change the way dashboard sends fields. Currently, if an input is on the screen, it gets sent in the API request even if its empty. This leads to some unexpected parsing in Go (e.g. asset alias `""` pareses to asset ID `000000000...`). Does the API already enumerate required fields for an action in an error message if fields are omitted instead of sent as empty strings?

kr:
> I think an alternative would be creating a new missing error, and use the detail message as a comma separate list of required field names that need to be provided.

Oh I like this idea. It'll easily grow to accommodate all current and future field names in all requests. We even have a place to put structured data in error responses now, so we don't have to cram everything into the (ostensibly human-readable) detail string.

I'll whip up a PR to see how it looks.

dominic:
Okay great - I'll let this hang out until I see yours.

dominic:
Closing in favor of https://github.com/chain/chain/pull/127

PR #116, opened 2016-10-31
cmd/benchcore: set MAXDBCONNS, fix m3.xlarge ami

jbowens:
Set MAXDBCONNS to 100. Also, fix the m3.xlarge ami to support EBS.


jbowens:
PTAL

erykwalder:
LGTM

PR #117, opened 2016-11-01
core/pin: add block processors

erykwalder:
This removes blocking callbacks from the protocol's commit block process
in favor of separate block processors that pull blocks. The processors
maintain pins that keep track of the last block that has been processed.
In the case of the cored process being restarted, the processors will
pick up where they last left off.

The advantage of separating block processors is that it allows the
validation and application of blocks to proceed unblocked by secondary
computations, such as indexing account utxos. It also paves the way for
some aspects of block processing to happen outside of the leader
process.


kr:
:eyes:

core/processor/processor.go
```diff
@@ -0,0 +1,122 @@
...
+				log.Error(ctx, err)
+				continue
+			}
+			err = cb(ctx, block)
+			if err != nil {
+				log.Error(ctx, err)
+				continue
+			}
+			// This could cause issues, since it is not inside of a
+			// database transaction.
```
kr: I think we should try to write the update with a where clause that ensures the value never decreases. We're already assuming that the callback is idempotent. Would that solve all the issues?

core/processor/processor.go
```diff
@@ -0,0 +1,122 @@
...
+	"context"
+	"sync"
+
+	"chain/database/pg"
+	"chain/log"
+	"chain/protocol"
+	"chain/protocol/bc"
+)
+
+type CursorStore struct {
```
kr: The naming stuff we discussed offline: it would be cool if this could convey some of the notion that it's not safe to prune block data for the block the cursor is on or any subsequent block. (It may or may not be safe to prune earlier blocks.) I was thinking "pin" might work for that. So perhaps we would have types `pin.Store` and `pin.Pin`?

core/processor/processor.go
```diff
@@ -0,0 +1,122 @@
...
+	return c
+}
+
+func (c *Cursor) Height() uint64 {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.height
+}
+
+func (c *Cursor) Increment(ctx context.Context) error {
```
kr: I think this should take the new value (and maybe have a different name (Update?) then).

core/processor/processor.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (c *Cursor) WaitForHeight(height uint64) <-chan struct{} {
+	ch := make(chan struct{}, 1)
+	go func() {
+		c.mu.Lock()
+		for c.height < height {
+			c.cond.Wait()
+		}
+		c.mu.Unlock()
```
kr: I'm pretty sure Wait never panics, but it would still be safer/more idiomatic to put this Unlock call in a defer stmt before the loop.

core/processor/processor.go
jbowens: If we go with the 'Pin' naming, it could be 'Move'?

core/pin/pin.go
```diff
@@ -0,0 +1,123 @@
...
+		return err
+	}
+	if height > c.height {
+		c.height = height
+		c.cond.Broadcast()
+	}
+	return nil
+}
+
+func (c *Pin) WaitForHeight(height uint64) <-chan struct{} {
```
jbowens: nit: maybe `p` now?

protocol/protocol.go
```diff
@@ -106,8 +104,7 @@ type Chain struct {
 	InitialBlockHash  bc.Hash
 	MaxIssuanceWindow time.Duration // only used by generators
 
-	blockCallbacks []BlockCallback
```
jbowens: hallelujah

erykwalder:
So testbot will continue to be a heisentest until we wait for the tx processor to finish. However, since this currently only runs on the leader, we'd have to introduce cross-process updating of the pins (probably using pq notify/listen). Should I go ahead and do that in this PR?

erykwalder:
Went ahead and added postgres listen/notify setup for pins. PTAL

jbowens:
👀

boymanjor:
Should we change the integration test?

PR #119, opened 2016-11-01
Entry-based transactions (DRAFT)

oleganza:
This is a draft PR inroducing new transaction structure based on _entries_ rather than _inputs_ and _outputs_.

Rationale: 

1. This resolves naming issues nicely. Outputs and inputs are separated from issuance entries and retirement entries. Non-witness content is called _content_. We no longer have "input commitments" and "output commitments".
2. Removed redundancy in definitions of reference data and witness fields. Every entry has extensible content and extensible witness.
3. Top-level transaction reference data can be plural - via special-type _reference entries_.
4. Future extensions to the protocol are much easier to introduce without a need to fit everything into input or output fields. Common commitment / witness fields are gone. Entries are abstract enough to allow any sort of extension, be it Bitcoin-style (inputs/outputs) or Ethereum-style (a piece of code that changes the blockchain state).
5. VM introspection opcodes are cleaned up and as a side-effect provide introspection to both inputs and outputs. CHECKOUTPUT, TXREFDATAHASH, REFDATAHASH are replaced by a single ENTRY opcode. Witness data is not exposed to the VM, other than by initializing current entry's stack with this entry's program arguments.

### DIFF
--- a/core/api-spec.md
+++ b/core/api-spec.md
@@ -1019,6 +1019,7 @@ POST /info
   "generator_block_height": <integer>,
   "is_production": <boolean>,
   "network_rpc_version": <integer>,
+  "core_id": <string>,
   "build_commit": <string>,
   "build_date": <string>
 }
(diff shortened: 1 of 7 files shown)


bobg:
The spec should make clear that all entries - inputs, outputs, and others - share the same numbering beginning at zero, and that entries of different types may be interleaved with one another. This is in contrast to earlier versions in which there's an input 0 and an output 0, etc.

PR #120, opened 2016-11-01
cmd/benchcore: raise open file limit

jbowens:
We hit the open file limit if we try to do > 1000 txs/sec. Raise the limit on the cored machine.

### DIFF
--- a/cmd/benchcore/main.go
+++ b/cmd/benchcore/main.go
@@ -765,13 +765,11 @@ $HOME/corectl create-token -net benchcorenet > $HOME/network-token.txt
 
 const coredsh = `#!/bin/bash
 set -eo pipefail
-sudo bash<<EOFSUDO
-ulimit -n 65535
+sudo bash -c "ulimit -n 65535 && exec su ubuntu"
 export DATABASE_URL='{{dbURL}}'
 export MAXDBCONNS=100
 export GOTRACEBACK=crash
 ./cored 2>&1 | tee -a cored.log
-EOFSUDO
 `
 
 const clientsh = `#!/bin/bash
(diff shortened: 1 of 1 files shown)


bobg:
This PR has no description, and the title belies the biggest change here, which is to run `cored` under `sudo`. Is it possible to avoid that? If not, please document it.

jbowens:
This a benchmark, not a production server ;)

It is likely possible, but I don't know how. @kr and I paired on this and he might have an idea.

ryandotsmith:
I bet we only need `sudo` for ulimit.

jbowens:
I think ulimit only applies to the current shell, so as soon as we exit the inner shell,  we lose it. We might need to exec something as the original user.

kr:
Good call @bobg, we should switch back to an unprivileged user after the ulimit. Maybe like

```
sudo bash <<EOFROOT
ulimit -n 65535

sudo -u ubuntu bash <<EOFUBUNTU
export DATABASE_URL='{{dbURL}}'
export MAXDBCONNS=100
export GOTRACEBACK=crash
./cored 2>&1 | tee -a cored.log
EOFUBUNTU

EOFROOT
```

erykwalder:
Another option to one-line it:
`sudo bash -c "ulimit -n 65535 && exec su ubuntu"`

cmd/benchcore/main.go
```diff
@@ -765,6 +765,7 @@ $HOME/corectl create-token -net benchcorenet > $HOME/network-token.txt
 
 const coredsh = `#!/bin/bash
 set -eo pipefail
+sudo bash -c "ulimit -n 65535 && exec su ubuntu"
```
kr: This line doesn't do anything useful. It starts a bash process, which changes its own ulimit, then execs a new process as a different user, which does nothing and then exits.

jbowens:
PTAL

kr:
LGTM

PR #121, opened 2016-11-01
core: handle panics in batch endpoints

jbowens:
Previously, panics in batch endpoints were never recovered and would crash the process. This adds a `batchRecover` function that for recovering panics and converting errors into errorInfo structs.

### DIFF
--- a/core/accounts.go
+++ b/core/accounts.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 
 	"chain/core/signers"
+	"chain/net/http/reqid"
 )
 
 // This type enforces JSON field ordering in API output.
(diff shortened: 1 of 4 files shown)


erykwalder:
I also know previously we had wanted explicit return types because auditing our responses was much more difficult because of the number of `interface{}` return types.

jbowens:
> I also know previously we had wanted explicit return types because auditing our responses was much more difficult because of the number of interface{} return types.

Erf, not sure what to do about that. We might need to duplicate the recover logic in each handler if we want to keep the typed return responses.

kr:
I think we can probably write a generic error handler for panics that will work for them all. Something along the lines of

```
func batchRecover(v *interface{})
```

Then it won't be too hard to put

```
defer batchRecover(&result)
```

in all of the batch handlers.

jbowens:
updated to not wrap the closure, but still struggling to keep the `assetOrError` type

jbowens:
rebased

kr:
LGTM. We can figure out logging nested errors and logging error data items later.

PR #123, opened 2016-11-01
core/account: split utxo/reservation mapping into its own db table

bobg:
Explicit locking is for the birds.

### DIFF
--- a/core/account/indexer.go
+++ b/core/account/indexer.go
@@ -7,7 +7,6 @@ import (
 
 	"chain/core/signers"
 	"chain/database/pg"
-	"chain/database/sql"
 	"chain/encoding/json"
 	"chain/errors"
 	"chain/protocol/bc"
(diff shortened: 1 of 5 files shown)


bobg:
PTAL

bobg:
Remains to be seen whether this change is better with https://github.com/chain/chain/pull/123/commits/17f61e14fe00b71c1511911733926121424793be or without it.

core/account/indexer.go
```diff
@@ -107,14 +106,7 @@ func (m *Manager) indexAccountUTXOs(ctx context.Context, b *bc.Block) error {
 	}
 	defer dbtx.Rollback(ctx)
```
bobg: Oops, thanks.

bobg:
> Remains to be seen whether this change is better with https://github.com/chain/chain/pull/123/commits/17f61e14fe00b71c1511911733926121424793be or without it.

With, by a factor of about 2x and many many more rejected txs due to double spending.

PR #124, opened 2016-11-01
dashboard: expose tx errors on individual actions

dominic:


dominic:
PTAL @jeffomatic

dominic:
screen shot 2016-11-01 at 11 58 33 am

dashboard/src/chain/errors.js
```diff
@@ -1,10 +1,12 @@
 const lib = {
   create: function(type, message, props = {}) {
     let err = new Error(message)
-    err = Object.assign(err, props, {
+    err = {
+      ...err,
+      ...props.data,
```
jeffomatic: Just by eyeballing it, I don't know if this will work.

- I think we want to do `Object.assign` here so it preserves the `Error` constructor and associated prototype.
- `props` looks like it's being used by the caller to inject arbitrary data into the error object. In other words, it is not the body of the error response. I think we have to update the request method itself to do that: https://github.com/chain/chain/blob/main/dashboard/src/chain/client.js#L89

dominic: Okay, I can walk that back easily enough.

dominic:
Updated - PTAL again

PR #125, opened 2016-11-01
errors: make WithData merge maps

kr:
This lets us accumulate data in an error rather than
entirely replacing what was there before.

This is the first step in an alternate approach to #115.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -138,7 +138,7 @@ func withData(err error, v map[string]interface{}) error {
 // Note that if err already has a data item of any other type,
 // it will not be accessible via the returned error value.
 func WithData(err error, keyval ...interface{}) error {
-	// TODO(kr): add vet check for odd-length keyval
+	// TODO(kr): add vet check for odd-length keyval and non-string keys
 	newkv := make(map[string]interface{})
 	for k, v := range Data(err) {
 		newkv[k] = v
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -15,7 +15,7 @@ func New(text string) error {
 type wrapperError struct {
 	msg    string
 	detail []string
-	data   interface{}
+	data   map[string]interface{}
 	stack  []StackFrame
 	root   error
 }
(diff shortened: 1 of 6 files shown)


errors/errors.go
```diff
@@ -126,6 +126,29 @@ func WithData(err error, v interface{}) error {
...
+// Note that if err already has a data item of any other type,
+// it will not be accessible via the returned error value.
+func WithDataKV(err error, keyval ...interface{}) error {
+	newkv := make(map[string]interface{})
+	oldkv, _ := Data(err).(map[string]interface{})
+	for k, v := range oldkv {
+		newkv[k] = v
+	}
+	for i := 0; i < len(keyval); i += 2 {
+		newkv[keyval[i].(string)] = keyval[i+1]
```
erykwalder: There's a potential panic here.

errors/errors.go
```diff
@@ -126,6 +126,29 @@ func WithData(err error, v interface{}) error {
...
+	}
+	return WithData(err, newkv)
```
erykwalder: Should `WithData` become private? It feels strange to have one method that expects cooperation and another that doesn't.

errors/errors.go
kr: Yeah, I want it to panic if it's not a string. That's why the doc says "Values kN must be strings".

errors/errors.go
kr: Hm actually I'm on the fence over this. We include a json array of action errors as the top level data item. We could just make all data items be maps, but that would add one more nesting layer to an already annoyingly-nested structure.

What do you think?

errors/errors.go
erykwalder: Also from an odd number of args.

errors/errors.go
erykwalder: I guess I'm just worried about accidentally clobbering. Using WithDataKV after WithData will clobber, as will the other way around.

erykwalder: But if this is what we've got for now, that's fine.

kr: Ok, I'll make them all be maps.

erykwalder: Actually, perhaps we could panic if data is not nil or not a map.

errors/errors.go
kr: Same thing for odd number of args. It should panic.
I added a TODO to check for this in vet.

kr:
PTAL

kr:
PTAL @erykwalder @dominic

jbowens:
lgtm

PR #126, opened 2016-11-01
dashboard: restore balance sum_by filter

dominic:
This addresses two issues with the `sum_by` paramemter:

- the sum_by filter was not sent with the filter submission
- the sum_by filter was ignored when a default filter was present


dominic:
PTAL

dashboard/src/features/shared/components/SearchBar/SearchBar.jsx
```diff
@@ -83,24 +83,24 @@ class SearchBar extends React.Component {
...
-    } else if (this.props.defaultFilter) {
+    if (this.state.sumBy) query.sum_by = this.state.sumBy
+    if (this.props.defaultFilter) {
       state.query = this.props.defaultFilter
       query.filter = this.props.defaultFilter
     }
-    if (this.state.sum_by) query.sum_by = this.state.sumBy
+
+    if (this.state.query == this.props.defaultFilter) {
+      this.setState({ showClear: false })
```
jeffomatic: It seems like you can't clear the sum-by field if the filter is equal to the default, but those two should be separate, right? Maybe we can get rid of this entire statement if we change the `state` object to initialize to the following:

```
     const state = {
        showClear: (this.state.query && (this.state.query != this.props.defaultFilter)) || this.state.sumBy
      }
```

jeffomatic:
Note for future selves: we should make a TODO to eventually refactor the naming and state management of this component...it may make it easier for us to reason about if we can somehow put the state in one canonical location, as opposed to inside both props and local state.

dominic:
PTAL again @jeffomatic

PR #127, opened 2016-11-01
core/txbuilder: report missing field names

kr:
This adds a new txbuilder error ErrMissingFields, and
updates account and asset builders to return that error
along with a list of missing field names.

Depends on #125.

### DIFF
--- a/core/account/builder.go
+++ b/core/account/builder.go
@@ -46,7 +46,8 @@ func (a *spendAction) Build(ctx context.Context, maxTime time.Time) (*txbuilder.
 	var missing []string
 	if a.AccountID == "" {
 		missing = append(missing, "account_id")
-	} else if a.AssetID == (bc.AssetID{}) {
+	}
+	if a.AssetID == (bc.AssetID{}) {
 		missing = append(missing, "asset_id")
 	}
 	if len(missing) > 0 {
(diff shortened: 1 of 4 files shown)


core/account/builder.go
```diff
@@ -43,6 +43,16 @@ type spendAction struct {
 }
 
 func (a *spendAction) Build(ctx context.Context, maxTime time.Time) (*txbuilder.BuildResult, error) {
+	var missing []string
+	if a.AccountID == "" {
+		missing = append(missing, "account_id")
+	} else if a.AssetID == (bc.AssetID{}) {
```
kr: This shouldn't be in an else branch.

core/account/builder.go
```diff
@@ -191,6 +201,16 @@ type controlAction struct {
 }
 
 func (a *controlAction) Build(ctx context.Context, maxTime time.Time) (*txbuilder.BuildResult, error) {
+	var missing []string
+	if a.AccountID == "" {
+		missing = append(missing, "account_id")
+	} else if a.AssetID == (bc.AssetID{}) {
```
kr: This shouldn't be in an else branch.

kr:
rebased

kr:
PTAL @dominic

kr:
rebased

PR #129, opened 2016-11-01
docs/tools/checklinks.rb: fix reports for broken cross-file references

oleganza:
Lets say you have file `alice.md` with a cross-file link `[link](bob.md#bad-anchor)`. 

Previous version of this tool was reporting:

    Broken link in bob.md: [link (from alice.md)](#bad-anchor)

Since usually it's the link that is broken, not the title, checklinks.rb now reports:

    Broken link in alice.md: [link](bob.md#bad-anchor)

Which is much more useful: we know that we have to go to `alice.md` and fix the link there.

### DIFF
--- a/docs/tools/checklinks.rb
+++ b/docs/tools/checklinks.rb
@@ -1,5 +1,7 @@
 #!/usr/bin/env ruby
 require 'open-uri'
+require 'pathname'
+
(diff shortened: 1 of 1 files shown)


dominic:
LGTM

PR #131, opened 2016-11-01
docs: add development instructions to readme

dominic:


dominic:
PTAL

oleganza:
@dominic WDYT of having this, friendlier command to run from $CHAIN:

    $ go run ./cmd/md2html/*.go

ryandotsmith:
@dominic do we need the TOC in the readme?
Also, it would be helpful to also describe how we deploy documentation. Imagine a case where someone would like to change the docs. What are the exact steps required to get them deploy to chain.com/docs and optionally included in cored.

kr:
Please avoid using 'go run' with a list of .go files for stable tools like this. It's slow and fragile (it doesn't respect build constraints, so if we ever have to add a build tag or platform-specific code, it'll break).

It's much better to `go install chain/cmd/md2html` so you get a true Unix command, and from then on run the command.

dominic:
@kr :+1: I think that should def be our policy for all READMEs then

kr:
Ok. Oh also I just caught up on some of the slack history. If it's common to be editing md2html along with the docs, then sure 'go run' can be more convenient than 'go install && md2html', but in that case I think it deserves to be called out in the readme why we suggest doing that. Because it's the exception rather than the rule.

dominic:
PTAL again @ryandotsmith @kr

PR #133, opened 2016-11-02
core/config: new package

jbowens:
Move `core.Config` and related functions into a new package, `core/config`.
This allows corectl to not import `chain/core` and by extension the
dashboard & documentation.

The `corectl` binary size is reduced from 52M to 15M.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -11,8 +11,8 @@ import (
 	"strconv"
 	"time"
 
-	"chain/core"
 	"chain/core/accesstoken"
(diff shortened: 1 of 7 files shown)


jbowens:
PTAL

oleganza:
Personal thank you!

bobg:
LGTM

The `corectl` binary size is a great side benefit; I think the main benefit is that this is a natural package boundary. I've gone looking for the `core/config` package myself a number of times only to remember the config stuff was buried in `core`.

PR #134, opened 2016-11-02
core/query: fix block processor race

### LINKED ISSUE #76: core/query: race condition annotating remote issuances
There's a race condition when a block lands that includes a transaction issuing a new, remote asset. Since block indexing callbacks run in parallel now, it's possible for the `core/query` callback to annotate the transaction before the `core/asset` package has recorded the new asset in the database. The resulting annotated transaction will be missing the asset definition.

Transactions using the asset in any subsequent blocks will be correctly annotated with the asset definition.

Possible solutions:
1. Don't run block callbacks in parallel. This would impact performance.
2. Include the block height when calling out to the asset package to annotate transactions. In the asset package's annotation function, explicitly wait until the asset package's callback has finished indexing the block at the provided height.
3. When annotating a transaction, if the asset isn't indexed in the database, manually pull the asset definition out of the input's issuance program. We'll also need to keep it around through the duration of annotating the block, because other inputs/outputs/transactions within the block may also refer to the new asset.

erykwalder:
Previously, if the transaction index processor completed before the
asset processor, the transaction index would be missing data. This
change introduces a wait that blocks until the asset processor has
finished the given block.

Closes #76

### DIFF
--- a/core/query/index.go
+++ b/core/query/index.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/lib/pq"
 
+	"chain/core/asset"
 	"chain/database/pg"
 	"chain/errors"
 	"chain/protocol/bc"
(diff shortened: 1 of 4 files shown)


jbowens:
Can add "Closes #76" to your commit message

erykwalder:
Tests passing, PTAL

jbowens:
lgtm

PR #135, opened 2016-11-02
core/pin: remove racy Listen test

erykwalder:
There is no way to guarantee that this test will pass, since notify may
be called before Listen is ready.


kr:
LGTM

PR #136, opened 2016-11-02
core/coreunsafe: truncate block_processors on reset

erykwalder:
This prevents block processors from ignoring blocks after a reset.


kr:
LGTM

PR #137, opened 2016-11-02
core/coreunsafe: remove explicit truncate list

erykwalder:
Rather than having a maintained list of tables that are truncated, it is
more simple to maintain a list of tables which should not be reset. This
change makes it so that all tables are truncated unless they are
contained in the persistBlockchainReset or neverReset lists.


kr:
LGTM

PR #140, opened 2016-11-02
metrics: save space with fewer significant figures

kr:
We care about the difference between 54ms and 55ms or
230ms and 240ms, but not really between 54ms and 54.1ms
or 230ms and 231ms. Storing only two significant figures
instead of three uses less space. In our case, it uses
about 15% as much space (860kB instead of 5.8MB). This
matters more for efficiency of sampling and storing
metrics than for the performance of cored itself.


erykwalder:
LGTM

PR #141, opened 2016-11-02
core/account: remove utxo from SpendAction

jbowens:
Remove the transaction hash and index parameters from the account spend
action. Now that we have the SpendUTXOAction, we no longer need to be
able to filter account UTXOs by transaction hash.

### DIFF
--- a/core/account/builder.go
+++ b/core/account/builder.go
@@ -14,12 +14,10 @@ import (
 	"chain/protocol/bc"
 )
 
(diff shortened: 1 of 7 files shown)


jbowens:
PTAL

erykwalder:
LGTM as long as we don't need the ability to spend from tx without specifying position.

jeffomatic:
> LGTM as long as we don't need the ability to spend from tx without specifying position

We don't...this feature was originally added for a partner project, and we didn't need to specify position for that situation. The SpendFromAccountUnspentOutput action should handle that case now.

PR #142, opened 2016-11-02
core: wait until transaction block is processed

erykwalder:
After introducing block processors, the submit call would return before
the block that contained the transaction had been processed by all
processors. Since this was a change of behavior and since it could cause
confusion when other requests are made before processing, it has been
changed to wait for processing by default. Since finer grain control is
possible, an additional field "WaitFor" has been introduced that will
determine when the submit call returns.

### DIFF
--- a/core/transact.go
+++ b/core/transact.go
@@ -94,9 +94,9 @@ func (h *Handler) build(ctx context.Context, buildReqs []*buildRequest) (interfa
 }
 
 type submitSingleArg struct {
-	tpl     *txbuilder.Template
-	wait    chainjson.Duration
-	waitFor string
+	tpl       *txbuilder.Template
+	wait      chainjson.Duration
+	waitUntil string
 }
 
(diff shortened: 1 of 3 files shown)


core/transact.go
```diff
@@ -264,6 +272,7 @@ func waitForTxInBlock(ctx context.Context, c *protocol.Chain, tx *bc.Tx, height
 type submitArg struct {
 	Transactions []*txbuilder.Template
 	wait         chainjson.Duration
+	WaitFor      string `json:"wait_for"` // values none, confirmed, processed. default: processed
```
jbowens: nit: if we use the values `confirmed` and `processed` maybe this should be `wait_until`?

jbowens:
should we update the java SDK here or separately?

kr:
> should we update the java SDK here or separately?

Here would be awesome.

kr:
Oh but it's compatible right? Then w/e.

erykwalder:
SDK update included. PTAL

jbowens:
lgtm

PR #143, opened 2016-11-02
cmd/testnet-reset: restrict execution time

boymanjor:
The cmd is set up on a Heroku cron job with a max
interval of one day. This commit restricts execution
to when the cmd is initiated manually (SCHEDULED=false)
or the execution time is within an hour of Sunday at
midnight.


kr:
LGTM (with a nit)

kr:
oh second nit: s/setup/set up/

(that is s/noun/verb/)

PR #146, opened 2016-11-02
dashboard: add option to automatically generate Mock HSM keys

dominic:
Account and asset forms now default to generating new Mock HSM
keys for each entry. Generated keys use the following pattern for aliases:

```
(alias || "generated")-uuid
```

Additionally, each option for Mock HSM xpub data uses separate state,
so that switching from "Use existing Mock HSM key" to another option
will not implicitly pass that value in the subsequent API request.


jeffomatic:
HOOO BOY

dominic:
@jeffomatic PTAL

dominic:
screen shot 2016-11-02 at 4 15 05 pm

dominic:
@jeffomatic Updated

screen shot 2016-11-02 at 4 42 30 pm

dashboard/src/features/shared/actions/form.js
```diff
@@ -34,10 +31,27 @@ export default function(type, options = {}) {
...
+          promise = promise
+            .then(() => {
+              const alias = (key.value || '').trim()
+                ? key.value.trim()
+                : (data.alias || 'generated') + '-' + uuid.v4()
 
-        return object.create(context())
+              return new chain.MockHsm({alias}).create(context())
+            }).then(newKey => {
+              data.root_xpubs = [...(data.root_xpubs || []), newKey.xpub]
```
jeffomatic: Might as well initialize `data.root_xpubs` above the loop, where `const xpubs` is getting initialized. That will clean up the append statements here.

dashboard/src/components/Common/XpubField/XpubField.jsx
```diff
@@ -23,11 +28,27 @@ class XpubField extends React.Component {
...
+    const {
+      typeProps,
+      valueProps,
+      mockhsmKeys,
+    } = this.props
+
+    const typeOnChange = event => {
+      const value = typeProps.onChange(event).value
+
+      typeProps.onChange(value)
```
jeffomatic: Can you explain what's going on here? It looks like we're calling `typeProps.onChange()` twice, and passing its return value from the first call as an argument to the second call.

dashboard/src/components/Common/XpubField/XpubField.jsx
```diff
@@ -39,29 +60,46 @@ class XpubField extends React.Component {
...
-        {this.state.selectedType == 'provide' &&
-          <TextField key={this.props.index} fieldProps={this.props.fieldProps} />}
+        {typeProps.value == 'provide' &&
+          <TextField
+            fieldProps={{...valueProps, onChange: valueOnChange}}
+            placeholder='Enter Xpub' />}
+
+        {typeProps.value == 'generate' &&
+          <TextField
+            fieldProps={{...valueProps, onChange: valueOnChange}}
```
jeffomatic: Since we are re-using `valueProps` here, won't this propagate the value of the text field to the other options' text fields if the radio option changes?

jeffomatic: One alternative may be to treat each text field as its own prop, and customize the post-submit-click behavior based on which radio option is selected.

dominic: It doesn't propoagte because the `type` and `value` `onChange` functions work in tandem to sync with the component state for each option individually.

dashboard/src/components/Common/XpubField/XpubField.jsx
dominic: That looks like a typo/bad copy paste.

dashboard/src/features/shared/actions/form.js
```diff
@@ -15,14 +16,10 @@ export default function(type, options = {}) {
     created,
     submitForm: (data) => {
       const className = options.className || type.charAt(0).toUpperCase() + type.slice(1)
+      let promise = new Promise((resolve) => resolve())
```
dominic: We can't remove the temporary because we want to optionally chain the Xpub calls and ensure they run before the object creation.

dashboard/src/features/shared/actions/form.js
```diff
@@ -34,10 +31,27 @@ export default function(type, options = {}) {
         data[fieldName] = parseInt(data[fieldName])
       })
 
-      return function(dispatch) {
-        let object = new chain[className](data)
+      const xpubs = data.xpubs || []
+      xpubs.map(key => {
```
dominic: Agreed - it's more like `/features/shared/actions/create.js`

dominic:

PR #147, opened 2016-11-03
cmd/testnet-reset: turn off bot before resetting

boymanjor:
The chain-core-ccte Heroku app adds a light load to the 
testnet. This app needs to be shut down before resetting
the network. Once the network is reset, the app's config
vars should be updated and it should be restarted.


kr:
LGTM

PR #148, opened 2016-11-03
cmd/sha3: remove redundant info in help output

kr:
Package flag prints the default value on the end of the
line, so we don't need to say it again. This also makes
the other wording a bit more consise.

### DIFF
--- a/cmd/sha3/sha3.go
+++ b/cmd/sha3/sha3.go
@@ -11,7 +11,7 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


kr:
Before:

```
; sha3 -h
Usage of sha3:
  -n int
    	size in bits of the desired hash: 224, 256 (default), 384, or 512 (default 256)
```

After:

```
; sha3 -h
Usage of sha3:
  -n bits
    	output size in bits: 224, 256, 384, or 512 (default 256)
```

kr:
PTAL

PR #149, opened 2016-11-03
core/pin: start new pin at chain height

erykwalder:
This prevents the pin from blocking on heights that may not be present,
or no longer relevant.

### DIFF
--- a/core/generator/generator.go
+++ b/core/generator/generator.go
@@ -51,7 +51,6 @@ func Generate(
 	db pg.DB,
 	period time.Duration,
 	health func(error),
-	networkReady func(),
 ) {
 	// This process just became leader, so it's responsible
 	// for recovering after the previous leader's exit.
@@ -59,8 +58,6 @@ func Generate(
 	if err != nil {
 		log.Fatal(ctx, log.KeyError, err)
 	}
-
-	networkReady()
 
 	g := &generator{
 		db:             db,
--- a/core/generator/generator.go
+++ b/core/generator/generator.go
@@ -51,6 +51,7 @@ func Generate(
 	db pg.DB,
 	period time.Duration,
 	health func(error),
+	networkReady func(),
 ) {
 	// This process just became leader, so it's responsible
 	// for recovering after the previous leader's exit.
@@ -58,6 +59,8 @@ func Generate(
 	if err != nil {
 		log.Fatal(ctx, log.KeyError, err)
 	}
+
+	networkReady()
 
 	g := &generator{
 		db:             db,
--- a/core/pin/pin.go
+++ b/core/pin/pin.go
@@ -13,31 +13,30 @@ import (
 )
 
 type Store struct {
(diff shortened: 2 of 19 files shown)


kr:
:eyes:

core/generator/generator.go
```diff
@@ -59,6 +60,8 @@ func Generate(
 		log.Fatal(ctx, log.KeyError, err)
 	}
 
+	networkReady()
```
kr: Instead of a callback here maybe we should have a `close(c.ready)` at the bottom of protocol.Chain.Recover, plus an accessor method for that chan.

So in main, we would do

```
go func() {
	<-c.Ready()
	height := c.Height()
	...
}()
```

That would avoid the extra plumbing in generator and fetch right? WDYT?

core/pin/pin.go
```diff
@@ -49,92 +48,63 @@ func (s *Store) ProcessBlocks(ctx context.Context, c *protocol.Chain, pinName st
...
-		s.mu.Unlock()
-		for _, name := range pins {
-			<-s.Pin(name).WaitForHeight(height)
+		defer s.mu.Unlock()
+		for {
+			if p, ok := s.pins[name]; ok {
+				ch <- p
+				return
+			}
+			s.cond.Wait()
```
kr: Maybe this is just me being too rigid, but I prefer to see a condition variable loop in a more idiomatic form:

```go
s.mu.Lock()
defer s.mu.Unlock()
for s.pins[name] == nil {
	s.cond.Wait()
}
c <- s.pins[name]
```

IMO saving the extra map lookup isn't worth the difference in readability.

core/pin/pin.go
```diff
@@ -146,18 +116,36 @@ func (p *Pin) WaitForHeight(height uint64) <-chan struct{} {
...
+	listener, err := pg.NewListener(ctx, dbURL, "pin-"+pinName)
 	if err != nil {
 		log.Error(ctx, err)
 		return
 	}
-
 	go func() {
 		defer func() {
 			listener.Close()
 		}()
```
kr: Sorry I missed this before, but why not `defer listener.Close()`?

core/pin/pin.go
```diff
@@ -170,6 +158,18 @@ func (p *Pin) Listen(ctx context.Context, dbURL string) {
 					return
 				}
 
+				if p == nil {
+					s.mu.Lock()
+					var ok bool
+					p, ok = s.pins[pinName]
+					if !ok {
+						p = newPin(s.db, pinName, height)
```
kr: Should newPin be called insertPin and also do the map assignment and broadcast? Dunno.

erykwalder: For now I left this

erykwalder:
PTAL

erykwalder:
Any more feedback?

kr:
LGTM

PR #150, opened 2016-11-03
cmd/testbot: report pending status to github

erykwalder:
This will add an indicator to github that there is a pending check until
it either passes or fails.


kr:
LGTM

PR #151, opened 2016-11-04
docs: fix Assets.java example

jeffomatic:
Edits:

- misspelled definition field was causing query to fail
- second list-asset example was not printing
- tx queries were looking for input.action field; should be input.type


jeffomatic:
PTAL

PR #152, opened 2016-11-04
cmd/shake: new command

kr:


kr:
PTAL @oleganza or @bobg or whoever

PR #153, opened 2016-11-04
core/leader: encapsulate leadership state

jbowens:
Return a leadership struct from `leader.Run`. This prevents the need for
callers to pass in a pg.DB to `leader.Address`. Also, it will allow
tests to write mock Leadership implementations if they need to test code
that is only run on the leader (like the in-memory reserver).

### DIFF
--- a/core/leader/leader.go
+++ b/core/leader/leader.go
@@ -13,12 +13,8 @@ import (
 	"chain/log"
 )
 
-type Leadership interface {
-	IsLeading() bool
-	Address(ctx context.Context) (string, error)
-}
-
-type leadership struct {
+// Leadership describes the leader process of Chain Core.
+type Leadership struct {
 	db      pg.DB
 	key     string
 	lead    func(context.Context)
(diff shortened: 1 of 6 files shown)


jbowens:
PTAL

core/leader/leader.go
```diff
@@ -13,18 +13,29 @@ import (
...
+}
 
-// IsLeading returns true if this process is
-// the core leader.
-func IsLeading() bool {
-	lock.Lock()
-	l := isLeading
-	lock.Unlock()
-	return l
+type leadership struct {
```
kr: Do we need the interface? It would be more conventional to export this type.

jbowens: > Do we need the interface? It would be more conventional to export this type.

The main motivation for this PR was for an interface so that this test could mock the leadership: https://github.com/chain/chain/pull/103/files#diff-319ba1ae4ee007599fec4654b2e71b6cR224

Not sure if there's a better alternative.

kr: Got it. An interface makes sense, but if possible it would be slightly better to define it in the place where it's consumed, and have this package still return concrete types.

jbowens:
updated to move interface into `chain/core`

jbowens:
rebased, PTAL

PR #158, opened 2016-11-05
docs: add Ruby documentation to bundled docs

jeffomatic:
Adds generation of Ruby docs to generate-docs command, and bundles
the results into the cored executable.

Also:
- fixes some YARD annotation bugs that was breaking generate-docs
- re-bundles the documentation

This does not add a link to the Ruby YARD documentation, which
will happen when we officially release the Ruby SDK. For now, you
can navigate to http://localhost:1999/docs/ruby/doc/index.html .


jeffomatic:
PTAL

PR #159, opened 2016-11-06
perf: only wait for confirmation

jbowens:
In perf tests, only wait until the tx is confirmed. Also, make
UtxoReservation.java a little more ambitious. I also reduced
its duration to make it a little easier to iterate quickly.


bobg:
LGTM

Issue #160, opened 2016-11-07
Tx is not "final" in certain cases


bobg:
When a transaction is constructed in multiple steps, all steps but the last one must specify "allow additional actions" to instruct the signing phase to produce signature programs that commit only to the aspects of the tx already present (so that additional actions don't invalidate the ones that have been signed). But for idempotence, and to harden against certain kinds of replay attacks, when publishing the tx [we require the presence of at least one input](https://github.com/chain/chain/blob/76a6ca04631378d869ed7bd377a64661d8fd35d7/core/txbuilder/finalize.go#L44) that commits to the tx as a whole via its hash.

But this doesn't work when the last build step involves only issuance inputs. (Exception: [when _all_ inputs are issuances](https://github.com/chain/chain/blob/76a6ca04631378d869ed7bd377a64661d8fd35d7/core/txbuilder/finalize.go#L122).) That's because the check for tx finality is looking for the specific signature program

`DATA_32 <hash> TXSIGHASH EQUAL`

but we don't use signature programs for issuances.

This is producing spurious "transaction is not final" errors.

bobg:
Correction: we _do_ use signature programs for issuances, [we simply don't check them](https://github.com/chain/chain/blob/76a6ca04631378d869ed7bd377a64661d8fd35d7/core/txbuilder/finalize.go#L92) in `checkTxSighashCommitment`.

PR #161, opened 2016-11-07
core/txbuilder: check issuance inputs as well as spends for tx finality

bobg:
Fixes https://github.com/chain/chain/issues/160

### DIFF
--- a/core/txbuilder/finalize_test.go
+++ b/core/txbuilder/finalize_test.go
@@ -25,6 +25,69 @@ import (
 	"chain/testutil"
 )
 
+func TestSighashCheck(t *testing.T) {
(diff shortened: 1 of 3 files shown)


bobg:
PTAL

core/txbuilder/finalize_test.go
```diff
@@ -25,6 +25,69 @@ import (
 	"chain/testutil"
 )
 
+func TestSighashCheck(t *testing.T) {
```
jbowens: is it possible to have this test in txbuilder so we don't need to export anything we wouldn't otherwise?

bobg: Not without considerable effort, since this test depends on `account` and `account` depends on `txbuilder`.

PR #163, opened 2016-11-08
docker/testbot: use race detector

jbowens:
When compiling cored for integration tests, compile with -race to enable
the race detector.


jbowens:
Not sure yet why the testbot step fails.

PR #164, opened 2016-11-08
core/migrate: remove not null constraint from txfeeds client_token

### LINKED ISSUE #145: /create-transaction-feed returns 500 error on null client_token param
Sending a null or undefined `client_token` param to the `/create-transaction-feed` endpoint will yield the following:

```
app=cored ... at=http.go:51 t=2016-11-02T23:07:29.398290512Z
status=500 chaincode=CH000 path=/create-transaction-feed
error="pq: null value in column \"client_token\" violates not-null constraint"
```

This is the client's problem, not the server's, so `CH000` isn't quite right here. The endpoint should either tolerate a blank idempotency token (recommended), or return a 400-class error.

tessr:
Resolves #145.

### DIFF
--- a/core/migrate/data.go
+++ b/core/migrate/data.go
@@ -30,4 +30,5 @@ var migrations = []migration{
 			height bigint DEFAULT 0 NOT NULL
 		);
 	`},
+	{Name: "2016-11-07.0.core.remove-client-token-not-null.sql", SQL: "ALTER TABLE txfeeds ALTER COLUMN client_token DROP NOT NULL;\n"},
 }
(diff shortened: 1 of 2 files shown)


tessr:
PTAL

jeffomatic:
Do we still dump the full schema?

jbowens:
yeah, we do `bin/dumpschema`

tessr:
This brave new golang migration world is too much for me.

tessr:
P

tessr:
PTAL

jeffomatic:
LGTM

PR #165, opened 2016-11-08
core/pin: enable multi-process block processing

erykwalder:
None of the block processors need to be processed in sequence, or by
only the leader. This introduces a queue with jobs that allows any
process to perform the block processing work.


erykwalder:
PTAL

This does also feel a bit more complicated than I would like, so I'm also fine scrapping it in favor of a simpler solution.

PR #166, opened 2016-11-08
protocol, core/pin: WaitForX -> XWaiter

bobg:
The functions `WaitForBlock`, `WaitForBlockSoon`, `WaitForPin`, and `WaitForAll` did not do any actual waiting, so rename them to indicate that waiting can be done with the values they return.


bobg:
PTAL

jbowens:
lgtm

PR #167, opened 2016-11-08
core: cache account and asset aliases

jbowens:
Filtering aliases is the source of > 50% of all SQL queries in one of my
benchmarks.

### DIFF
--- a/core/account/accounts.go
+++ b/core/account/accounts.go
@@ -28,10 +28,11 @@ var ErrDuplicateAlias = errors.New("duplicate account alias")
 
 func NewManager(db *sql.DB, chain *protocol.Chain) *Manager {
 	return &Manager{
-		db:     db,
-		chain:  chain,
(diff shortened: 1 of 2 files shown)


jbowens:
PTAL

bobg:
:eyes:

bobg:
This looks LGTM but if filtering aliases is as common as you say, it is probably worth trying to combine the alias->id lookup with reading of the actual account or asset data via a join (in addition to caching), to avoid a separate db roundtrip. Maybe give that a try?

jbowens:
> This looks LGTM but if filtering aliases is as common as you say, it is probably worth trying to combine the alias->id lookup with reading of the actual account or asset data via a join (in addition to caching), to avoid a separate db roundtrip. Maybe give that a try?

Since we already have caching of asset id -> asset data, I'm not too worried about it. We can follow up if it's a problem.

PR #168, opened 2016-11-09
sdk/java: use larger connection pool

jbowens:
The default OkHttp connection pool is capped at five idle connections.
The OkHttp documenation describes the default connection pool as
"appropriate for a single-user application."

Before setting this, I was frequently seeing "Cannot assign requested
address" errors while running benchmarking tests.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -14,6 +14,7 @@ import java.util.List;
 import com.google.gson.Gson;
 
 import com.squareup.okhttp.CertificatePinner;
+import com.squareup.okhttp.ConnectionPool;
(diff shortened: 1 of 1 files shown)


sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -61,8 +62,7 @@ public Client() {
     }
 
     this.url = url;
-    this.httpClient = new OkHttpClient();
```
ryandotsmith: I'm a tad confused why these two lines changed. Just out of curiosity, why were we not using `this.defaultHttpClient()` prior to this change?

jbowens: Yeah, I think that was a previous mistake.

jeffomatic:
LGTM

PR #173, opened 2016-11-09
installer/mac: remove unused dependencies and add sha256 checks

oleganza:
This removes unused dependencies from the mac app and adds sha256 digest checks to the remaining ones.

SHA256 digests for Postgres, OpenSSL and libxml2 are hardcoded in the makefile. If the digests of the downloaded archives do not match, make cannot continue building the dependencies.

### DIFF
--- a/installer/mac/pg/src/makefile
+++ b/installer/mac/pg/src/makefile
@@ -6,23 +6,15 @@ PATH=$(PREFIX)/bin:/bin:/usr/bin:/opt/local/bin
 
 POSTGRES_VERSION=9.5.4
 POSTGRES_MAJOR_VERSION=9.5
-
-JPEG_VERSION=8d
-LIBEDIT_VERSION=20130611-3.1
-
-#http://download.osgeo.org/gdal/
-LIBJASPER_VERSION=1.900.1
+POSTGRES_SHA256=cf5e571164ad66028ecd7dd8819e3765470d45bcd440d258b686be7e69c76ed0
 
 # http://xmlsoft.org/news.html
 LIBXML2_VERSION=2.9.3
+LIBXML2_SHA256=4de9e31f46b44d34871c22f54bfc54398ef124d6f7cafb1f4a5958fbcd3ba12d
 
 # https://www.openssl.org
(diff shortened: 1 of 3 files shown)


kr:
:eyes:

installer/mac/pg/src/makefile
```diff
@@ -82,15 +75,16 @@ $(SHAREDPREFIX)/lib/libssl.dylib: openssl-${OPENSSL_VERSION}/Makefile
...
-openssl-${OPENSSL_VERSION}/Configure: openssl-${OPENSSL_VERSION}.tar.gz
-	$(TAR) openssl-${OPENSSL_VERSION}.tar.gz
+openssl-${OPENSSL_VERSION}/Configure: openssl-${OPENSSL_VERSION}.tar.gz-${OPENSSL_SHA256}
+	$(TAR) openssl-${OPENSSL_VERSION}.tar.gz-${OPENSSL_SHA256}
 	touch $@
 
-openssl-${OPENSSL_VERSION}.tar.gz:
+openssl-${OPENSSL_VERSION}.tar.gz-${OPENSSL_SHA256}:
 	$(CURL) "https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz"
+	ruby ../../tools/renamesha256.rb openssl-${OPENSSL_VERSION}.tar.gz
```
kr: Feel free to use `$CHAIN` here instead of relative paths, that might be a bit more practical. (It's robust when copy-pasting code from this file to another place.) So for example

```
$CHAIN/bin/renamesha256 openssl-${OPENSSL_VERSION}.tar.gz
```

installer/mac/tools/renamesha256.rb
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env ruby
+# Renames `filename.tar.gz` into `filename.tar.gz-<sha256digest>`
```
kr: This seems like a nice, general-purpose tool. Maybe consider putting it in `$CHAIN/bin` and mark it as executable.

PR #174, opened 2016-11-10
docs/reference: add versioning doc

devongundry:


devongundry:
@jeffomatic any suggestions here?
@kr how do these package names and descriptions look? anything we're missing?

docs/core/reference/versioning.md
```diff
@@ -0,0 +1,72 @@
...
+Sometimes we make bugfixes and minor feature updates to individual software packages. When this occurs, we update the third number in the version string for that package only. This number doesn't affect compatibility with other packages. For example, you can safely use version **1.1.1** of the Chain Core Mac App with version **1.1.3** of the Java SDK.
+
+### Scope
+
+The package versioning scheme covers a suite of individual software packages that make up Chain Core, including:
+
+- `cored`, the Chain Core server daemon
+- `corectl`, a CLI utility for operating Chain Core
+- `signerd`, the HSM signing server daemon
+- HSM firmware
```
kr: Perhaps we should treat the HSM firmware as a whole separate "product" and give it its own version sequence. Updates to it should be rare.

We probably don't want to be making people upgrade HSM firmware just because they went from cored 1.4 to cored 1.5. Both because that's probably unnecessary most of the time, and because touching the HSM is relatively difficult and expensive compared to deploying things on normal servers.

I'm not super confident of this, it's just my first instinct.

docs/core/reference/versioning.md
```diff
@@ -0,0 +1,72 @@
...
+
+The build version of a specific package may change independently of other packages.
+
+### Semantics
+
+This scheme has semantics that are unique to Chain Core, despite superficial similarities to other versioning schemes such as [Semantic Versioning](http://semver.org/).
+
+#### Compatibility between packages
+
+Software packages in the Chain Core suite are compatible if they share the same major and minor version. For example, if you are running version 1.1 of the Chain Core server, you should use version 1.1 of the SDK.
```
kr: This mostly makes sense, but a strict reading of this sentence means they would have to take downtime during an upgrade. It would be best if we can accommodate incremental rollouts, which implies having some minimal compatibility overlap between different values for the first two numbers in the version string.

docs/core/reference/versioning.md
```diff
@@ -0,0 +1,72 @@
...
+## Summary
+
+Chain Core versioning falls into two categories: **package** versioning and **network** versioning.
+
+Package versioning applies to the pieces of software that work together to participate in a blockchain network. For example, the Chain Core Mac app and the Java SDK.
+
+Network versioning applies to the network API - the interface through which different Chain Cores on a network comunicate to exchange transactions and blocks. Any breaking changes to the network API interface or Chain protocol are reflected by a change to the network version.
+
+## Package Versioning
+The various packages of Chain Core each use a three number versioning scheme - **x.y.z**. The first two numbers indicate compatablity between packages. For example, if you are a running a version of the Chain Core Mac app whose first two numbers are **1.1**, then you should also use a version of an SDK whose first two numbers are also **1.1**.
```
kr: One thing that @oleganza made sure to do in his proposal, that it seems like we've lost here, is to give a name (he called it "product version") to those first two numbers. Right now we have to awkwardly refer to them as "the first two numbers".

docs/core/reference/versioning.md
devongundry: For now we could just pull HSM firmware out of this doc. you're OK with keeping signerd in there?

docs/core/reference/versioning.md
devongundry: I think for now we're going to be strict and require downtime. And then i know we can do better in the future.

PR #175, opened 2016-11-10
dashboard: improve build performance and developer experience

dominic:
Reduces Dashboard development startup time from 15s to 3s. 
Reduces Dashboard development rebuild time from 4s to 500ms-1s.

This PR packages unchanging dependencies in a webpack DLL, allowing
development server to avoid reparsing all dependencies individually
with each rebuild.


dominic:
@jeffomatic PTAL

dashboard/src/config/webpack.app.js
```diff
@@ -57,19 +64,22 @@ var config = getConfig({
...
-
-// Enable CSS modules
+// Customize loader configuration
 let loaders = config.module.loaders
 
 for (let item of loaders) {
-  if (item.loader) {
-    item.loader = item.loader.replace('css-loader','css-loader?modules&importLoaders=1&localIdentName=[name]__[local]__[hash:base64:5]')
+  // Enable CSS module support
+  if (item.loader && item.loader.indexOf('css-loader') > 0) {
```
jeffomatic: If we made this an `if/else if/else` ladder it would clarify that these are mutually exclusive cases.

dashboard/bin/dependencies.js
```diff
@@ -0,0 +1,45 @@
+/*eslint-env node*/
```
jeffomatic: Think this is also from an external source too?

dashboard/src/config/webpack.app.js
dominic: They are not mutually exclusive: CSS loader and SASS loader modifications can be required on the same loader. Looking forward to a clearer non-HJS config.

PR #177, opened 2016-11-10
.github: remove pull request template

jeffomatic:
At least for the time being, the PR template seems superfluous
given that PRs almost entirely come from within the org. Having a
blank template saves a bit of time and effort, and eliminates the
possibility of mistakenly adding the template text to a commit
message.


jeffomatic:
PTAL

jeffomatic:
Curious to hear if there are any votes against?

PR #178, opened 2016-11-10
cmd/cored: use mempool

jbowens:
Use a mempool instead of the txdb pool implementation. Forward submit
calls to the leader for now until we can gossip txs within a Core.


jbowens:
PTAL

core/transact.go
```diff
@@ -286,9 +286,14 @@ type submitArg struct {
 	WaitUntil    string `json:"wait_until"` // values none, confirmed, processed. default: processed
 }
 
-// POST /v3/transact/submit
-// Idempotent
-func (h *Handler) submit(ctx context.Context, x submitArg) interface{} {
+// POST /submit-transaction
+func (h *Handler) submit(ctx context.Context, x submitArg) (interface{}, error) {
+	if !leader.IsLeading() {
+		var resp map[string]interface{}
```
erykwalder: `json.RawMessage` maybe?

protocol/mempool/mempool.go
```diff
@@ -23,14 +25,18 @@ func New() *MemPool {
...
 }
 
 // Dump returns all pending transactions in the pool and
 // empties the pool.
 func (m *MemPool) Dump(context.Context) ([]*bc.Tx, error) {
-	txs := m.pool[:len(m.pool):len(m.pool)]
-	m.pool = nil
+	m.mu.Lock()
+	txs := m.pool
+	m.pool = make([]*bc.Tx, 0, len(txs))
```
erykwalder: Why do we preserve the length?

jbowens: It was intended as an estimate of what the next pool size might be. Might not be worth it if it's confusing.

jbowens:
updated, PTAL

erykwalder:
LGTM

PR #179, opened 2016-11-10
dashboard: allow Safari to submit balances form with sum_by parameter

dominic:
Safari will not allow a form to be submitted with the enter key unless
there is a submit button on the page that is not hidden. This is in contrast
to Chrome, which worked fine with a hidden submit button.


dominic:
PTAL

PR #180, opened 2016-11-10
sdk/ruby: version 1.0.0

jeffomatic:
This commit finalizes changes for Ruby SDK 1.0.0:

- set VERSION constant
- add license document
- update readme with Rubygems instructions

### DIFF
--- a/sdk/ruby/README.md
+++ b/sdk/ruby/README.md
@@ -4,7 +4,7 @@
 
 ### Get the gem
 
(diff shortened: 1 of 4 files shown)


sdk/ruby/README.md
```diff
@@ -2,26 +2,21 @@
...
-TBA
-
-#### Via downloaded .gem
-
-Install the gem into your gem library:
+For most applications, you can simply add the following to your `Gemfile`:
 
 ```
-gem install --local chain-sdk-<VERSION>.gem
+gem 'chain-sdk', '~> 1.0.0', require: 'chain'
```
dominic: I think pessimistic `1.0.0` works with our versioning. They'll just get Ruby patch updates; let's keep this until we see a reason not to?

PR #181, opened 2016-11-10
vendor/github.com/agl/ed25519: vendor a new version of ed25519

bobg:


bobg:
Note: in this version, `edwards25519` is not an "internal" package, so we can move `chainkd` out from under `ed25519` and make it its own package under `crypto`.

bobg:
PTAL

PR #183, opened 2016-11-10
dashboard: provide help links when incorrect query is entered

Zaryafaraj:


Zaryafaraj:
@dominic PTAL

dashboard/src/features/shared/components/SearchBar/SearchBar.jsx
```diff
@@ -167,6 +167,9 @@ class SearchBar extends React.Component {
             </span>
           </span>}
         </span>
+          <a href="/docs/core/build-applications/queries" target="_blank" className={styles.viewdocs}>
+            Queries
```
dominic: I don't think this label fits very well - might need a little @jeffomatic wordsmithing.

dashboard/src/features/shared/components/SearchBar/SearchBar.scss
```diff
@@ -76,9 +75,26 @@
...
   }
 }
+.viewdocs {
+  background-image: url('assets/images/info.png');
+  background-repeat: no-repeat;
+  background-size: 15px 15px;
+  background-position: 5px center;
+  display: inline-block;
+  padding: 4px 5px 4px 28px;
+  margin-left: 18px;
```
dominic: If the goal is to match the spacing of the maginifying glass and placeholder text above, these sizes don't add up. The exact sizes are:

```
  background-position: 0px center;
  padding: 4px 25px;
  margin-left: 21px; // 20px + 1px border from the input element
```

This lets us align the `i` icon from the left edge of the magnifying glass explicitly.

dashboard/src/features/shared/components/SearchBar/SearchBar.scss
```diff
@@ -76,9 +75,26 @@
     cursor: pointer;
   }
 }
+.viewdocs {
+  background-image: url('assets/images/info.png');
+  background-repeat: no-repeat;
+  background-size: 15px 15px;
```
dominic: If we're using a 15x15 image, let's use a 30x30 variant for the 2x size instead of 64x64.

Zaryafaraj:
@dominic PTAL

dashboard/src/features/shared/components/BaseList/EmptyList.scss
```diff
@@ -26,8 +26,24 @@
   margin: 0 auto;
   line-height: 26px;
 }
-
+.noResultsLabel {
+  padding-bottom: 0;
+}
```
dominic: Can you add new lines between CSS rules? (Multiples to fix in this file).

dashboard/src/features/shared/components/SearchBar/SearchBar.scss
```diff
@@ -80,5 +79,6 @@
 .queryTime {
   display: inline-block;
   padding: 4px 20px;
-  opacity: 0.66
+  opacity: 0.66;
+  float: right;
```
dominic: Is floating right a vestige of the previous design when there was a button on the left? If so, can we revert this? If not, let's use `text-align` instead of `float`. We don't use `float` anywhere else in our CSS.

Zaryafaraj:
@dominic PTAL

PR #184, opened 2016-11-11
cored: add release version to compile

jeffomatic:
This commit proposes a method for including version information into the
cored binary.

A new global constant, "version", is present in the cored executable.
It is returned by the /info endpoint.

When we have decided that the main branch contains the features and
bugfixes sufficient for a new release, we add a new commit that updates
the version constant. Initially, the version will be set to "?", but we
should schedule an uptick to 1.0.1 as soon as possible.

For posterity, we will also annotate this commit with the git tag
"cmd/cored-<VERSION>". (If we want a nice, public UI to navigate the
release history, we might consider using GitHub's releases feature to
perform the tagging).

When preparing new installers, we can refer back to these tags to get a
precise state of the repository.

Worth noting: the version constant is useful for tagging public releases
and communicating with outside users. However, it is not very useful for
our internal development process, since the value of the version
constant will change infrequently. In these cases, other means of
communicating the source version (such as the buildcommit expvar, or the
local HEAD ref) should be used.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -49,6 +49,7 @@ import (
 const (
 	httpReadTimeout  = 2 * time.Minute
 	httpWriteTimeout = time.Hour
+	version          = "?" // most recent release version, adjusted manually
 )
 
 var (
@@ -83,6 +84,7 @@ func init() {
 	expvar.NewString("buildtag").Set(buildTag)
 	expvar.NewString("builddate").Set(buildDate)
 	expvar.NewString("buildcommit").Set(buildCommit)
+	expvar.NewString("version").Set(version)
 	expvar.NewString("runtime.GOOS").Set(runtime.GOOS)
 	expvar.NewString("runtime.GOARCH").Set(runtime.GOARCH)
 	expvar.NewString("runtime.Version").Set(runtime.Version())
--- a/core/core.go
+++ b/core/core.go
@@ -99,6 +99,7 @@ func (h *Handler) leaderInfo(ctx context.Context) (map[string]interface{}, error
 
 	buildCommit := json.RawMessage(expvar.Get("buildcommit").String())
 	buildDate := json.RawMessage(expvar.Get("builddate").String())
+	version := json.RawMessage(expvar.Get("version").String())
 
 	m := map[string]interface{}{
 		"is_configured":                     true,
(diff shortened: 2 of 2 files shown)


jeffomatic:
PTAL @kr would like your opinion!

kr:
The existing buildTag is meant to serve a similar purpose, maybe we should get rid of that while we're at it.

The only thing that's not great about this is that, as you point out, during our own devleopment process the version string will be wrong almost always. That's why we were inserting the build tag during linking, so that a simple `go install` would not pick up a stale value.

The only other thought I had is, if we are going to keep it in Go source code, consider splitting it out into its own file (e.g. `cmd/cored/version.go`) so it's easier to generate as a step in some automated release process.

jeffomatic:
> The only thing that's not great about this is that, as you point out, during our own devleopment process the version string will be wrong almost always. That's why we were inserting the build tag during linking, so that a simple go install would not pick up a stale value.

Right, I see that. So the original workflow would be that we rely fully on tagging to track our version history (as opposed to packing the version data into the commits themselves). The build scripts for our installers would do something like:

```
git checkout $VERSION
commit=`git rev-parse HEAD`
date=`date +%s`
ldflags="-X main.buildTag=$VERSION -X main.buildCommit=$commit -X main.buildDate=$date"
go build -ldflags ldflags ...
```

I'm down with changing fewer things if what we've got provides this mechanism already. How would you feel about simply renaming "buildTag" to "version" or "buildVersion", so there is more alignment to how the value will be presented to users?

jeffomatic:
Also, I'm not sure if I'm inventing problems here, but is there value of having the version itself in source control, as it would be for our SDKs? Tags have the disadvantage of being easy to delete or move without a paper trail.

kr:
> How would you feel about simply renaming "buildTag" to "version" or "buildVersion", so there is more alignment to how the value will be presented to users?

That's fine by me.

> is there value of having the version itself in source control, as it would be for our SDKs?

Maybe we can sort of do that by putting the version string in a file somewhere (e.g. `cmd/cored/version`) and having the build script use the contents of that file. Then a plain 'go install' still won't pick it up, and it doesn't rely on git tags, but instead actual committed file data.

kr:
> (e.g. `cmd/cored/version`)

Or even `cmd/cored/last-release-version` or some variation or something in between to be very explicit.

jeffomatic:
Superseded by #200

PR #185, opened 2016-11-11
core/account/utxodb: remove unused confirmedIn

erykwalder:


jbowens:
lgtm, but I think we might want to start using this to prefer confirmed UTXOs over unconfirmed soon

PR #186, opened 2016-11-11
dashboard: display key configuration options vertically

dominic:


dominic:
screen shot 2016-11-10 at 6 38 15 pm

dominic:
PTAL

dashboard/src/features/shared/components/XpubField/XpubField.jsx
```diff
@@ -49,40 +49,45 @@ class XpubField extends React.Component {
       this.setState({ [typeProps.value]: value })
     }
 
+    const fields = {
+      'mockhsm': <SelectField options={mockhsmKeys}
```
jeffomatic: Does it work to set autoFocus on these when they appear?

dominic: Good idea. It only autofocuses after the first change.

PR #188, opened 2016-11-14
cmd/cored: set MaxIdleConns to MAXDBCONNS too

jbowens:
If max conns and max idle conns are not equal, cored will create many db
connections when build & submit requests are coming in, but close them
soon after while waiting for the block to land. Our DB conn needs are very
periodic.


jbowens:
PTAL

erykwalder:
LGTM

PR #190, opened 2016-11-14
sdk/java: make sdk Java 1.7 compliant

### LINKED ISSUE #187: Make Java SDK 1.7 compliant
#### Environment 
1. Chain Core Version: latest
2. Chain Core SDK Language and Version: Java latest 
3. Host: Docker on linux

#### Issue
To achieve a more broad usage of the Java SDK it would be nice to make it Java 7 compliant.
I've downloaded the source code and there are a few steps to convert the SDK to Java 7.

1. com.chain.http.Client -> remove lambdas (just create 3 inner classes o anonymous classes)
2. com.chain.api.PagedItems -> implement the remove method (the Java 8 default is throwing a UnsupportedOperationException("remove"))
3. com.chain.signing.HsmSigner.addKey(String, Client) -> add type String to new ArrayList<>

boymanjor:
Fixes #187

### DIFF
--- a/sdk/java/src/main/java/com/chain/api/PagedItems.java
+++ b/sdk/java/src/main/java/com/chain/api/PagedItems.java
@@ -25,6 +25,11 @@ public abstract class PagedItems<T> implements Iterator<T> {
   private int pos;
 
   /**
+   * Keeps track of whether it is safe to call {@link PagedItems#remove()}
+   */
+  private boolean canRemove;
+
+  /**
    * Page of api objects returned from the most recent query.
    */
   @Expose(serialize = false)
@@ -54,6 +59,7 @@ public abstract class PagedItems<T> implements Iterator<T> {
     this.pos = 0;
     this.list = new ArrayList<>();
     this.lastPage = false;
+    this.canRemove = false;
   }
 
   /**
@@ -77,6 +83,7 @@ public abstract class PagedItems<T> implements Iterator<T> {
    * @return api object of type T
    */
   public T next() {
+    canRemove = true;
     return list.get(pos++);
   }
 
@@ -104,5 +111,16 @@ public abstract class PagedItems<T> implements Iterator<T> {
         return false;
       }
     }
+  }
+
+  /**
+   * Removes the last element returned by {@link PagedItems#next()} from the underlying collection.
+   */
+  public void remove() throws IllegalStateException {
+    if (!canRemove) {
+      throw new IllegalStateException();
+    }
+    this.list.remove(pos - 1);
+    this.canRemove = false;
   }
 }
--- a/sdk/java/src/main/java/com/chain/api/PagedItems.java
+++ b/sdk/java/src/main/java/com/chain/api/PagedItems.java
@@ -25,11 +25,6 @@ public abstract class PagedItems<T> implements Iterator<T> {
   private int pos;
 
   /**
-   * Keeps track of whether it is safe to call {@link PagedItems#remove()}
-   */
-  private boolean canRemove;
-
-  /**
    * Page of api objects returned from the most recent query.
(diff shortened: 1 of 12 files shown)


boymanjor:
@jeffomatic PTAL

jeffomatic:
👀

sdk/java/src/main/java/com/chain/api/PagedItems.java
```diff
@@ -105,4 +112,15 @@ public boolean hasNext() {
       }
     }
   }
+
+  /**
+   * Removes the last element returned by {@link PagedItems#next()} from the underlying collection.
+   */
+  public void remove() throws IllegalStateException {
+    if (!canRemove) {
```
jeffomatic: TL;DR we should consider simplifying this function to `throw UnsupportedOperationException`.

According to the [Java 7 docs](https://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html), it is perfectly fine to implement this method by throwing `UnsupportedOperationException`. For our case, I believe this is the more appropriate behavior. `remove` makes sense if the iterator object refers back to a collection that the user interacts with directly. In other words, it seems to be appropriate for this type of situation:

```
Collection c;
Iterator i = c.iterator();
while (i.hasNext()) {
  Object o = i.next();
  if (shouldRemove(0)) {
    i.remove();
  }
}

// Now collection c is updated.
```

However, I believe `PagedItems` works a bit differently. While there is an underlying `List<T>`, we don't really encourage using that list directly. Calling `remove` on a `PagedItems` object is a weird use case.

Moreover, it may be better for us to make `list` a protected member, so that users are not tempted to use it directly.

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -141,13 +143,15 @@ public Client(URL url, String accessToken) {
...
-        action,
-        body,
-        (Response response, Gson deserializer) ->
-            new BatchResponse(response, deserializer, tClass, eClass));
+  public <T> BatchResponse<T> batchRequest(
+      String action, Object body, final Type tClass, final Type eClass) throws ChainException {
+    ResponseCreator<T> rc =
+        new ResponseCreator<T>() {
+          public T create(Response response, Gson deserializer) throws ChainException, IOException {
+            return (T) new BatchResponse<T>(response, deserializer, tClass, eClass);
```
jeffomatic: Does this actually work at runtime? There appear to be two competing meanings of the generic specifier `T` in this method.

1. The type of the object returned by the ResponseCreator lambda class. In this case, it should be `ResponseCreator<BatchResponse<T>>`
2. The type of the object contained in the `BatchResponse` generic class.

My assumption would be that the code in the PR will break at runtime, but this might work:

```
ResponseCreator<BatchResponse<T>> rc =
        new ResponseCreator<BatchResponse<T>>() {
          public BatchResponse<T> create(Response response, Gson deserializer) throws ChainException, IOException {
            return new BatchResponse<T>(response, deserializer, tClass, eClass);
          }
        }
```

sdk/java/src/main/java/com/chain/api/PagedItems.java
boymanjor: I'll definitely buy this. I'll implement `remove` as stated above.

sdk/java/src/main/java/com/chain/http/Client.java
boymanjor: This does work at runtime, but your suggested changes add clarity to the code. I'll update.

boymanjor:
updated, PTAL

fcanevar:
Hi! Thanks for updating the SDK. Do you know when will it be available through the website?

PR #191, opened 2016-11-14
perf: add TradeMirror.java, a new perf benchmark

jbowens:
Add a new perf test that mirrors trades on the blockchain, issuing if
necessary.


jbowens:
jk

PR #192, opened 2016-11-14
core/account/utxodb: refactor findMatchingUTXOs call

jbowens:
Refactor the findMatchingUTXOs call into sourceReserver.reserve so
that the sourceReserver can cache UTXOs that it knows are available.


core/account/utxodb/reserve.go
```diff
@@ -282,12 +289,6 @@ func (sr *sourceReserver) reserve(rid uint64, amount uint64, utxos []*UTXO) ([]*
 			break
 		}
 	}
-
-	if reserved+unavailable < amount {
```
erykwalder: I think there's a "race", so maybe this should be left in

jbowens:
decided to take the baby step of refactoring the `findMatchingUTXOs` call into `reserve()` for now. This will help with caching UTXOs on the sourceReserver.

erykwalder:
LGTM with title change

PR #197, opened 2016-11-14
sync/idempotency: make errors non-idempotent

erykwalder:
The only use case for this right now is in utxo reservation. Errors such
as ErrReserved can be retried, and will continue to only result in
ErrReserved, even if there is an underlying change.


kr:
Imperative voice for subject? (e.g.)

```
sync/idempotency: make errors non-idempotent
```

jbowens:
lgtm

PR #198, opened 2016-11-14
Add Go AST Scanner (gas)

tarcieri:
gas is a static analyzer for security issues:

https://github.com/GoASTScanner/gas

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -41,6 +41,15 @@ build:
           bin/cfmt
           git diff --exit-code
     - script:
+        name: gas
+        code: |
+          go get github.com/GoASTScanner/gas
+          cd $GOPATH/src/chain
+          # TODO: Remove exclusions on gas checks
+          # G101: Hardcoded credentials (Presently only false positives)
+          # G204: Unsafe command execution (Presently only affects build tools)
+          gas -quiet -exclude=G101,G204 ./{cmd,core}/**/*.go
+    - script:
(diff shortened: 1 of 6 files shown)


wercker.yml
```diff
@@ -41,6 +41,15 @@ build:
           bin/cfmt
           git diff --exit-code
     - script:
+        name: gas
+        code: |
+          cd $GOPATH/src
+          git clone https://github.com/GoASTScanner/gas.git
+          cd gas
+          go install
```
kr: These four lines could be written as

```
go get github.com/GoASTScanner/gas
```

wercker.yml
```diff
@@ -41,6 +41,15 @@ build:
...
+          cd $GOPATH/src/chain
+          gas
```
kr: When I run just `gas` locally, I get a usage message. Should this be like `gas chain/...` or something?

tarcieri:
@kr this is a WIP (perhaps I should've labeled it as such). At least now I'm getting the same output on Werker that I'm getting locally (presently 42 issues, a mixture of false positives and some legitimate ones)

kr:
Aha, sorry, I jumped the gun there. 😛

tarcieri:
We appear to have hit a regression in gas introducing a false positive for G404: Insecure random number generation. I reported an issue here and disabled the check for now:

https://github.com/GoASTScanner/gas/issues/87

PR #200, opened 2016-11-14
cmd/cored: add release version file

jeffomatic:
This gives us space to track the latest version of the cored binary in
version control.

When we have decided that the main branch contains the features and
bugfixes sufficient for a new release, we add a new commit that updates
the version in this file. Initially, the version will be set to "?", but
we should schedule an uptick to 1.0.1 as soon as possible.

For posterity, we will also annotate this commit with the git tag
"cmd/cored-<VERSION>". (If we want a nice, public UI to navigate the
release history, we might consider using GitHub's releases feature to
perform the tagging).

When compiling cored for a new release, use the contents of this file as
the buildTag build variable:

    buildTag=`cat cmd/cored/last-release-version | tr -d '[:space:]'`
    ldflags="-X main.buildTag=$buildTag ..."
    go build -ldflags "$ldflags" ...

The release version will only be included in builds that are built in
this manner.


jeffomatic:
PTAL @kr, this a simpler alternative to our [previous conversation](https://github.com/chain/chain/pull/184).

kr:
:eyes:

kr:
Lovely. Haha. Presumably we will need some future update to builds/core/build or some equivalent script?

jeffomatic:
Resolution: won't land.

Since build tools will need to check out a specific Git tag for building the corresponding cored version, we may as well use the tagname itself to provide a value for the `buildTag` build variable, e.g., if `$tag` contains the Git tag, `${tag#cmd/cored-}` strips the prefix and yields the release version encoded in the tag.

This normalizes the release version by making the tag the source of truth. Tags are malleable, but there should be backups across the development team to make that less of a problem. We can complicate the versioning process in the future.

PR #201, opened 2016-11-14
core/account/utxodb: cache confirmed UTXOs

jbowens:
Cache confirmed UTXOs on the sourceReserver. Since they're confirmed,
we're guaranteed that they will stay valid. Also, since the reserver is only
run on one leader process, no one else can have spent the UTXOs unless
they circumvented reservation.


erykwalder:
LGTM

PR #205, opened 2016-11-15
core/account/utxodb: revert cache confirmed UTXO

jbowens:
This reverts commit 1248d643afdb425e35dd934dd933f44fd31726eb. We can double reserve
in some situations.


erykwalder:
LGTM

PR #206, opened 2016-11-15
docker/ci: new image for continuous integration

boymanjor:
Testing commands are now split into separate scripts
and included in the ci image. The wercker.yml has
been updated to run two pipelines. The ci image has
dependencies to handle dashboard, core and java sdk
testing and can be expanded for future use cases.
It also allows us to try out different ci solutions,
regardless of their configuration rules.

Future work involves shutting down testbot and updating
required statuses for pull requests.


boymanjor:
PTAL, note: wercker/build should be failing. It's no longer referenced in wercker.yml

PR #207, opened 2016-11-15
protocol/mempool: topologically sort

jbowens:
Sort transactions topologically when dumping the contents of the
pool, if they're not already in topological order. The old txdb pool
used this, but I forgot to port when we moved to the mempool.

Also, don't add a tx to the pool if it's already there.

If we're already topo sorting, should we just use an unordered
`map[bc.Hash]*bc.Tx`?

### DIFF
--- a/protocol/mempool/mempool.go
+++ b/protocol/mempool/mempool.go
@@ -9,34 +9,49 @@ import (
 	"context"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


erykwalder:
LGTM, I'd also be fine with just making it a map.

PR #208, opened 2016-11-15
dashboard: change width of Replication Lag info in Network Status

chrisgarvin:


chrisgarvin:
Resolves issue where Replication Lag info's width is determined by the width of the generator/local block info

Original: 
image

Modified:
image

chrisgarvin:
PTAL

PR #210, opened 2016-11-15
core/account/utxodb: cache confirmed utxos

erykwalder:
Cache confirmed UTXOs on the sourceReserver. Since they're confirmed,
we're guaranteed that they will stay valid. Also, since the reserver is
only run on one leader process, no one else can have spent the UTXOs
unless they circumvented reservation.

This is the same as commit 1248d643afdb425e35dd934dd933f44fd31726eb,
with an additional check that already reserved utxos are not cached,
since that could cause a double spend.

### DIFF
--- a/core/account/utxodb/reserve.go
+++ b/core/account/utxodb/reserve.go
@@ -38,6 +38,7 @@ type UTXO struct {
 	bc.AssetAmount
 	Script []byte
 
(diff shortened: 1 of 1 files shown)


jbowens:
lgtm

erykwalder:
jk lol

jbowens:
lolz

PR #213, opened 2016-11-15
core/account: move utxodb reserver here

erykwalder:
The utxodb package was only called by the account package, so it can be
moved in unexported.

### DIFF
--- a/core/account/reserve.go
+++ b/core/account/reserve.go
@@ -1,4 +1,4 @@
-package utxodb
+package account
 
 import (
 	"context"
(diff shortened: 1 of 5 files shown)


core/account/reserve.go
```diff
@@ -136,35 +136,35 @@ func (re *Reserver) reserve(ctx context.Context, source Source, amount uint64, c
 	return res, nil
 }
 
-// ReserveUTXO reserves a specific UTXO for spending. The resulting
+// ReserveUTXO reserves a specific utxo for spending. The resulting
```
jbowens: we got a couple of comments too

jbowens:
lgtm

PR #216, opened 2016-11-16
core/query: Use bytes.Buffer to construct outputs query

tarcieri:
This change both resolves a TODO and makes the code amenable to static analysis
for SQLi vulnerabilities by gas.

Though safe in its previous form, it was written in a way in which gas was
unable to assert its safety.

### DIFF
--- a/core/query/outputs.go
+++ b/core/query/outputs.go
@@ -1,6 +1,7 @@
 package query
 
 import (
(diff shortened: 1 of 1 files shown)


kr:
:eyes:

kr:
LGTM

PR #217, opened 2016-11-16
docs/reference: introduce versioning schemes

devongundry:
This document introduces the Chain versioning schemes: core versioning, network versioning, and signer versioning.


devongundry:
@ryandotsmith PTAL

PR #218, opened 2016-11-16
dashboard: input autocomplete suggestions trigger on first keydown

chrisgarvin:


chrisgarvin:
Fixes issue where autocomplete suggestions would not trigger on first keydown upon entering page.

Original:
image

Modified:
image

chrisgarvin:
PTAL

dominic:
One comment - can you remove the `cb` parameter from the `fetchAll` definition at https://github.com/chain/chain/blob/d6b12a02797e86e8286d0dd51f76a5ea6fedf40d/dashboard/src/components/Common/Autocomplete/AutocompleteField.jsx#L92 ? It's not used anymore, and it turns out `fetchAll` doesn't call it in its current implementation.

chrisgarvin:
@dominic removed cb from fetchAll 👍

dominic:
LGTM

PR #219, opened 2016-11-16
docs: remove docker for mac option

Zaryafaraj:
Due to performance issues, docker for mac is being removed as an
offering.


Zaryafaraj:
@ryandotsmith PTAL

PR #220, opened 2016-11-16
core/account: rename Script to ControlProgram

jbowens:
It looks like we missed this one when renaming scripts to control programs.


kr:
LGTM

PR #221, opened 2016-11-17
cmd/benchcore: capture trace profiles

erykwalder:
This makes the "-p" flag also capture trace profiles, which can be
examined with the pprof trace tool.

### DIFF
--- a/cmd/benchcore/main.go
+++ b/cmd/benchcore/main.go
@@ -33,7 +33,7 @@ import (
 
 var (
 	flagD        = flag.Bool("d", false, "delete instances from previous runs")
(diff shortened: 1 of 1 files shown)


jbowens:
👀

cmd/benchcore/main.go
```diff
@@ -587,52 +587,30 @@ func profile(coreURL, clientToken string) {
 
 	ticker := time.Tick(profileFrequency)
```
jbowens: what do you think about making it more frequent too?

jbowens:
lgtm

PR #222, opened 2016-11-17
core/txfeed: Use bytes.Buffer to construct txfeed queries

tonychain:
Using bytes.Buffer instead of string concatenation makes the code amenable to
static analysis by gas.


kr:
:eyes:

kr:
LGTM

PR #223, opened 2016-11-17
dashboard: provide help links when incorrect query is entered

Zaryafaraj:


Zaryafaraj:
@dominic PTAL again :)

PR #224, opened 2016-11-17
wercker.yml: Add Go AST Scanner (gas)

tonychain:
gas is a static analyzer for security issues:

https://github.com/GoASTScanner/gas

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -41,6 +41,18 @@ build:
           bin/cfmt
           git diff --exit-code
     - script:
+        name: gas
+        code: |
+          git clone https://github.com/GoASTScanner/gas $GOPATH/src/github.com/GoASTScanner/gas
+          # gas is somewhat unstable, so pin to a known-good commit hash
+          cd $GOPATH/src/github.com/GoASTScanner/gas && git reset --hard d30c5cde3613e9ba0129febda849e4d4df1d57cd
+          go install github.com/GoASTScanner/gas
+          cd $GOPATH/src/chain
+          # TODO: Remove exclusions on gas checks
+          # G101: Hardcoded credentials (Presently only false positives)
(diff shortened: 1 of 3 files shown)


tonychain:
Note: this branch includes #222, which should get merged first

kr:
:eyes:

wercker.yml
```diff
@@ -41,6 +41,15 @@ build:
           bin/cfmt
           git diff --exit-code
     - script:
+        name: gas
+        code: |
+          go get github.com/GoASTScanner/gas
```
kr: Do we want to pin this to a particular revision of gas? Or maybe put it in the docker image? See the goimports stuff in `docker/wercker/Dockerfile` (and instructions in the Readme there). I can think of tradeoffs on both sides.

kr:
LGTM with one question

wercker.yml
tonychain: I think pinning to a particular commit hash of gas makes sense at least, especially as there seem to be pretty frequent (breaking) changes upstream

wercker.yml
```diff
@@ -41,6 +41,18 @@ build:
           bin/cfmt
           git diff --exit-code
     - script:
+        name: gas
+        code: |
+          git clone https://github.com/GoASTScanner/gas $GOPATH/src/github.com/GoASTScanner/gas
+          # gas is somewhat unstable, so pin to a known-good commit hash
+          cd $GOPATH/src/github.com/GoASTScanner/gas && git reset --hard d30c5cde3613e9ba0129febda849e4d4df1d57cd
+          go get github.com/GoASTScanner/gas
```
kr: nit: s/get/install/ to emphasize that we expect this not to do network stuff

PR #225, opened 2016-11-17
core/account: don't index unconfirmed outputs

jbowens:
Only index confirmed account UTXOs. Indexing and spending unconfirmed
UTXOs allows for cascading failures if a single transaction fails.
Previously, we indexed unconfirmed outputs only if we had reasonable
certainty that the transaction would be accepted by the network (all of
the associated keys are local to the Chain Core). Transactions can still
fail due to latency issues and the transaction's MaxTime elapsing.

If other transactions had spent the failed tx's pool outputs, the entire
tree of transactions would fail. These cascading failures are difficult
to recover from. If we can avoid indexing/spending pool outputs, we'll
have a simpler system.

Long-term the removal of pool output spending can simplify tx rejection
detection because a Core can validate the transaction against its local
state tree.

### DIFF
--- a/core/account/indexer.go
+++ b/core/account/indexer.go
@@ -184,10 +184,7 @@ func (m *Manager) upsertConfirmedAccountOutputs(ctx context.Context, outs []*out
 		SELECT unnest($1::text[]), unnest($2::bigint[]), unnest($3::text[]),  unnest($4::bigint[]),
 			   unnest($5::text[]), unnest($6::bigint[]), unnest($7::bytea[]), unnest($8::bytea[]),
 			   $9, unnest($10::bigint[]), $11
-		ON CONFLICT (tx_hash, index) DO UPDATE SET
-			confirmed_in    = excluded.confirmed_in,
-			block_pos       = excluded.block_pos,
-			block_timestamp = excluded.block_timestamp;
+		ON CONFLICT (tx_hash, index) DO NOTHING
 	`
 	_, err := m.db.Exec(ctx, q,
 		txHash,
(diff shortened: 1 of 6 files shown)


jbowens:
Not sure whether we want to go this route. I'm running some benchmarks against a branch with this change and UTXO caching enabled by this.

core/account/indexer.go
```diff
@@ -263,15 +180,14 @@ func (m *Manager) upsertConfirmedAccountOutputs(ctx context.Context, outs []*out
 
 	const q = `
 		INSERT INTO account_utxos (tx_hash, index, asset_id, amount, account_id, control_program_index,
-			control_program, metadata, confirmed_in, block_pos, block_timestamp, expiry_height)
+			control_program, metadata, confirmed_in, block_pos, block_timestamp)
 		SELECT unnest($1::text[]), unnest($2::bigint[]), unnest($3::text[]),  unnest($4::bigint[]),
 			   unnest($5::text[]), unnest($6::bigint[]), unnest($7::bytea[]), unnest($8::bytea[]),
-			   $9, unnest($10::bigint[]), $11, NULL
+			   $9, unnest($10::bigint[]), $11
 		ON CONFLICT (tx_hash, index) DO UPDATE SET
```
erykwalder: `DO NOTHING`?

jbowens: good call

erykwalder:
LGTM

kr:
LGTM in principle (but I haven't (yet) looked at the code here, sorry)

PR #227, opened 2016-11-17
core/account: drop control program id column

jbowens:
The control program ID column was unused. In stats from benchmarks, inserting a control program was the most expensive query in terms of total execution time. The second most expensive query was the subquery retrieving the next value in the ID sequence.


jbowens:
PTAL

erykwalder:
LGTM

PR #232, opened 2016-11-17
all: extend gas coverage to all directories except vendor

tonychain:
Includes minor fixes for gas compliance, or #nosec for innocuous cases

The vendor directory is omitted as there are numerous false positives for issues in third-party libraries, and adding #nosec to vendored code is probably not maintainable long-term.

### DIFF
--- a/log/rotation/file.go
+++ b/log/rotation/file.go
@@ -28,7 +28,7 @@ type File struct {
 }
 
 // Create creates a log writing to the named file
-// with mode 0666 (before umask),
+// with mode 0644 (before umask),
 // appending to it if it already exists.
 // It will rotate to files name.1, name.2,
 // up to name.n.
@@ -83,11 +83,14 @@ func (f *File) write(p []byte) (int, error) {
 	}
 	if f.f == nil {
 		var err error
-		f.f, err = os.OpenFile(f.base, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0666)
+		f.f, err = os.OpenFile(f.base, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644) // #nosec
 		if err != nil {
 			return 0, err
 		}
-		f.w, _ = f.f.Seek(0, 2)
+		f.w, err = f.f.Seek(0, os.SEEK_END)
+		if err != nil {
+			return 0, err
+		}
 	}
 	n, err := f.f.Write(p)
 	f.w += int64(n)
--- a/net/http/gzip/gzip_test.go
+++ b/net/http/gzip/gzip_test.go
@@ -26,7 +26,7 @@ func (n noOpWriter) Write(d []byte) (int, error) {
 func (n noOpWriter) WriteHeader(int) {}
 
 func BenchmarkGzipSmall(b *testing.B) {
-	r, _ := http.NewRequest("GET", "/foo", nil)
+	r, _ := http.NewRequest("GET", "/foo", nil) // #nosec
 	r.Header.Set("accept-encoding", "gzip")
 	h := Handler{http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Write(small)
--- a/protocol/validation/tx.go
+++ b/protocol/validation/tx.go
@@ -227,7 +227,10 @@ func CheckTxWellFormed(tx *bc.Tx) error {
 			} else {
 				program = input.ControlProgram()
 			}
-			scriptStr, _ := vm.Disassemble(program)
+			scriptStr, e := vm.Disassemble(program)
+			if e != nil {
+				scriptStr = "disassembly failed: " + e.Error()
+			}
 			args := input.Arguments()
 			hexArgs := make([]string, 0, len(args))
 			for _, arg := range args {
--- a/wercker.yml
+++ b/wercker.yml
@@ -51,7 +51,7 @@ build:
           # TODO: Remove exclusions on gas checks
           # G101: Hardcoded credentials (Presently only false positives)
           # G204: Unsafe command execution (Presently only affects build tools)
-          gas -quiet -exclude=G101,G204 ./{cmd,core}/**/*.go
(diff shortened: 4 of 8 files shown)


log/rotation/file.go
```diff
@@ -83,11 +83,14 @@ func (f *File) write(p []byte) (int, error) {
 	}
 	if f.f == nil {
 		var err error
-		f.f, err = os.OpenFile(f.base, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0666)
+		f.f, err = os.OpenFile(f.base, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644) // #nosec
```
tonychain: I don't think it makes sense for the logfile to be world-writable, so I changed the perms to `0644`. At the very least I don't think they should be looser than `0664`.

I had to add `#nosec` anyway as it prefers `0600`, which we might also consider.

net/http/gzip/gzip_test.go
```diff
@@ -26,7 +26,7 @@ func (n noOpWriter) Write(d []byte) (int, error) {
 func (n noOpWriter) WriteHeader(int) {}
 
 func BenchmarkGzipSmall(b *testing.B) {
-	r, _ := http.NewRequest("GET", "/foo", nil)
+	r, _ := http.NewRequest("GET", "/foo", nil) // #nosec
```
tonychain: I'm not sure why `-skip="*_test.go"` failed to prevent this file from getting scanned (it seems to be working successfully for several other tests). Possibly a bug in gas?

tonychain:
Bonus points: `gas` is now passing locally on OS X with the parameters in `wercker.yml`

log/rotation/file.go
```diff
@@ -83,11 +83,14 @@ func (f *File) write(p []byte) (int, error) {
...
+		if err != nil {
+			return 0, err
+		}
+		f.w, err = f.f.Seek(0, os.SEEK_END)
```
tonychain: Switched the `2` literal to the `os.SEEK_END` constant here, FYI, in addition to handling the error as opposed to adding a `#nosec` (I believe an `EBADF`-style error could possibly occur here)

log/rotation/file.go
kr: I had been assuming there would be a umask of 0002 or 0022 or 0077. (Or, if the operator really wants them to be world-writable, I suppose they could use 0 as the umask. But why would they ever want that.) This is safer for sure. :+1:

syscall/zsyscall_windows.go
```diff
@@ -18,7 +18,7 @@ var (
 )
 
 func GetModuleFileName(hmodule uint32, buf *uint16, buflen uint32) (n uint32, err error) {
-	r0, _, e1 := syscall.Syscall(procGetModuleFileNameW.Addr(), 3, uintptr(hmodule), uintptr(unsafe.Pointer(buf)), uintptr(buflen))
+	r0, _, e1 := syscall.Syscall(procGetModuleFileNameW.Addr(), 3, uintptr(hmodule), uintptr(unsafe.Pointer(buf)), uintptr(buflen)) // #nosec
```
kr: This is a generated file. It could be clobbered at any time by a future `go generate` command. See the file syscall_windows.go in this same directory. Is there another way of getting gas to ignore this file that doesn't involve modifying it? Or I guess we could always just redo the edit when necessary (since presumably gas will flag it as an error again).

protocol/validation/tx.go
```diff
@@ -227,7 +227,7 @@ func CheckTxWellFormed(tx *bc.Tx) error {
 			} else {
 				program = input.ControlProgram()
 			}
-			scriptStr, _ := vm.Disassemble(program)
+			scriptStr, _ := vm.Disassemble(program) // #nosec
```
kr: One option for at least making that case less confusing would be to check the error but not alter the rest of the execution, just print it out:

```
scriptStr, err := vm.Disassemble(program)
if err != nil {
	t.Log(err)
}
...
```

syscall/zsyscall_windows.go
tonychain: I can add it to the `-skip` list. Should I just do this file, or all of `syscall/`?

tonychain: Added `*syscall*` to the skip list and removed the changes to `zsyscall_windows.go`

protocol/validation/tx.go
tonychain: I tried setting `scriptStr` to the disassembly error in this case. I think that's probably the most helpful.

tonychain:
LGTY?

syscall/zsyscall_windows.go
kr: Actually ideally it would be just this file. (Or I guess files in syscall that start with 'z' which appears to be the convention for the tool that generates this stuff.)

wercker.yml
```diff
@@ -51,7 +51,7 @@ build:
           # TODO: Remove exclusions on gas checks
           # G101: Hardcoded credentials (Presently only false positives)
           # G204: Unsafe command execution (Presently only affects build tools)
-          gas -quiet -exclude=G101,G204 ./{cmd,core}/**/*.go
+          gas -quiet -exclude=G101,G204 -skip="*vendor*" -skip="*syscall*" -skip="*_test.go" ./...
```
kr: (Not sure if my comment on the other thread is effectively buried, so repeating here.)

Actually ideally it would be just [zsyscall_windows.go]. (Or I guess files in syscall that start with 'z' which appears to be the convention for the tool that generates this stuff.)

syscall/zsyscall_windows.go
tonychain: Let me see if I can find an appropriate `-skip` glob for that

kr:
I left a comment but LGTM either way.

tonychain:
I'm having trouble constructing a `-skip` statement that actually captures either `zsyscall_windows.go` or z-prefixed go files. I'm going to go ahead and land this, and we can revisit it if it turns out to be problematic.

PR #233, opened 2016-11-17
core/account: reuse control programs on rollback

jbowens:
If an action creates a control program but the build gets rolled back
due to an error elsewhere in the transaction, the control program is
safe to reuse. We can cache these control programs on the Manager and
reuse them in a future build call.


erykwalder:
LGTM

jbowens:
Closing in favor of #249.

PR #235, opened 2016-11-17
vendor/github.com/agl/ed25519: vendor latest ed25519

bobg:
Vendor github.com/agl/ed25519 at commit 278e1ec8e8a6e017cd07577924d6766039146ced of 30 Aug 2015.

In this version, `edwards25519` is no longer an "internal" subpackage, so we'll be able to move `chainkd` (and future crypto code) out from under `ed25519` and into its own subpackage of `chain/crypto`.


bobg:
PTAL

kr:
:eyes:

kr:
LGTM

PR #236, opened 2016-11-17
chain: update Readme intro

bobg:
Restoring these changes from a recent round of copy-editing that never reached completion.


bobg:
PTAL

bobg:
rebased

PR #241, opened 2016-11-18
all: add changelogs

dominic:


dominic:
PTAL @jeffomatic @tessr @ryandotsmith

PR #243, opened 2016-11-18
crypto: replace chain/crypto/ed25519 with github.com/agl/ed25519 and adapt

bobg:
Well, gee. When I vendored `github.com/agl/ed25519` in https://github.com/chain/chain/commit/397d824dc8b3b656500d7ba4f240e42eeac73678 I expected it to be a drop-in replacement for the version we had in `chain/crypto`, but it's not. This PR contains the changes needed to adapt to the new `ed25519`'s API.

The result is not an improvement, readability-wise, over what we had before, though it looks like type-safety is slightly enhanced. Our choices: stick with the old `ed25519`, or plough forward with this one, ideally abstracting common patterns into our own shim (which can precede or follow this PR).


kr:
Is this because we switched from https://godoc.org/golang.org/x/crypto/ed25519 (which is newer and closer to "official") to https://godoc.org/github.com/agl/ed25519? I don't feel super strongly about the interface changes; they seem like six of one vs half a dozen of the other.

I feel unqualified to comment on implementation quality. I'd defer to anyone with a strong opinion.

bobg:
> https://godoc.org/golang.org/x/crypto/ed25519 (which is newer and closer to "official")

I can no longer remember where the idea to switch to github.com/agl/ed25519 originated, but if golang.org/x/crypto/ed25519 is newer and more official, that's what we should stick with, for sure. (I just got excited when I saw a version of `edwards25519` that wasn't "internal.")

If there are no objections I'll close this PR and revert https://github.com/chain/chain/commit/397d824dc8b3b656500d7ba4f240e42eeac73678.

PR #244, opened 2016-11-18
core/txbuilder: refactor Build

jbowens:
Refactor txbuilder.Build and txbuilder.Action so that actions modify a shared txbuilder.TemplateBuilder instead of returning txbuilder.BuildResult. Some tests depend on actions orderings matching their corresponding inputs and outputs, so this removes the parallelism from building, at least for now.

This change has one API implication: ErrBadRefData will now correctly be returned as a part of the ErrAction multierror.

### DIFF
--- a/core/account/builder.go
+++ b/core/account/builder.go
@@ -37,7 +37,7 @@ type spendAction struct {
 	ClientToken   *string       `json:"client_token"`
 }
 
-func (a *spendAction) Build(ctx context.Context, maxTime time.Time) (*txbuilder.BuildResult, error) {
+func (a *spendAction) Build(ctx context.Context, maxTime time.Time, b *txbuilder.TemplateBuilder) error {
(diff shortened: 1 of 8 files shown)


jbowens:
Is this too crazy? My goal is to be able to insert all of a tx's control programs in a batch and only if the reservations succeed.

With this interface, the account package could store the information about what control programs to create on the TemplateBuilder using SetData and add OnBuild callbacks to actually create the control programs and corresponding outputs.

bobg:
An intermediate object type during `Build` calls - brilliant. This is exactly the missing ingredient from when Jeff and I were trying to figure out how to get the mintime/maxtime API right. (Banishing maxtime from the `Build` interface parameter list, and adding flexible mintime/maxtime logic to the builder object, should be a follow-up task.) Very nice! LGTM

PR #246, opened 2016-11-18
docker/ci: new image for continuous integration

boymanjor:
Testing commands are now split into separate scripts
and included in the ci image. The wercker.yml has
been updated to run two pipelines (cored & java). The 
ci image has dependencies to handle dashboard, core
and java sdk testing and can be expanded for future use
cases. It also allows us to try out different ci solutions,
regardless of their configuration rules.

Future work involves shutting down testbot and removing
the source from the monorepo. Also, adding separate
pipelines for the dashboard and ruby tests.


boymanjor:
PTAL. The wercker/build status should fail. The new "cored" and "java" pipelines are passing. Reviewers can verify in the wercker UI.

jbowens:
lgtm

PR #248, opened 2016-11-18
dashboard: tx autocomplete tab/hover events autofilling input

chrisgarvin:


chrisgarvin:
Original:
image

Modified:
image

chrisgarvin:
PTAL

devongundry:
this makes me happy

jeffomatic:
👀

dashboard/src/components/Common/Autocomplete/AutocompleteField.jsx
```diff
@@ -36,7 +37,9 @@ class AutocompleteField extends React.Component {
 
   renderSuggestion(suggestion) {
     return (
-      <span>{suggestion.alias}</span>
+      <div onMouseOver={this.handleHover.bind(this, suggestion.alias)}>
```
jeffomatic: Would suggest this pattern:

```
<div onMouseOver={() => this.handleHover(suggestion.alias)}>
```

jeffomatic: Or even, we can remove the intermediate function:

```
<div onMouseOver={() => this.props.fieldProps.onChange(suggestion.alias)}>
```

dashboard/src/components/Common/Autocomplete/AutocompleteField.jsx
```diff
@@ -57,6 +60,20 @@ class AutocompleteField extends React.Component {
     })
   }
 
+  handleHover(suggestion) {
+    this.props.fieldProps.onChange(suggestion)
+  }
+
+  tabCheck(event, suggestions, fieldProps) {
```
jeffomatic: It's a bit subtle, but I think `suggestions` and `fieldProps` are local to this object, so you may just want to look them up from `this` rather than passing them in. `tabCheck` is not meant to be used outside the context of the local component, so we might as well assume those fields are available via `this`.

dashboard/src/components/Common/Autocomplete/AutocompleteField.jsx
```diff
@@ -74,7 +91,8 @@ class AutocompleteField extends React.Component {
           className: `form-control ${this.props.className}`,
           value: fieldProps.value,
           placeholder: this.props.placeholder,
-          onChange: (event, { newValue }) => fieldProps.onChange(newValue) }}
+          onChange: (event, { newValue }) => fieldProps.onChange(newValue),
+          onKeyDown: (event) => this.tabCheck(event, suggestions, fieldProps)}}
```
jeffomatic: The calling pattern here makes the constructor's `this.tabCheck.bind` call redundant.

jeffomatic:
This is NICE. General functionality comments:

- It feels a bit weird that the tab completer doesn't show a visual cue of the option that is selected by default. In other words, we should highlight the top option (i.e.: treat the top option as selected) as soon as there are options available.
- On that note, the styling of the selected element is a bit dark. While we're tweaking this component, maybe we can experiment with some other visual styles? Maybe I'm the only one that feels that way.
- I think the enter key should do the same as the tab key.

dashboard/src/components/Common/Autocomplete/AutocompleteField.jsx
chrisgarvin: I think you're right on the second suggestion, so I updated for that route. No reason to have a separate handleHover in this case.

dashboard/src/components/Common/Autocomplete/AutocompleteField.jsx
```diff
@@ -60,16 +59,12 @@ class AutocompleteField extends React.Component {
...
-  handleHover(suggestion) {
-    this.props.fieldProps.onChange(suggestion)
-  }
-
-  tabCheck(event, suggestions, fieldProps) {
-    if (event.keyCode == 9) {
-      const suggestion = suggestions[0]["alias"]
-      const input = fieldProps.value.toLowerCase()
+  tabCheck(keyCode) {
+    if (keyCode == 9 || keyCode == 13) {
```
jeffomatic: Can we add a comment reminding the reader what key 9 and key 13 are?

dashboard/src/components/Common/Autocomplete/AutocompleteField.scss
```diff
@@ -20,14 +20,27 @@
...
 .suggestion {
   border-bottom: 1px solid $input-border;
-  padding: 12px;
 
   &:last-child {
     border: none;
   }
+
+  &:first-child {
+    @extend .suggestionFocused;
```
jeffomatic: On my computer, this keeps the first child selected at all times:

screen shot 2016-11-19 at 10 29 16 am

Do you think there might be a way to do this so that the first child is selected *logically*, not just in CSS?

jeffomatic:
This feels great Chris! One last tweak that would help a bit (and of course this was probably an issue with the component before you ever got to it): I notice that the order of the options is a bit weird.

screen shot 2016-11-19 at 1 56 14 pm

I think most autocomplete systems use some kind of proximity and/or frequency heuristic for sorting the results, but I think in our case the best thing we can do is alphabetical sorting. What do you think?

dashboard/src/components/Common/Autocomplete/AutocompleteField.jsx
```diff
@@ -24,9 +24,12 @@ class AutocompleteField extends React.Component {
       return []
     }
 
-    const regex = new RegExp(escapedValue, 'i')
+    const regex = new RegExp('^' + escapedValue, 'i')
 
-    return this.props.items.filter(item => regex.test(item.alias))
+    const suggestions = this.props.items.filter(item => regex.test(item.alias))
+    suggestions.sort((a,b) => a.alias.localeCompare(b.alias))
```
jeffomatic: Might wanna try it out, but I think `sort` will work with no argument function in this case.

PR #250, opened 2016-11-19
all: update changelog formats

jeffomatic:
Minor tweaks to the changelogs:

- Use root changelog as index document
- Move cored changelog to cmd/cored
- Add h1-level title headings to each changelog


jeffomatic:
@dominic PTAL

PR #251, opened 2016-11-19
cmd/benchcore: tune Postgres

jbowens:
Tune Postgres according to the instance type. These values are guesswork
and could use some refining, but they dramatically increase
performance.


jbowens:
PTAL

erykwalder:
LGTM

PR #253, opened 2016-11-19
core/pin: process blocks in parallel

erykwalder:
Each block processor is independent and idempotent, and so they can be
run in parallel. This particularly helps out the transaction indexer,
which can fall behind due to the long time it takes to process blocks.

### DIFF
--- a/core/pin/pin.go
+++ b/core/pin/pin.go
@@ -229,7 +229,7 @@ func (p *pin) complete(ctx context.Context, height uint64) error {
 		max = p.completed[i]
 	}
 
-	if max < p.height {
+	if max == p.height {
 		return nil
 	}
 
--- a/core/pin/pin.go
+++ b/core/pin/pin.go
@@ -2,6 +2,7 @@ package pin
 
 import (
 	"context"
+	"sort"
 	"strconv"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


jbowens:
Should we limit the parallelism? If a processor is _very_ slow, we could process an unbounded number of blocks in parallel slowing it down even more.

core/pin/pin.go
```diff
@@ -199,25 +189,71 @@ func (p *pin) getHeight() uint64 {
...
+		max = p.height
+		i   int
+	)
+	for i = 0; i < len(p.completed); i++ {
+		if p.completed[i] <= max {
+			continue
+		} else if p.completed[i] > max+1 {
+			break
+		}
+		max = p.completed[i]
```
jbowens: should this line be above the continue?

erykwalder:
Added a concurrency limit, PTAL

core/pin/pin.go
```diff
@@ -179,16 +176,18 @@ func (s *Store) Listen(ctx context.Context, pinName, dbURL string) {
...
+	completed []uint64
 
 	db   pg.DB
 	name string
+	sem  chan bool
 }
 
 func newPin(db pg.DB, name string, height uint64) *pin {
-	p := &pin{db: db, name: name, height: height}
+	p := &pin{db: db, name: name, height: height, sem: make(chan bool, 10)}
```
jbowens: should this be `make(chan bool, processorWorkers)`?

core/pin/pin.go
jbowens: Ah, I see. I misinterpreted this

jbowens:
lgtm

core/pin/pin.go
erykwalder: ha, yes

PR #257, opened 2016-11-21
core/account: cache UTXOs in reserver

jbowens:
Cache account UTXOs in the reserver. This allows some reservation
requests to avoid hitting the database if there are enough cached
UTXOs to cover the reservation amount.

### DIFF
--- a/core/account/reserve.go
+++ b/core/account/reserve.go
@@ -40,6 +40,8 @@ type utxo struct {
 
 	AccountID           string
 	ControlProgramIndex uint64
+
+	confirmedIn, blockPos uint64
 }
 
 func (u *utxo) source() source {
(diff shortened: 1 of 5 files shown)


core/account/reserve.go
```diff
@@ -233,41 +243,74 @@ func (re *reserver) source(src source) *sourceReserver {
...
+		reservedUTXOs = append(reservedUTXOs, u)
+		if reserved >= amount {
+			break
+		}
+	}
+	if reserved >= amount {
+		// We've found enough to satisfy the request.
+		for _, u := range reservedUTXOs {
+			sr.reserved[u.Outpoint] = rid
+			delete(sr.cached, u.Outpoint)
```
erykwalder: Should we just leave this until its no longer valid?

core/account/reserve.go
```diff
@@ -233,41 +243,74 @@ func (re *reserver) source(src source) *sourceReserver {
...
+		sr.mu.Unlock()
+		return reservedUTXOs, reserved, nil
+	}
+	sr.mu.Unlock()
+	reserved = 0
+	reservedUTXOs = nil
+
 	// Find the set of UTXOs that match this source.
-	utxos, err := sr.findMatchingUTXOs(ctx)
+	utxos, err := findMatchingUTXOs(ctx, sr.db, sr.src)
```
erykwalder: If we didn't delete reserved utxos, we could add a cursor here, and only load new utxos. We could even only do that if the processor pin height is greater than the last cursor height.

erykwalder:
Wanna try retriggering wercker?

erykwalder:
LGTM

PR #258, opened 2016-11-21
sdk/java: add Client.Builder

erykwalder:
This introduces a builder for client objects. Certain parameters should be set at the start and never again, e.g. connection pool details. This allows those to be set in a clean manner.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -245,6 +245,16 @@ public class Client {
   }
 
   /**
+   * Sets the max idle http connections and how long they should be kept alive.
+   * @param maxIdle the maximum number of idle connections to keep in the pool
+   * @param keepAliveDuration the number of time units an idle connection is kept open in the pool
+   * @param unit the unit of time
+   */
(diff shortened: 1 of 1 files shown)


erykwalder:
PTAL

jbowens:
We might need to document that you must set the connection pool _before_ sharing the SDK client. From the OkHttp 2.x docs:
> Instances of OkHttpClient are intended to be fully configured before they're shared - once shared they should be treated as immutable and can safely be used to concurrently open new connections. If required, threads can call clone() to make a shallow copy of the OkHttpClient that can be safely modified with further configuration changes.

Also, if we update to OkHttp 3.x at some point, we'll need to supply the connection pool when creating the `OkHttpClient`. OkHttp 3.x moves `setConnectionPool` onto a `OkHttpClient.Builder` class and makes `OkHttpClient` immutable.

jeffomatic:
Probably best to make the connection pool size a constructor variable in that case. Maybe we can follow suit and have a Client.Builder that allows you to set the connection pool size. We can keep the existing constructors intact.

erykwalder:
something like this?

jeffomatic:
LGTM

PR #259, opened 2016-11-21
core: guard against nil templates

### LINKED ISSUE #128: Submitting a transaction with nil value crashes core
#### Environment 
1. Chain Core Version: latest build
2. Chain Core SDK Language and Version: Ruby
3. Host: local

#### What steps will reproduce the problem?
Attempt submitting a transaction with a `nil` value. If using the Ruby SDK, attempt submitting a tx such as `client.transactions.submit(nil)` to replicate 

#### What result did you expect?
Expected a helpful error to return to point me in the right direction of what was causing it, and expected Chain Core to continue running

#### What result did you observe?
This crashed the locally running Chain Core, throwing the following error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14376a]
goroutine 343 [running]:
panic(0x64f240, 0xc42000c0a0)
	/usr/local/Cellar/go/1.7.1/libexec/src/runtime/panic.go:500 +0x1a1
chain/core.(*Handler).finalizeTxWait(0xc4201c80b0, 0x153e160, 0xc4202f97a0, 0xc4201ec0c0, 0x0, 0xc420ccb250, 0xc4202d6810)
	/Users/chrisgarvin/src/chain/core/transact.go:164 +0x3a
chain/core.(*Handler).submitSingle(0xc4201c80b0, 0x153e160, 0xc4202f97a0, 0xc4201ec0c0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/chrisgarvin/src/chain/core/transact.go:102 +0x116
chain/core.(*Handler).submit.func1(0xc4201c80b0, 0x153e1a0, 0xc4202d6810, 0xc4202dab00, 0x1, 0x4, 0x0, 0xc420ccb1f0, 0x1, 0x1, ...)
	/Users/chrisgarvin/src/chain/core/transact.go:259 +0xec
created by chain/core.(*Handler).submit
	/Users/chrisgarvin/src/chain/core/transact.go:267 +0x16a
```

jbowens:
Deserialize into `[]txbuilder.Template` instead of `[]*txbuilder.Template`
This prevents the need for an explicit check for a nil template if
`null` is provided instead of a transaction template JSON object.

Instead, txbuilder.ErrMissingRawTx will be returned:
https://github.com/chain/chain/blob/main/core/errors.go#L125

Fixes #128.

### DIFF
--- a/core/transact.go
+++ b/core/transact.go
@@ -267,7 +267,7 @@ func waitForTxInBlock(ctx context.Context, c *protocol.Chain, tx *bc.Tx, height 
 }
 
 type submitArg struct {
-	Transactions []*txbuilder.Template
+	Transactions []txbuilder.Template
 	wait         chainjson.Duration
 	WaitUntil    string `json:"wait_until"` // values none, confirmed, processed. default: processed
 }
@@ -290,7 +290,7 @@ func (h *Handler) submit(ctx context.Context, x submitArg) (interface{}, error) 
 			defer batchRecover(subctx, &responses[i])
 
 			tx, err := h.submitSingle(subctx, submitSingleArg{
-				tpl:       x.Transactions[i],
+				tpl:       &x.Transactions[i],
 				wait:      x.wait,
 				waitUntil: x.WaitUntil,
 			})


erykwalder:
Should a different error be returned?

jbowens:
> Should a different error be returned?

IMO, we should avoid separate errors because this error can only be exercised if there's a bug in the SDK. Maybe we could make the existing error less specific?

PR #260, opened 2016-11-21
cmd/cored: use 'prod' tag to determine production mode

jeffomatic:
When determining whether cored is in "production" mode, there is
some confusion between the build tags "prod" and "buildTag". This
commit uses "prod" to determine production status, which frees
"buildTag" to convey version information for official releases.


jeffomatic:
PTAL @erykwalder @boymanjor

Issue #239, opened 2016-11-17
Breaking changes to the Protocol 1


oleganza:
This is a draft of a list of breaking (hard forking) changes for P1 before we fully commit to it long-term.

Each feature has its own PR open which is meant to be merged when fully implemented and covered with tests.

### ☑️ ~1. Asset ID without version~ 

Asset ID should not include version tag to allow their reuse in new version of outputs. Initially it was thought as a necessary security feature for non-upgraded nodes, but balancing rules are enough to force assets to balance in such case. And removing version tag allows nice reuse of already deployed asset ids in the new versions of outputs.

PR: https://github.com/chain/chain/pull/261

### ☑️ ~2. Drop RIPEMD160 and SHA1~

They are unnecessary, outdated, and (in the case of SHA1, anyway) weak.

PR: https://github.com/chain/chain/pull/268

### ☑️ ~3. Fixed-length outpoint format~

`SHA3(txid || index || SHA3(output))` - constant size, commits to the output contents directly for HSM-friendliness.

PR: https://github.com/chain/chain/pull/417

### ☑️ ~4. Get rid of all _Optional Hashes_~

Optional Hash is a premature optimization that mixes two concerns: minimizes amount of data to transmit when the hashed string is actually empty, and minimizes amount of data to be hashed. The first one belongs to the transport protocol that can compress common patterns (we have an assumption that such protocol will exist already for other pieces). The second concern (hashing a bit more data) seems to not a serious overhead VS added complexity to implementation.

This is going to be a side effect of the new txgraph redesign. See below.

### 5. Transaction entries

Replace two lists of inputs/outputs with a single flat list of _entries_. Entries can be issuances, inputs, outputs or retirements. We can remove `OP_FAIL` opcode. We can have multiple "reference data" entries, so each party in the transaction can attach that data. Arbitrary accounting rules and behaviours can be added with future entry types.

PR: https://github.com/chain/chain/pull/264

### 6. Review the implementation to check if NOPs are disallowed in tx v1 and allowed in unknown versions.

We need to carefully review implementation of extensibility rules.

### 7. Refactor tx validation spec (compatible change) 

Change validate+validate-well-formedness+apply to validate+apply procedures for txs. Same for block.

### 8. THRESHOLD instruction

It is extra-hard to combine programs in multisig manner which is necessary for consensus programs. Since block VM version is hard to change, we should have a workable THRESHOLD instruction to make it possible to combine individual programs. This will not eliminate, but augment CHECKMULTISIG.

[...]

PR #263, opened 2016-11-22
protocol: use explicit VERIFY in all programs

oleganza:
Currently VM does implicity VERIFY instruction in the end of each program execution. This is similar to how Bitcoin scripts work, but breaks nice symmetry and makes consensus code needlessly more complex. This PR removes implicit VERIFY, so the programs must add it explicitly. The execution succeeds simply if the program did not fail. Stack can be empty or not.


kr:
LGTM

danrobinson:
I think this is a good idea, but I'm wondering again if it should wait for a new VM version. It would require changes to every single control program (including ones hardcoded into Chain Core but also for ones on existing blockchain networks) for what amounts to an aesthetic improvement.

PR #265, opened 2016-11-22
installer/mac: add configuration for development builds (DevRelease)

oleganza:
This adds a DevRelease config in addition to Release and Debug. It is intended for development builds that we can distribute to beta-testers.


kr:
This is pretty stale. Closing for now. We can always reopen later if we want.

PR #266, opened 2016-11-22
installer/mac: add sha256 checks to dependencies

oleganza:
This cleans up unused dependencies and adds mandatory sha256 checksum verification to all downloadable dependencies. This allows us not to store large blobs in this repository and at the same time make sure the integrity of the build is verified.


installer/mac/tools/renamesha256.rb
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env ruby
```
kr: This line makes me hesitate a little bit only because it'd be nice not to introduce too many dependencies for building Chain Core, and ruby is a pretty significant new dep here. (As far as I know it's new.)

This ruby script in particular pretty much just calls out to the shell, so maybe it would be clearer as a shell script?

```bash
#!/bin/bash
set -eo pipefail
digest=`shasum -a 256 "$1"|cut -b-64`
mv "$1" "$1"-$digest
```

installer/mac/tools/renamesha256.rb
```diff
@@ -0,0 +1,16 @@
...
+# Renames `filename.tar.gz` into `filename.tar.gz-<sha256digest>`
+path = ARGV[0]
+if !File.exist?(path)
+    $stderr.puts "File not found: #{path}"
+    exit(1)
+end
+digest = `shasum -a 256 #{path}`[0,64]
+
+name = File.basename(path) + "-#{digest}"
+newpath = File.dirname(path) + "/" + name
```
kr: Aren't these two lines the same as `newpath = path + "-#{digest}"`?

installer/mac/pg/src/makefile
```diff
@@ -82,15 +75,16 @@ $(SHAREDPREFIX)/lib/libssl.dylib: openssl-${OPENSSL_VERSION}/Makefile
...
-openssl-${OPENSSL_VERSION}/Configure: openssl-${OPENSSL_VERSION}.tar.gz
-	$(TAR) openssl-${OPENSSL_VERSION}.tar.gz
+openssl-${OPENSSL_VERSION}/Configure: openssl-${OPENSSL_VERSION}.tar.gz-${OPENSSL_SHA256}
+	$(TAR) openssl-${OPENSSL_VERSION}.tar.gz-${OPENSSL_SHA256}
 	touch $@
 
-openssl-${OPENSSL_VERSION}.tar.gz:
+openssl-${OPENSSL_VERSION}.tar.gz-${OPENSSL_SHA256}:
 	$(CURL) "https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz"
+	ruby ../../tools/renamesha256.rb openssl-${OPENSSL_VERSION}.tar.gz
```
kr: Consider using the environment variable `$CHAIN` to avoid `../..` in paths. It's a bit easier for me to read `$(CHAIN)/installer/mac/tools/renamesha256.rb` because it doesn't require me to remember which file I'm looking at and where it is on disk.

oleganza:
Changed the renamesha256 script from Ruby to to bash, per @kr's suggestion. Also cleaned up invocation paths by using `$(TOOLS)` prefix. I still prefer keeping global variable to a minimum, hopefully this version of the makefile will be readable enough :)

kr:
> I still prefer keeping global variable to a minimum, 

If it helps, think of it as a global constant. :trollface: 

(But for real, it isn't effectively mutable during the execution of build scripts, so it's more like a global constant, and there's nothing wrong with those.)

oleganza:
I mean, good deal of trouble with the mac app was disentangling all sort of global-path assumptions scattered all across postgres codebase. I don't want to recreate that minefield in our project for no good reason :) IMHO, relative paths is how it should be done _right_ in any software package, and readability is easier solved with flatter directory structures and helper variables to label common paths.

kr:
LGTM

PR #267, opened 2016-11-22
protocol: make VM numbers fixed-length when placed on the stack

oleganza:
In the transaction format, all integers are varints, but on the stack they should be fixed-length 8-byte signed integers for simplicity of implementation and security of smart contracts: fixed-length integers can be concatenated with other data without extra precautions like length-prefixing or extra hashing.

This also removes `NUMEQUAL`, `NUMEQUALVERIFY`, `NUMNOTEQUAL` since they become redundant when integers have always canonical representation.

### DIFF
--- a/docs/protocol/specifications/vm1.md
+++ b/docs/protocol/specifications/vm1.md
@@ -197,24 +197,20 @@ Certain arithmetic operations use conservative bounds checks (explicitly specifi
 
 #### String to Number
 
-1. If the string is longer than 8 bytes, fail execution.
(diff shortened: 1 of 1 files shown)


danrobinson:
Agree on the change.

Maybe drop the `NUMEQUAL`, `NUMEQUALVERIFY`, `NUMNOTEQUAL` opcodes, now that they are redundant.

oleganza:
@danrobinson good call.

oleganza:
Need to make sure booleans are defined as 8-byte ints and coercion from strings still works the same way.

PR #268, opened 2016-11-22
protocol: remove legacy hash functions RIPEMD-160 and SHA1

oleganza:
These hash functions are unnecessary, outdated, and (in the case of SHA1, anyway) weak.


kr:
LGTM

oleganza:
@kr @danrobinson PTAL. Implementation is updated with RIPEMD160 and SHA1 removed.

WARNING: this is a hard-forking change to the protocol.

kr:
:eyes:

kr:
LGTM (we should also probably follow up by removing directory vendor/golang.org/x/crypto/ripemd160).

oleganza:
@kr I removed `ripemd160` package in this PR for completeness.

PR #269, opened 2016-11-22
core: separate indexer from block processing

erykwalder:
Previously, block processing for assets or accounts required the
presence of an indexer. However, both the asset and account package have
crucial pieces to manage in the block processor outside of calling the
indexer.

This separates those processors from the indexer and the
INDEX_TRANSACTIONS environment variable.


erykwalder:
PTAL

bobg:
:eyes:

bobg:
LGTM

PR #270, opened 2016-11-22
REPLACED protocol: [hardfork] fixed-length OutputID replacing Outpoint

oleganza:
# THIS PR IS REPLACED BY https://github.com/chain/chain/pull/417

---

**Problem**

Outpoint is a variable-length structure `<txid>:<index>` which is 33-40 bytes long (33 bytes for most transactions). It is used by transaction inputs to identify exact output in the UTXO set ("Assets Merkle Tree"). The tree leafs contain `SHA3(output)` which allows save space and requires transactions to carry redundant copies of spent outputs to perform validation (otherwise nodes would have to store the entire outputs instead of their hashes — over 2x more data, and the ratio is much bigger in protocol 2). Also, for HSM-friendliness the TXSIGHASH must contain a redundant output's hash: `SHA3(txid || input index || SHA3(output))`.

**Proposal:**

We define two new terms:

* `OutputID = SHA3(TxHash || OutputIndex)`
* `UnspentID = SHA3(OutputID || SHA3(OutputCommitment))`

How are these used:

1. Transaction input contains **OutputID** to identify the output being spent. This is a unique identifier of the output.
2. Transaction input uses **second serialization flag** to indicate if it contains the entire previous Output Commitment, or its hash (instead of empty place).
3. UTXO set becomes a **proper set** containing **UnspentIDs** instead of `{Outpoint -> SHA3(OutputCommitment)}`.

When a node validates a transaction, it computes `UnspentID` using provided `OutputID` and previous `OutputCommitment`. If the given unspent ID is present in the UTXO set, then previous output is proved to be both authentic and available for spending.

**The upsides:**

1. The outputID is constant-size and shorter: 32 bytes instead of 33-40 bytes. This simplifies merkle tree design, transaction data structure and all pieces of software that need to handle outpoints.
2. All outputs (via unspentIDs) in the transaction are randomized across the Assets Merkle Tree instead of being crammed inside a common subpath `<txid>||...`.
3. Inputs automatically commit directly to the spent outputs, so TXSIGHASH does not need to do that and can be simplified to `SHA3(txid || input index)`. HSM is able to verify which output this input commits to without having access to the entire parent transaction.
4. We keep the term _outpoint_ to mean a pair `(txid, index)`, but is internal to Chain Core to support random access to UTXOs. Validation protocol no longer uses outpoints.
5. UTXO takes 2x less RAM because it only contains unpent IDs (32 bytes) instead of a key-value pair (64+ bytes).

**The downside:**

1. OutputID no longer indicates the transaction ID which makes it impossible to navigate the chain of transactions without also having a mapping `outpoint -> txid:index`. UTXO tree is not enough as it's only reflecting the latest state of the chain and throws away spent outpoints. Note that in order to navigate the transactions in practice one still needs the mapping `txid -> tx`, so maintaining one more index might not be a significant increase in complexity. Chain is doing this indexing already and we keep that mapping.

PS. On a second thought, we can redefine OutputID back to Outpoint since we do not really use txid:index as a structure after this change. And then maybe UnspentID can become an OutputID.


bobg:
I would prefer to separate the protocol change PR from the code PR.

oleganza:
@bobg why exactly? With a single PR we can keep the spec in sync with implementation. File-wise, specs and code are sitting in separate folders anyway, so there's separation.

oleganza:
@bobg i've also figured to not redefine "Outpoint", but instead add "OutputID" only because I worked on the actual code and noticed that we have a lot of places around API where we'd like to keep Outpoint as-is. So I could change the spec within the same PR without being mentally committed to the original plan.

protocol/bc/txinput.go
```diff
@@ -53,22 +53,21 @@ type (
 var errBadAssetID = errors.New("asset ID does not match other issuance parameters")
 
 func NewSpendInput(txhash Hash, index uint32, arguments [][]byte, assetID AssetID, amount uint64, controlProgram, referenceData []byte) *TxInput {
+	av := uint64(1)
```
bobg: Why put this in a variable? If anything it should be a `const`, but a literal is best IMO for clarity.

bobg:
> why exactly?

1. it's been our MO so far
2. smaller PR better
3. review for each can be more focused
4. separation of protocol spec and implementation improves both (i.e., I think it's a better idea if you don't do both, at least not at the same time)

protocol/bc/txinput.go
oleganza: Yup. Fixed: https://github.com/chain/chain/pull/270/commits/abec806c1699743a4b4fae42191f10fd47749d44

bobg:
> we have a lot of places around API where we'd like to keep Outpoint as-is

Where keeping the Outpoint API as-is would conceal a new database access that didn't exist before, I would argue against trying to keep it as-is to make the cost clear.

oleganza:
Re: 4. I agree, that's why each of these PRs started as a complete spec with no code. But then code informs us to do some tweaks that we might want to do in the spec.

Re: 3. I think review can be better informed about complexity and sanity of the change if we can see how proposed spec impacts actual code. Alternatively, we could have two PRs for spec and code separately, but prepared for review simultaneously, instead of committing to one, then coding up the other.

protocol/bc/txinput.go
```diff
@@ -63,10 +62,10 @@ func NewSpendInput(txhash Hash, index uint32, arguments [][]byte, assetID AssetI
 		ControlProgram: controlProgram,
 	}
 	return &TxInput{
-		AssetVersion:  av,
+		AssetVersion:  assetVersion,
```
bobg: I still prefer a literal here. This function produces a version 1 txinput only, not whatever kind of txinput `assetVersion` is defined to be in the future.

oleganza:
> Outpoint API as-is would conceal a new database access that didn't exist before.

My plan is to keep existing DB access and only add another column "outputID" that can be used by verification code, while keeping "txid,index" columns for API code.

If it makes sense to get rid of outpoint entirely, we can do that in the following PR that removes "txid,index" pair from DB.

WDYT?

protocol/bc/txinput.go
oleganza: Fixed: https://github.com/chain/chain/pull/270/commits/31dc70869117321766a0b2029df891c8d39b6ae0

kr:
(Personally in this case I'm ok with doing the spec changes separately from the implementation, or together. The PR isn't *that* big as-is, but it would indeed be easier to review if split up. And while it's nice to have the spec and implementation stay in sync, it hasn't been a problem before to do these changes sequentially.)

oleganza:
@bobg @kr Oh, I've discovered a spec deviation. TxInput always writes the entire prevoutput without checking the serialization flags. 

* Per current rules, it should omit the prevout entirely in the context of txid serialization. 
* Per this PR's rules it should switch between writing the prevout as-is and writing _it's hash_ if we are computing txid.

bobg:
> TxInput always writes the entire prevoutput without checking the serialization flags

By my reading of the spec, that's correct. It's the input's witness data that may be omitted depending on the serialization flags, but the prevout is in the input's commitment.

Is my understanding incorrect?

oleganza:
@bobg i think you misinterpreted the spec (sorry it was unclear!). 

Here: https://github.com/chain/chain/blob/main/docs/protocol/specifications/data.md#transaction-serialization-flags

> The second bit indicates whether the output commitment from the spent output is present in the input spend commitment. If set to zero, the output commitment field is absent.

And about txid where 0x00 flags are coming from:

> The transaction ID (also called txid or transaction hash) is defined as SHA3-256 of the transaction serialized with 0x00 serialization flags.

oleganza:
Replaced by https://github.com/chain/chain/pull/417

PR #271, opened 2016-11-22
dashboard: empty state educational content for objects

chrisgarvin:


chrisgarvin:
Original:
image

Modified example:
image

chrisgarvin:
PTAL

dominic:
I think the border is a little inconsistent with the existing designs. We don't seem to use borders when there's only one section of content and no background color difference.

dashboard/src/features/shared/components/BaseList/BaseList.jsx
```diff
@@ -10,6 +10,7 @@ class ItemList extends React.Component {
   render() {
     const label = this.props.label || pluralize(humanize(this.props.type))
     const objectName = label.slice(0,-1)
+    const showWelcome = this.props.showWelcome || false
```
dominic: If we put the `h2` for `showWelcome` in the empty content for the transaction list, we can use one less prop and bit less code.

dashboard/src/features/unspents/components/List.jsx
```diff
@@ -1,12 +1,29 @@
...
+  <p>
+    You do not have any unspent outputs.
+  </p>
+  <p>
+    An output is considered unspent when it has not yet been used as an input
+    to a new transaction. All asset units on a blockchain exist in the unspent output set.
+  </p>
+  <p>
+    Learn more about the basic functions of unspent outputs in the&nbsp;
+    <a href="/docs/core/build-applications/accounts" target="_blank">Unspent Outputs</a> guide
```
dominic: This should be https://chain.com/docs/core/build-applications/unspent-outputs

chrisgarvin: good catch, thanks @dominic

jeffomatic:
I added recommendations for content here: https://trello.com/c/06A1SRH7/344-empty-state

chrisgarvin:
Updated:
image

image

dashboard/src/features/shared/components/BaseList/EmptyList.jsx
```diff
@@ -13,28 +13,17 @@ class EmptyList extends React.Component {
...
-      emptyBlock = <div>
-        <span className={styles.emptyLabel}>
-          There are no {this.props.objectName.toLowerCase()}s on the blockchain
-        </span>
-      </div>
+      emptyBlock = <div>{this.props.emptyContent}</div>
     } else if (this.props.showFirstTimeFlow) {
       emptyBlock = <div>
-        <span className={styles.emptyLabel}>Create your first {this.props.objectName}</span>
+        {this.props.emptyContent}
```
dominic: I think we should come up with a better prop name for this, or make `emptyContent` and object so that it's rendered as:

```
{this.props.emptyContent.firstTime}
```

dashboard/src/features/shared/components/BaseList/EmptyList.jsx
```diff
@@ -13,28 +13,17 @@ class EmptyList extends React.Component {
...
         {this.props.newButton}
       </div>
     } else if (!this.props.showFirstTimeFlow) {
       classNames.push(styles.noResults)
-      emptyBlock = <div>
+      emptyBlock = <div className={styles.emptyContainer}>
         <span className={`${styles.emptyLabel} ${styles.noResultsLabel}`}>No results for query:</span>
         <code className={styles.code}>{this.props.currentFilter.filter}</code>
-        <div className={styles.queryHelp}>
```
dominic: I might have missed a discussion, but why are we removing the doc links from this branch? It's a [relatively new addition](https://github.com/chain/chain/pull/223).

dashboard/src/features/accessTokens/components/List.jsx
```diff
@@ -1,17 +1,33 @@
+import React from 'react'
 import { BaseList, TableList } from 'features/shared/components'
 import Item from  './ListItem'
 import actions from 'actions'
 
 const clientType = 'client_access_token'
 const networkType = 'network_access_token'
 
-const stateToProps = (type) => (state, ownProps) =>
+const emptyContentClient = <div className="emptyContainer">
```
dominic: (I realize this branch is pretty far along so I am not married to this - just want to keep it in mind)

dashboard/src/features/shared/components/BaseList/EmptyList.jsx
dominic: Ah ha! Gotcha - @jeffomatic, can you weigh in?

devongundry:
In the welcome to Chain Core, the "learn more" text feels a little odd. Like Were instructing them to Do something but don't tell them where. Maybe "...submit transactions in the documentation", where documentation is the link?

Or maybe just make "learn more" the link?

chrisgarvin:
@devongundry I see what you're saying there. Each of the empty state views has the object name as the link such as the following:

image

Would you suggest adding "in the documentation" to each of these views? Here's what it would look like using "Learn more" as the link (which we could do in each view without adding text):

image

devongundry:
I like the learn more as a link, since it's a call to action

dashboard/src/features/accounts/components/List.jsx
```diff
@@ -1,14 +1,23 @@
...
+</EmptyContent>
+
 export default BaseList.connect(
   BaseList.mapStateToProps(type, ListItem, {
     wrapperComponent: TableList,
     wrapperProps: {
       titles: ['Account Alias', 'Account ID']
-    }
+    },
+    firstTimeContent: emptyListContent
```
dominic: Approved - but let's rename the variable `emptyListContent` to `firstTimeContent`. Then we can use it in the object like this:

```js
       ...
    },
    firstTimeContent
  }),

dominic: (Where applicable - it doesn't work so cleanly for access tokens).

PR #272, opened 2016-11-22
core/account: fix paged reservation

erykwalder:
Since the findMatchingUTXOs function only returns new UTXOs, any UTXOs
identified in the cache should be retried in a reservation attempt.

### DIFF
--- a/core/account/reserve.go
+++ b/core/account/reserve.go
@@ -274,6 +274,7 @@ func (sr *sourceReserver) reserve(ctx context.Context, rid uint64, amount uint64
 	for o, u := range sr.cached {
 		// If the UTXO is already reserved, skip it.
 		if _, ok := sr.reserved[u.Outpoint]; ok {
+			unavailable += u.Amount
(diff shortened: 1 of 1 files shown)


erykwalder:
PTAL @jbowens

jbowens:
In the case where there's a not enough in the cache, I wanted to avoid keeping the cached UTXOs locked while querying the DB. We don't actually reserve the cached UTXOs unless there's enough to satisfy the request.

erykwalder:
Bleh. right. Ok, fix coming in

erykwalder:
updated, PTAL

jbowens:
Is this just a refactoring? I think we do retry cached UTXOs after the cache fill, right?

jbowens:
Ah, nvm, I see.

erykwalder:
Previously we only ranged over the new utxos for the second attempt.

jbowens:
lgtm

PR #274, opened 2016-11-22
core/account: use pin for paging utxos

erykwalder:
This maintains the ability to grab only new UTXOs while introducing the ability to skip a database query altogether by seeing if any additional work has been done.


jbowens:
🆒

PR #275, opened 2016-11-22
cmd/build-cored-release: new command

jeffomatic:
build-cored-release compiles or cross-compiles cored and corectl
at the specified ref. Its main use case is within automated build
scripts that generate installer packages.

Usage is as follows:

```
build-cored-release <releaseRef> <outputPath>
```

- releaseRef must be a valid git ref in the locally-cloned repo.
- outputPath must be a valid absolute or relative directory.

The buildTag build variable is set to releaseRef.

Optional parameters are specified via environment variables:

- GOOS
- GOARCH
- DATE: a value for the buildDate build variable. By default, the current time
  is used.

### DIFF
--- a/bin/build-cored-release
+++ b/bin/build-cored-release
@@ -18,7 +18,10 @@ if [ -z "$1" -o -z "$2" ]; then
 fi
 
 releaseRef=$1
-cd $2; outputDir=`pwd` # verify output dir and expand to absolute
+
+# verify output dir and expand to absolute
+cd $2 > /dev/null
+outputDir=`pwd`
 
 echo "Building ref $releaseRef to output directory $outputDir..."
 
@@ -55,3 +58,5 @@ echo "building corectl..."
 go build\
   -o "$outputDir/corectl"\
   chain/cmd/corectl
+
+echo "Done, build artifacts placed in $outputDir"
--- a/bin/build-cored-release
+++ b/bin/build-cored-release
@@ -36,6 +36,7 @@ cd $newChain
 git checkout $releaseRef
 
 # Cross-compilation args for cored and corectl
+export CGO_ENABLED=0 # There are issues with cgo (e.g. https://github.com/lib/pq/issues/395), so disable it entirely.
 export GOPATH=$buildGoPath
 export GOOS
 export GOARCH
--- a/bin/build-cored-release
+++ b/bin/build-cored-release
@@ -43,10 +43,9 @@ export GOARCH
 
 echo "building cored..."
 
-version=`echo $releaseRef | sed -e s/^cmd\.cored-//` # version is extracted if releaseRef fits cmd.cored-<version> pattern
 commit=`git rev-parse HEAD`
 DATE=${DATE:-`date +%s`} # date can be set via envvar
-ldflags="-X main.buildTag=$version -X main.buildCommit=$commit -X main.buildDate=$DATE"
+ldflags="-X main.buildTag=$releaseRef -X main.buildCommit=$commit -X main.buildDate=$DATE"
 
 go build\
   -tags 'insecure_disable_https_redirect'\
--- a/bin/build-cored-release
+++ b/bin/build-cored-release
@@ -1,0 +1,57 @@
+#!/bin/sh
+set -e
+
+usage() {
+  echo "Usage: $0 <releaseRef> <outputDir>"
+  echo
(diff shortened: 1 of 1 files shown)


jeffomatic:
PTAL

jeffomatic:
@boymanjor @tessr @oleganza Hoping that this tool can be used to generate `cored` for the Mac/Windows installers and for the Docker image.

kr:
If we're going to make the buildTag be the version string instead of the actual tag name, we should probably rename that variable :/ (sorry this didn't occur to me earlier)

jeffomatic:
@kr I left it as "buildTag" to indicate that it could be something else, but perhaps it is better to rename it to the thing it will primarily be used for, which is conveying the version.

Alternately, I can just avoid stripping the `cmd.cored-` prefix out, and we can just use the whole Git tag as the build tag no matter what. WDYT?

bin/build-cored-release
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
+set -e
```
boymanjor: If we use `set -eu` the `u` will fail the script if any variables are unset. I think its a good case to always protect against.

bin/build-cored-release
```diff
@@ -0,0 +1,57 @@
...
+newChain=$buildGoPath/src/chain
+mkdir -p $newChain
+git clone $CHAIN $newChain
+cd $newChain
+git checkout $releaseRef
+
+# Cross-compilation args for cored and corectl
+export GOPATH=$buildGoPath
+export GOOS
+export GOARCH
```
boymanjor: Currently, I don't think this script will fail if `GOOS` and `GOARCH` are not set. You can assert them with the following statements:
```
: ${GOOS:?"some error text"}
: ${GOARCH:?"some error text"}
```
I also don't think you need to re-export them since they are passed in as env vars.

boymanjor: Optionally, you could do a similar thing to assert that the arguments `$1` and `$2` are passed to the script.

bin/build-cored-release
```diff
@@ -0,0 +1,57 @@
...
+buildGoPath=`mktemp -d`
+trap "rm -rf $buildGoPath" EXIT
+
+newChain=$buildGoPath/src/chain
+mkdir -p $newChain
+git clone $CHAIN $newChain
+cd $newChain
+git checkout $releaseRef
+
+# Cross-compilation args for cored and corectl
```
boymanjor: Oh, also we've been disabling CGO in scripts building `cored`. Not sure if this has been resolved. Any insight @kr?
```
# Disable cgo entirely.
# We have seen one too many problems with cgo,
# including an occasional (rare) malloc failure.
# See https://github.com/lib/pq/issues/395.
export CGO_ENABLED=0
```

bin/build-cored-release
boymanjor: Oh, I just realized `GOOS` and `GOARCH` are optional...my bad.

kr:
> Alternately, I can just avoid stripping the cmd.cored- prefix out, and we can just use the whole Git tag as the build tag no matter what. WDYT?

I like this. It's nice to be able to cut and paste the build tag output to look up the actual hit tag. We can strip off the version string as additional output in cored.

kr:
s/hit tag/git tag/

jeffomatic:
> I like this. It's nice to be able to cut and paste the build tag output to look up the actual hit tag. We can strip off the version string as additional output in cored.

Great. Given that, I think the following makes sense for a subsequent PR in cored itself:

- Retain "buildTag" as a sufficiently generic name for this build variable
- Add a new expvar and field to `/info` that reflects the version. It will parse "cmd.core-x.y.z" for this, or be "?" otherwise.

bin/build-cored-release
jeffomatic: I had `-u` originally, but wanted support for optional envvars.

bin/build-cored-release
jeffomatic: 👍 Will add.

jeffomatic:
PTAL:

- @boymanjor: I added the CGO disabling, but kept the ability to have optional vars. Lemme know how that sits with you.
- @kr: This command leaves the tag/ref alone. I made a separate PR that slurps the version string and advertises it via /info; see https://github.com/chain/chain/pull/285

PR #279, opened 2016-11-22
encoding/bufpool: add bytes.Buffer pool

erykwalder:
Many of the allocations in the protocol package come from creating byte
buffers. These buffers can use sync.Pool to reduce allocations. In a
recent run of benchcore, InputCommitmentBytes, inputWitnessBytes, and
OutputCommitment.writeTo in particular had many allocations from
initializing byte buffers.

### DIFF
--- a/encoding/bufpool/bufpool.go
+++ b/encoding/bufpool/bufpool.go
@@ -1,0 +1,22 @@
+// Package bufpool is a freelist for bytes.Buffer objects.
+package bufpool
+
+import (
+	"bytes"
+	"sync"
+)
+
+var pool = &sync.Pool{New: func() interface{} { return bytes.NewBuffer(nil) }}
+
+// GetBuffer returns an initialized bytes.Buffer object.
+// It is like new(bytes.Buffer) except it uses the free list.
+// The caller should call PutBuffer when finished with the returned object.
+func GetBuffer() *bytes.Buffer {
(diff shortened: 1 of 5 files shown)


encoding/bufpool/bufpool.go
```diff
@@ -0,0 +1,22 @@
+// Package bufpool is a freelist for bytes.Buffer objects.
+package bufpool
+
+import (
+	"bytes"
+	"sync"
+)
+
+var pool = &sync.Pool{New: func() interface{} { return bytes.NewBuffer(nil) }}
```
kr: Maybe consider `new(bytes.Buffer)` which produces the same result with less machinery.

encoding/bufpool/bufpool.go
```diff
@@ -0,0 +1,22 @@
...
+
+// GetBuffer returns an initialized bytes.Buffer object.
+// It is like new(bytes.Buffer) except it uses the free list.
+// The caller should call PutBuffer when finished with the returned object.
+func GetBuffer() *bytes.Buffer {
```
kr: Considering the name of this package, how about `Get` ?

erykwalder:
Holding off until some of the returns/other things are sorted out. It is not safe to hold the result of `buf.Bytes()` if you are reusing the buffer.

PR #280, opened 2016-11-22
sdk/java: remove action level ttl setters

boymanjor:
4e503a6dffd53dc2f2d5a735d892507323f5f9fa moved the TTL param
from actions to the tx level. While the SDK was updated to expose this
param to the user, the action level setters were never removed. Each
setter is effectively a noop and can be removed.


boymanjor:
@jeffomatic PTAL

PR #281, opened 2016-11-22
protocol/bc: use pointers for func receivers

erykwalder:
The AssetAmount and Outpoint WriteTo functions were previously declared
on raw structs. Each call to these functions would cause an allocation.
In a recent run of benchcore these two function calls collectively
allocated 4.5% of all objects.

```
benchmark                         old ns/op     new ns/op     delta
BenchmarkAssetAmountWriteTo-8     65.5          38.8          -40.76%
BenchmarkOutpointWriteTo-8        69.0          37.4          -45.80%

benchmark                         old allocs     new allocs     delta
BenchmarkAssetAmountWriteTo-8     1              0              -100.00%
BenchmarkOutpointWriteTo-8        1              0              -100.00%

benchmark                         old bytes     new bytes     delta
BenchmarkAssetAmountWriteTo-8     48            0             -100.00%
BenchmarkOutpointWriteTo-8        48            0             -100.00%
```

### DIFF
--- a/protocol/bc/transaction.go
+++ b/protocol/bc/transaction.go
@@ -355,7 +355,7 @@ func (p Outpoint) String() string {
 
 // WriteTo writes p to w.
 // It assumes w has sticky errors.
-func (p Outpoint) WriteTo(w io.Writer) (int64, error) {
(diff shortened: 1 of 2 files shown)


jbowens:
lgtm

PR #282, opened 2016-11-23
cmd/benchcore: use different instance types

jbowens:
Provide different EC2 instance types and update the current
Postgres configuration to use I/O-optimized EC2 instances.

This also allows us to get rid of the EBS-mounted volume because the
storage-optimized instances come with adequate storage already.

Instead of using `-instance m4.16xlarge`, you use `-config max`.

### DIFF
--- a/cmd/benchcore/main.go
+++ b/cmd/benchcore/main.go
@@ -70,7 +70,7 @@ var instanceConfigs = map[string]instanceConfig{
 	"default": instanceConfig{
 		AMI:                     "ami-40d28157", // Ubuntu LTS 16.04
 		InstanceType:            "m3.xlarge",
-		CoredMaxDBConns:         "500",
+		CoredMaxDBConns:         "1000",
 		PostgresAMI:             "ami-2ef48339", // Ubuntu Server 16.04 LTS (HVM), SSD Volume Type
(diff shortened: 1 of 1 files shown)


jbowens:
PTAL

cmd/benchcore/main.go
```diff
@@ -69,35 +67,49 @@ var (
...
 		AMI:                     "ami-40d28157", // Ubuntu LTS 16.04
+		InstanceType:            "m3.xlarge",
 		CoredMaxDBConns:         "500",
-		MaxConnections:          530,
-		SharedBuffers:           "2GB",
-		EffectiveCacheSize:      "11GB", // ~3/4 total mem
-		WorkMem:                 "8MB",
+		PostgresAMI:             "ami-2ef48339", // Ubuntu Server 16.04 LTS (HVM), SSD Volume Type
+		PostgresInstanceType:    "i2.xlarge",
+		MaxConnections:          1030,
```
erykwalder: Any reason for increasing the max without increasing the cored max?

erykwalder:
One comment, otherwise lgtm

cmd/benchcore/main.go
jbowens: Whoop

PR #283, opened 2016-11-23
database/pg: reuse slices in ForQueryRows

erykwalder:
Rather than allocating new slices for each row in the results, the slices
can be emptied and reused.


jbowens:
lgtm

PR #284, opened 2016-11-23
core/migrate: audit account_utxos indexes

jbowens:
Delete unused columns of the account_utxos table.
Delete the old account_utxos indexes and replace with one tailored
specifically for the reservation query. There are only three non-INSERT
queries we perform against the table:

```
DELETE FROM account_utxos
WHERE (tx_hash, index) IN (SELECT unnest($1::text[]), unnest($2::integer[]))
```

```
SELECT tx_hash, index, amount, control_program_index, control_program
FROM account_utxos
WHERE account_id = $1 AND asset_id = $2 AND confirmed_in > $3
```

```
SELECT account_id, asset_id, amount, control_program_index, control_program
FROM account_utxos
WHERE tx_hash = $1 AND index = $2
```

The exact UTXO lookups are already covered by the table's primary key.
We only need one additional index for the 2nd query. This change brings
the 2nd query's query time from:

    Total Minutes: 67.752444
    Average MS: 23.977507
    Calls: 169540

down to:

    Total Minutes: 2.563405
    Average MS: 0.955413
    Calls: 160982

### DIFF
--- a/core/account/indexer.go
+++ b/core/account/indexer.go
@@ -163,8 +163,6 @@ func (m *Manager) upsertConfirmedAccountOutputs(ctx context.Context, outs []*out
 		accountID pq.StringArray
 		cpIndex   pq.Int64Array
 		program   pq.ByteaArray
-		metadata  pq.ByteaArray
-		blockPos  pg.Uint32s
 	)
 	for _, out := range outs {
 		txHash = append(txHash, out.Outpoint.Hash.String())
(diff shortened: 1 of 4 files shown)


jbowens:
Now that I'm thinking about it, do we even need the `metadata`, `block_pos` or `block_timestamp` columns? They don't appear to be used anywhere now.

```
        Table "public.account_utxos"
        Column         |  Type   | Modifiers
-----------------------+---------+-----------
 tx_hash               | text    | not null
 index                 | integer | not null
 asset_id              | text    | not null
 amount                | bigint  | not null
 account_id            | text    | not null
 control_program_index | bigint  | not null
 control_program       | bytea   | not null
 metadata              | bytea   | not null
 confirmed_in          | bigint  | not null
 block_pos             | integer | not null
 block_timestamp       | bigint  | not null
```

erykwalder:
If we can get rid of those too, great!

PR #285, opened 2016-11-23
cmd/cored: use fixed version string for cored and expose via API

jeffomatic:
This commit exposes the most recent release version of `cored` via the /info endpoint.

The version string reported by the API takes on slightly differing formats depending on the build tag:

- `MAJOR.MINOR.BUILD`: if the build tag begins with `cmd.cored-`. This prefix denotes an official release.
- `MAJOR.MINOR.BUILD-<build tag>`: if any other non-blank build tag was provided
- `MAJOR.MINOR.BUILD-dev`: if the build tag is empty

This commit also moves the canonical `cored` version from Git tags to a value in the source itself. While this denormalizes the version across Git tags and the source code (worth nothing: this condition is shared by all other releasable packages, such as SDKs and installers), it means that `cored` version updates can be tracked reliably using Git history. Moreover, it allows intermediate builds (i.e., non-official releases and development builds) to report the most recent version via the client API, which is useful information in general.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -51,6 +51,7 @@ import (
 const (
 	httpReadTimeout  = 2 * time.Minute
 	httpWriteTimeout = time.Hour
+	latestVersion    = "1.0.1"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


jeffomatic:
updated to include latest version in all builds, and updated description

jeffomatic:
PTAL @dominic

cmd/cored/main.go
```diff
@@ -50,6 +51,7 @@ import (
 const (
 	httpReadTimeout  = 2 * time.Minute
 	httpWriteTimeout = time.Hour
+	latestVersion    = "1.0.1"
```
dominic: Should this be `1.0.2` since `1.0.1 ` is already released?

dominic:
LGTM

cmd/cored/main.go
jeffomatic: Per conversation: `1.0.1` is accurate here. This value represents the version we have already reached, not the one we are looking forward to. It will tick to `1.0.2` once we're satisfied that the release is ready.

PR #286, opened 2016-11-23
core/migrate: use jsonb_path_ops

jbowens:
The `annotated_txs.data` index was using a regular jsonb GIN index. We
should use the smaller and faster `jsonb_path_ops` specific to the
`@>` contains operator. The other jsonb indexes are already correctly
using `jsonb_path_ops`.

This provides a modest improvement to the `annotated_txs` insert query
too because there's less data to write for this index:

     Total Minutes: 3.791963
     Average MS: 360.567044
vs

     Total Minutes: 4.384654
     Average MS: 414.951517


erykwalder:
LGTM

PR #287, opened 2016-11-23
protocol/validation: use state.OutputKey for deleting

jbowens:
The state.OutputKey function should be preferred over
state.OutputTreeItem where the value hash isn't needed. It avoids
unnecessarily serializing the output commitment.


erykwalder:
LGTM

PR #288, opened 2016-11-23
protocol/patricia: only hash nodes on demand

erykwalder:
Previously, to avoid bugs, nodes in the tree computed their hash on
insert and delete. With a large number of inserts and deletes, this
duplicates a large amount of work.

```
benchmark              old ns/op     new ns/op     delta
BenchmarkInserts-4     142224708     36920148      -74.04%

benchmark              old allocs     new allocs     delta
BenchmarkInserts-4     283501         283499         -0.00%

benchmark              old bytes     new bytes     delta
BenchmarkInserts-4     17648974      15540852      -11.94%
```

### DIFF
--- a/protocol/patricia/patricia.go
+++ b/protocol/patricia/patricia.go
@@ -45,13 +45,14 @@ func Reconstruct(vals []Leaf) (*Tree, error) {
 	t := new(Tree)
 	for _, kv := range vals {
 		key := bitKey(kv.Key)
+		hash := kv.Hash
 		if t.root == nil {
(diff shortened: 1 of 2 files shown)


jbowens:
lgtm

```
benchmark              old ns/op     new ns/op     delta
BenchmarkInserts-4     142224708     36920148      -74.04%

benchmark              old allocs     new allocs     delta
BenchmarkInserts-4     283501         283499         -0.00%

benchmark              old bytes     new bytes     delta
BenchmarkInserts-4     17648974      15540852      -11.94%
```

👌

PR #289, opened 2016-11-24
core/asset: use singleflight for lookups

jbowens:
Use singleflight for asset lookups and bump the size of the asset
and account caches. In one 500,000 transaction benchmark, there were
~600,000 asset lookups. This reduces it to ~50,000 asset lookups.

Also, now we have less contention for DB connections and can reduce our
max connections.

### DIFF
--- a/core/asset/asset.go
+++ b/core/asset/asset.go
@@ -131,8 +131,7 @@ func (reg *Registry) findByID(ctx context.Context, id bc.AssetID) (*Asset, error
 	}
 
 	untypedAsset, err := reg.idGroup.Do(id.String(), func() (interface{}, error) {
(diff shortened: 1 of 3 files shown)


core/asset/asset.go
```diff
@@ -125,10 +129,16 @@ func (reg *Registry) findByID(ctx context.Context, id bc.AssetID) (*Asset, error
 	if ok {
 		return cached.(*Asset), nil
 	}
-	asset, err := assetQuery(ctx, reg.db, "assets.id=$1", id)
+
+	untypedAsset, err := reg.idGroup.Do(id.String(), func() (interface{}, error) {
+		asset, err := assetQuery(ctx, reg.db, "assets.id=$1", id)
```
kr: Why not `return assetQuery(...)`?

PR #290, opened 2016-11-28
bin/build-docker-de: rename build-ccde and use build-cored-release

boymanjor:
bin/build-ccde has been renamed to build-docker-de. It
now uses bin/build-cored-release to build go binaries for
the developer edition docker image. Static version tags for
the binaries and the docker image are declared. Non-release
images can be built by passing image and cored version tags
as env vars. The docker image has also been updated to pin
the alpine version and remove the use of schema.sql.


boymanjor:
@jeffomatic PTAL

bin/build-docker-de
```diff
@@ -0,0 +1,17 @@
...
+  rm -f $CHAIN/docker/de/cored
+  rm -f $CHAIN/docker/de/corectl
+  rm -f $CHAIN/docker/de/build-cored-release
+}
+trap cleanup EXIT
+
+imageVersion=${IMAGE_VERSION:-1.0.0}
+coredVersion=${CORED_VERSION:-cmd.cored-1.0.1}
+
+cp bin/build-cored-release $CHAIN/docker/de
```
jeffomatic: Is there a reason to copy this script into the target directory? It should be runnable from anywhere.

boymanjor: Oh, good catch. This is remnants from my earlier work compiling the binaries inside the image.

boymanjor:
@jeffomatic updated.

PR #291, opened 2016-11-28
bin/build-cored-release: use bash interpreter

boymanjor:
The ERR signal is specific to bash. This script
fails to run on systems without bash installed.


boymanjor:
@jeffomatic PTAL

jeffomatic:
LGTM

PR #292, opened 2016-11-28
cmd/testnet-reset: specify host location

boymanjor:
Currently testnet-reset assumes the host's location.
The location should be set explicitly so it is only
allowed to run at midnight PST. Currently the filter
cancels the job during valid execution times.

### DIFF
--- a/cmd/testnet-reset/main.go
+++ b/cmd/testnet-reset/main.go
@@ -52,7 +52,6 @@ func main() {
 		log.Fatal(err)
 	}
 	cur := time.Now().In(pst)
-	log.Println("current time:", cur)
 	max := cur.Add(time.Hour).Weekday()
 	min := cur.Add(-1 * time.Hour).Weekday()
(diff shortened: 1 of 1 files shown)


boymanjor:
PTAL

cmd/testnet-reset/main.go
```diff
@@ -47,11 +47,16 @@ func main() {
 	ctx := context.Background()
 	env.Parse()
 
-	cur := time.Now()
+	pst, err := time.LoadLocation("America/Los_Angeles")
+	if err != nil {
+		log.Fatal(err)
+	}
+	cur := time.Now().In(pst)
+	log.Println("current time:", cur)
```
jeffomatic: Is this log line intentional?

boymanjor: Yes. The largest interval allowed for the scheduler is one day. I want the log to reflect what date/time the program was run (to verify the filtering).

kr:
LGTM

PR #293, opened 2016-11-29
core: tweak submitted_txs

jbowens:
* Split the submitted_txs insert/select into two separate queries. In the vast
majority of calls, only the insert should be needed.
* Rename tx_id to tx_hash to match other tables (ex, annotated_txs,
annotated_outputs, account_utxos).
* Convert tx_hash from text to bytea.


erykwalder:
LGTM

PR #295, opened 2016-11-29
protocol/prottest: add NewIssuanceTx

jbowens:
Add a new prottest function to create valid, signed issuance
transactions. This can be useful in tests that require valid
transactions but don't need to control the contents of the
transactions.

### DIFF
--- a/protocol/prottest/tx.go
+++ b/protocol/prottest/tx.go
@@ -1,0 +1,83 @@
+package prottest
+
+import (
+	"context"
+	"crypto/rand"
+	"testing"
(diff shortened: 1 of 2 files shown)


erykwalder:
LGTM, though do we need to involve keys and such, or could we just use `OP_TRUE`?

jbowens:
> LGTM, though do we need to involve keys and such, or could we just use OP_TRUE?

I was thinking for benchmarks, it'd be nice to resemble real transactions as closely as possible.

PR #296, opened 2016-11-29
encoding/bufpool: add bytes.Buffer pool

erykwalder:
Many of the allocations in the protocol package come from creating byte
buffers. These buffers can use sync.Pool to reduce allocations.

In a recent run of benchcore, TxInput.writeTo and
OutputCommitment.writeTo in particular had many allocations from
initializing byte buffers.


jbowens:
lgtm

PR #297, opened 2016-11-29
protocol/patricia: use pointer instead of raw hash

erykwalder:
Passing the raw hash allocates a new hash on every call to insert, which
is itself a recursive call. In a recent run of benchcore, insert
accounted for ~3% of all allocations, about half of which were from
passing the raw hash.

```
benchmark              old ns/op     new ns/op     delta
BenchmarkInserts-8     29747152      26422206      -11.18%

benchmark              old allocs     new allocs     delta
BenchmarkInserts-8     283503         161759         -42.94%

benchmark              old bytes     new bytes     delta
BenchmarkInserts-8     15546252      11647064      -25.08%
```

### DIFF
--- a/protocol/patricia/patricia.go
+++ b/protocol/patricia/patricia.go
@@ -52,7 +52,7 @@ func Reconstruct(vals []Leaf) (*Tree, error) {
 		}
 
 		var err error
-		t.root, err = t.insert(t.root, key, hash)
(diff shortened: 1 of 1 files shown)


erykwalder:
A gist of the mem profile from the function before:
https://gist.github.com/erykwalder/4c6001820988c256781392e77a9789a1

jbowens:
Nice!

PR #299, opened 2016-11-29
protocol/validation: refactor benchmarks

jbowens:
Remove the blocks.json test data. It's fragile because it needs to
be updated whenever we modify serialization formats. Instead, use
prottest.NewIssuanceTx.

Add a new benchmark for the CalcMerkleRoot function.


erykwalder:
LGTM

PR #300, opened 2016-11-29
protocol/validation: optimize CalcMerkleRoot

jbowens:
Avoid unneccessary allocs in CalcMerkleRoot by using a pool of sha3
hashers and writing directly to the hasher instead of an intermediary
byte slice.

```
BenchmarkValidateBlock-4      212996        206962        -2.83%
BenchmarkCalcMerkleRoot-4     53375679      46489514      -12.90%

benchmark                     old allocs     new allocs     delta
BenchmarkValidateBlock-4      115            113            -1.74%
BenchmarkCalcMerkleRoot-4     115036         90009          -21.76%

benchmark                     old bytes     new bytes     delta
BenchmarkValidateBlock-4      8183          7064          -13.67%
BenchmarkCalcMerkleRoot-4     13860092      3212794       -76.82%
```

### DIFF
--- a/protocol/bc/transaction.go
+++ b/protocol/bc/transaction.go
@@ -209,26 +209,26 @@ func (tx *TxData) Hash() Hash {
 // WitnessHash is the combined hash of the
 // transactions hash and signature data hash.
 // It is used to compute the TxRoot of a block.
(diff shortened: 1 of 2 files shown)


erykwalder:
LGTM

PR #301, opened 2016-11-29
protocol/patricia: avoid allocs in node.Hash

jbowens:
Use pointers to child node's hashes instead of copying the hash when
computing a parent's hash.

```
benchmark                      old ns/op     new ns/op     delta
BenchmarkInserts-4             31313382      30315774      -3.19%
BenchmarkInsertsRootHash-4     46718420      39971009      -14.44%

benchmark                      old allocs     new allocs     delta
BenchmarkInserts-4             161755         161755         +0.00%
BenchmarkInsertsRootHash-4     201784         171776         -14.87%

benchmark                      old bytes     new bytes     delta
BenchmarkInserts-4             11642779      11642829      +0.00%
BenchmarkInsertsRootHash-4     12932730      11970821      -7.44%
```

### DIFF
--- a/protocol/patricia/patricia.go
+++ b/protocol/patricia/patricia.go
@@ -317,23 +317,24 @@ func (n *node) Key() []byte { return byteKey(n.key) }
 
 // Hash will return the hash for this node.
 func (n *node) Hash() bc.Hash {
-	if n.hash != nil {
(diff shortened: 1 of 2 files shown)


erykwalder:
LGTM

PR #302, opened 2016-11-30
cmd/capturestats: captures database query stats

erykwalder:
Queries a database's db stats and formats them to stdout.


jbowens:
lgtm

PR #304, opened 2016-11-30
protocol/bc: use precomputed hash in WitnessHash

erykwalder:
Rather than declaring WitnessHash on TxData, which needs to compute its
hash, declare it on Tx, which has a precomputed hash already available.

```
BenchmarkCalcMerkleRoot-8     38432421      22408414      -41.69%

benchmark                     old allocs     new allocs     delta
BenchmarkCalcMerkleRoot-8     85011          45001          -47.06%

benchmark                     old bytes     new bytes     delta
BenchmarkCalcMerkleRoot-8     2814474       1441447       -48.78%
```


jbowens:
lgtm

jbowens:
🤘

PR #306, opened 2016-11-30
core/txbuilder: add explicit retire action

jbowens:
Add an explicit retire action instead of using the `control_program`
action and specifying the `OP_FAIL` program. This provides more
flexibility as our implementation of retirement changes.

### DIFF
--- a/sdk/java/src/main/java/com/chain/api/Transaction.java
+++ b/sdk/java/src/main/java/com/chain/api/Transaction.java
@@ -927,8 +927,7 @@ public class Transaction {
        * Default constructor defines the action type as "control_program"
        */
       public Retire() {
-        this.put("type", "control_program");
-        this.put("control_program", ControlProgram.retireProgram());
+        this.put("type", "retire");
       }
 
       /**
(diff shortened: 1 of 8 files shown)


jbowens:
PTAL @jeffomatic

sdk/java/src/main/java/com/chain/api/Transaction.java
```diff
@@ -927,8 +927,7 @@ public ControlWithProgram setAmount(long amount) {
        * Default constructor defines the action type as "control_program"
        */
       public Retire() {
-        this.put("type", "control_program");
-        this.put("control_program", ControlProgram.retireProgram());
+        this.put("type", "retire");
```
jeffomatic: We should delete `ControlProgram.retireProgram()` and related dependencies.

jbowens: I think I did?

jeffomatic: Yep sorry I missed it 😞

PR #311, opened 2016-12-01
dashboard: expose version in core status page

jeffomatic:
This commit replaces the build commit display with a version display, which is much more friendly to the user.


jeffomatic:
image

dominic:
LGTM. I think we can remove `buildCommit` from the app now too.

jeffomatic:
updated, removed `buildCommit` and `buildDate`, which were unused

PR #312, opened 2016-12-01
dashboard: remove unused reducer values

jeffomatic:


dashboard/src/features/app/components/Container.jsx
```diff
@@ -90,8 +90,6 @@ export default connect(
     authOk: !state.core.requireClientToken || state.core.validToken,
     configKnown: state.core.configKnown,
     configured: state.core.configured,
-    buildCommit: state.core.buildCommit,
-    buildDate: state.core.buildDate,
```
dominic: Can we remove these from https://github.com/chain/chain/blob/main/dashboard/src/features/app/components/Container.jsx#L93-L94 as well?

PR #316, opened 2016-12-02
protocol: remove Pool interface

jbowens:
The `protocol.Chain` no longer needs to know anything about the
pending transaction pool. We already only call `AddTx` on generator
nodes. We have an alternative method `ValidateTxCached` for just
performing the validation.

This removes the `protocol.Pool` interface and the `protocol.Chain.AddTx`
method. Instead, it introduces a `txbuilder.Submitter` type that is
implemented by `mempool.MemPool` and a wrapper around the remote
generator RPC client.

For now, `mempool.MemPool` still exists as a wedge between the
`core/txbuilder` package and the `core/generator`. In the future, we
may want to remove it and modify `core/generator.generator` to implement
`txbuilder.Submitter` directly. That will be helpful when modifying the
generator to build blocks incrementally as transactions are submitted to
the pool.

Note that this will also fix the behavior where if you setup your Core to
replicate with a non-generator Core, submitted txs would end up in the
non-generator Core's pool and not be propagated to the generator.


core/transact.go
```diff
@@ -223,15 +222,15 @@ func (h *Handler) finalizeTxWait(ctx context.Context, txTemplate *txbuilder.Temp
 	return nil
 }
 
-func waitForTxInBlock(ctx context.Context, c *protocol.Chain, tx *bc.Tx, height uint64) (uint64, error) {
+func (h *Handler) waitForTxInBlock(ctx context.Context, tx *bc.Tx, height uint64) (uint64, error) {
```
kr: It seems like this part is actually not connected to the rest of the PR? Maybe it should be separate (even though it is small)?

core/txbuilder/finalize.go
```diff
@@ -50,33 +32,18 @@ func publishTx(ctx context.Context, c *protocol.Chain, msg *bc.Tx) error {
...
+	if errors.Root(err) == validation.ErrBadTx {
+		detail := errors.Detail(err)
+		err = errors.Wrap(ErrRejected, err)
+		return errors.WithDetail(err, detail)
+	} else if err != nil {
+		return errors.Wrap(err, "tx rejected")
 	}
-	return nil
+
+	err = s.Submit(ctx, tx)
```
kr: Beautiful.

core/txbuilder/submitter.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"chain/core/rpc"
+	"chain/errors"
+	"chain/protocol/bc"
+)
+
+// Submitter submits a transaction to the generator so that it may
+// be confirmed in a block.
+type Submitter interface {
```
kr: Arguably this should be defined closer to where it's used, like at the top of finalize.go?

protocol/prottest/block.go
```diff
@@ -54,7 +53,7 @@ func NewChainWithStorage(tb testing.TB, store protocol.Store, pool protocol.Pool
 // MakeBlock always makes a block;
 // if there are no transactions in the pool,
```
kr: s/the pool/txs/

core/transact.go
jbowens: I actually needed to make it a method on `Handler` so that I could access `h.Submitter`. Or I could add the `Submitter` as another parameter to the function?

kr: Oh right. I guess I was thinking we could convert this to a method first, and then this PR would only touch the FinalizeTx call in this method. But not a big deal.

kr:
LGTM (forgot to comment yesterday, oops)

PR #319, opened 2016-12-05
installer/mac: use build-cored-release to bundle cored

dominic:
Embedded cored binary is now built directly from a tag. Removes unused
`Preferences` menu item.

### DIFF
--- a/installer/mac/ChainCore/ChainCore.swift
+++ b/installer/mac/ChainCore/ChainCore.swift
@@ -130,6 +130,8 @@ class ChainCore: NSObject {
             "DATABASE_URL": databaseURL,
             "LISTEN":       ":\(port)",
             "LOGFILE":      self.logURL.path,
+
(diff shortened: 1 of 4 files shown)


dominic:
PTAL @jeffomatic

installer/mac/ChainCore/ChainCore.swift
```diff
@@ -129,7 +129,8 @@ class ChainCore: NSObject {
         task.environment = [
             "DATABASE_URL": databaseURL,
             "LISTEN":       ":\(port)",
-            "LOGFILE":      self.logURL.path
+            "LOGFILE":      self.logURL.path,
+            "USER":         NSUserName(),
```
jeffomatic: Can we put a comment here to indicate why it's necessary? I think its role is a little less obvious than the other envvars.

dominic:
@jeffomatic updated

PR #324, opened 2016-12-07
protocol/validation: benchmark CheckTxWellFormed

kr:
On my machine:

```
BenchmarkValidateTx-8   	   10000	    141048 ns/op
```


kr:
PTAL

jbowens:
lgtm

PR #325, opened 2016-12-07
docs/protocol: specify maximum VM stack item length

danrobinson:
It doesn't make sense to support VM stack items larger than 65535 bytes, particularly since such a long stack item would necessarily cause the runlimit to fail (it would cost over 500,000 runlimit just to be pushed onto the stack). Setting this limit would simplify the design of certain additional VM opcodes, particularly serialization (since it would allow the length of any stack item to be represented in exactly two bytes). It might also slightly simplify the VM implementation.

### DIFF
--- a/docs/protocol/specifications/vm1.md
+++ b/docs/protocol/specifications/vm1.md
@@ -167,6 +167,8 @@ All items on the data and alt stacks are binary strings. Some instructions accep
 
 In the stack diagrams accompanying the definition of each operation, `x` and `y` denote [numbers](#vm-number), `m` and `n` denote non-negative [numbers](#vm-number), and `p` and `q` denote [booleans](#vm-boolean). If coercion fails (or if stack items designated as `m` or `n` coerce to negative numbers), the operation fails.
 
+The maximum length of a string on the VM stack is 65535 bytes.
+
 ### VM String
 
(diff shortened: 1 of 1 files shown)


danrobinson:
@oleganza @kr @bobg WDYT?

oleganza:
"640 Kb should be enough for everyone" — John Lennon.

We should strive to have flexible precommitments to any resource consumption for two opposite reasons:

1. We want to allocate as little memory and CPU as necessary to verify smth.
2. We want not to lock ourselves into arbitrary limits that in some context may be too limiting. E.g. Chain VM not necessarily will be only used for on-chain programs. And even some on-chain programs may want to operate on large data, especially if it's a slow and extra super duper secure internal ledger inside a single org.

danrobinson:
Whether we should have flexible precommitments to the virtual machine run cost for a program is an independent question from what the maximum stack item size should be.

I think this is premature allocation. If you want to use VM1 completely outside the context of a blockchain, and you might want to handle 65-kilobyte stack items, then IMO you can fork it.

That said, this is a pretty minor issue that I mostly proposed so that we could avoid introducing varints to the VM in #326. But I realized that the blockchain data structures themselves are serialized with varints, so it may actually improve consistency. So I'll close this and edit #326 to use varints.

PR #326, opened 2016-12-07
docs/protocol: add serialization/deserialization to VM

danrobinson:
The VM currently does not have low-level support for manipulation of lists, which makes it very difficult to fully support many desirable high-level language features in Ivy (including not only lists, but also transaction introspection, structs, signature programs, merklized paths, and generic `checkMultiSig`).

The proposed changes define a (bytelength-delimited) serialization format for lists in the VM, and include the four opcodes that I think would be necessary to support these features: 

1. `PACK`: converts a list of stack items into a single list.
2. `UNPACK`: takes a list and pushes each of its items onto the stack.
3. `ELEMENT`: gets the `i`th element of a list.
4. `SLICE`: gets a portion of a list by index; this one is less critical but would be a nice feature. 

For safety, `UNPACK` takes the length of the list as an argument, and `ELEMENT` and `SLICE` check the entire list for validity (not just the portion needed to pick out the element or elements).

I think it's worth adding this to VM1 rather than waiting for VM2 (which will likely support a more efficient implementation of lists), partly for the obvious reason that it would allow us to use these features sooner, but also because it would help us get some experience that would allow us to better design those parts of VM2 (as well as Ivy).

### DIFF
--- a/docs/protocol/specifications/vm1.md
+++ b/docs/protocol/specifications/vm1.md
@@ -165,13 +165,13 @@ Most instructions use only the data stack by removing some items and then placin
 
 All items on the data and alt stacks are binary strings. Some instructions accept or return items of other types. When values of those types are pushed to or popped from the data stack, they are coerced to and from strings in accordance with the rules specified below.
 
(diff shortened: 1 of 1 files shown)


danrobinson:
@bobg @oleganza @kr WDYT?

PR #327, opened 2016-12-08
cmd/md2html: fix HTML escaping in code snippets

jeffomatic:
This commit adds proper escaping of HTML entities in code snippets, a
bug that was introduced in when langage switching was added. This caused
issues such as incorrect rendering of Java generic expressions
("List<Asset>").

This bug was due to the following:

1. Adding switching behavior required adding logical divs that enclosed
the code snippets.
2. Enclosing divs cause blackfriday, the Markdown parser, to ignore
fenced code blocks inside those divs.
3. We switched to manually inserting <pre><code> wrappers, but neglected
to add HTML entity escaping.

### DIFF
--- a/cmd/md2html/main.go
+++ b/cmd/md2html/main.go
@@ -285,7 +285,7 @@ func writeCode(w io.Writer, path, snippet string) {
 	if err != nil {
 		code = err.Error()
 	} else {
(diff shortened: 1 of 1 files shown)


jeffomatic:
PTAL

PR #331, opened 2016-12-09
types: New top-level package

bobg:
This moves the types `Hash`, `AssetID`, and `AssetAmount` from the `protocol/bc` package to a new package (whose name is admittedly weak, suggestions welcome) in order to prevent a circular dependency when some other upcoming changes land.

It also requires callers to pass both the asset version and the vm version to `ComputeAssetID`, also in anticipation of upcoming changes.


bobg:
PTAL

bobg:
@jbowens had a better idea, closing this PR.

PR #332, opened 2016-12-10
core: fix Handler.init when unconfigured

jbowens:
If the Core is unconfigured, Submitter will be nil. Instead, wrap the
Submit call in a closure so that if the Core is unconfigured, we won't
try to access the Submitter's Submit method until we're processing a
request (and we shouldn't then either because requests will
be caught by the needConfig middleware).

The bug was introduced in #316.


jbowens:
PTAL

bobg:
LGTM

PR #333, opened 2016-12-10
core: use batchRecover in create-control-program

jbowens:
Other batch endpoints besides create-control-program use the
`batchRecover` function to format error responses and to recover
from panics. The `create-control-program` handler slipped through
the cracks.


bobg:
LGTM

PR #334, opened 2016-12-12
protocol/vm: error handling & spec-tracking updates

bobg:
Misc. changes:
- Remember outermost program during VM execution
- Streamline reporting and handling of VM and validation errors
- Track spec revision re: handling expansion opcodes
- Add some test cases.

### DIFF
--- a/protocol/vm/vm.go
+++ b/protocol/vm/vm.go
@@ -1,8 +1,10 @@
 package vm
 
 import (
+	"encoding/hex"
 	"fmt"
 	"io"
+	"strings"
 
 	// TODO(bobg): very little of this package depends on bc, consider trying to remove the dependency
 	"chain/protocol/bc"
@@ -11,10 +13,12 @@ import (
 const initialRunLimit = 10000
 
 type virtualMachine struct {
-	program      []byte
-	pc, nextPC   uint32
-	runLimit     int64
-	deferredCost int64
+	mainprog, curprog []byte
+	pc, nextPC        uint32
+	runLimit          int64
+	deferredCost      int64
+
+	expansionReserved bool
 
 	// Stores the data parsed out of an opcode. Used as input to
 	// data-pushing opcodes.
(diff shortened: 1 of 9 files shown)


bobg:
PTAL

protocol/vm/vm.go
```diff
@@ -258,3 +263,34 @@ func stackCost(stack [][]byte) int64 {
 	}
 	return result
 }
+
+type VMErr struct {
```
kr: Do we want to export this type? If so, it could use some documentation describing how the caller is meant to use it (e.g. https://godoc.org/os#PathError). If we don't think anyone is likely to do a `err.(vm.VMErr)` type assertion, then this might be better left unexported.

Btw, the naming convention for error types is `FooError` or `fooError`. In this case, since the package is already named `vm` it should probably be called just `Error`.

kr:
:eyes:

protocol/vm/vm.go
```diff
@@ -55,39 +59,43 @@ func verifyTxInput(tx *bc.Tx, inputIndex int) (bool, error) {
...
+			if err != nil {
+				return false, err
+			}
 		}
+		ok, err := vm.run()
+		return ok, wrapErr(err, &vm, args)
 	}
 
-	return vm.run()
+	switch inp := txinput.TypedInput.(type) {
```
kr: I'm still a little surprised we don't yet have a method `TxInput.Program`. Places like this (which is most of them) should just call that method. Stay tuned for a patch.

protocol/vm/vm.go
```diff
@@ -11,10 +13,12 @@ import (
 const initialRunLimit = 10000
 
 type virtualMachine struct {
-	program      []byte
-	pc, nextPC   uint32
-	runLimit     int64
-	deferredCost int64
+	mainprog, curprog []byte
```
kr: Do we need to rename the existing field?

I'm not sure it's that much clearer to say curprog instead of program. But for the other one (which is just for the opcode right?) mainprog or outerProg or whatever definitely helps, and what might help even more is an end-of-line comment on these two that says why there are two of them. Roughly like

```
program  []byte // actual program currently executing
mainprog []byte // outermost program, returned by PROGRAM opcode
```

PR #336, opened 2016-12-12
docs/internal: add type to Swagger tx builder action definitions

dominic:


dominic:
PTAL @devongundry

devongundry:
LGTM

PR #337, opened 2016-12-13
protocol/bc: add input program accessor

ameets:
In general, callers do not care which type of input returned the program.
Make a generic program accessor to return the correct program based on type.

### DIFF
--- a/protocol/bc/txinput.go
+++ b/protocol/bc/txinput.go
@@ -121,20 +121,6 @@ func (t *TxInput) VMVersion() uint64 {
 	return t.TypedInput.vmVersion()
 }
 
-func (t *TxInput) ControlProgram() []byte {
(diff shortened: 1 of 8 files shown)


erykwalder:
LGTM

protocol/bc/txinput.go
```diff
@@ -115,6 +117,10 @@ func (t *TxInput) Amount() uint64 {
 	return si.Amount
 }
 
+func (t *TxInput) VMVersion() uint64 {
+	return t.TypedInput.vmVersion()
+}
+
 func (t *TxInput) ControlProgram() []byte {
```
kr: I just noticed TxInput.ControlProgram and TxInput.IssuanceProgram are no longer called anywhere. We could potentially remove them in this patch.

kr:
We decided to put this idea on hold while we figure out some other protocol changes.

PR #341, opened 2016-12-15
all: use testutil.FatalErr

jbowens:
Replace uses of:
```
t.Log(errors.Stack(err))
t.Fatal(err)
```

with:
```
testutil.FatalErr(t, err)
```


kr:
nice

PR #344, opened 2016-12-16
protocol/vm, protocol/validation: improve error handling and tests

bobg:
Also keep track of the outermost program during CHECKPREDICATE calls, so it can be returned by OP_PROGRAM.

### DIFF
--- a/protocol/validation/tx.go
+++ b/protocol/validation/tx.go
@@ -2,9 +2,7 @@ package validation
 
 import (
 	"bytes"
-	"encoding/hex"
 	"math"
-	"strings"
 
 	"chain/errors"
 	"chain/math/checked"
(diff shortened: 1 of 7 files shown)


bobg:
PTAL

bobg:
PTAL

kr:
:eyes:

protocol/validation/tx.go
```diff
@@ -22,6 +20,44 @@ var (
 	ErrFalseVMResult = errors.New("false VM result")
 )
 
+var (
+	// "suberrors" for ErrBadTx
+	errTxVersion              = errors.New("unknown transaction version")
```
kr: I'm not totally on board with defining these errors up here instead of where they're used. It would be clearly useful if each error were used multiple times, or if callers were testing the error value to handle them programmatically. But in this case it seems like an unnecessary level of indirection. What do you think?

protocol/vm/vm.go
```diff
@@ -258,3 +265,34 @@ func stackCost(stack [][]byte) int64 {
...
 	return result
 }
+
+type Error struct {
+	Err  error
+	Prog []byte
+	Args [][]byte
+}
+
+func (vmerr Error) Error() string {
```
kr: nit: `e` would be a more conventional name for the receiver

PR #345, opened 2016-12-16
core/txbuilder: improve signature witness signing perf

tessr:
Working on improving TransferWithBlocks, and this is a tiny fix. 

Before
```
BenchmarkTransferWithBlocks-4   	     200	  31896144 ns/op	   61624 B/op	     919 allocs/op
```

After
```
BenchmarkTransferWithBlocks-4   	     300	  25861102 ns/op	   67099 B/op	     913 allocs/op
```

### DIFF
--- a/core/txbuilder/witness.go
+++ b/core/txbuilder/witness.go
@@ -130,9 +130,9 @@ func (sw *SignatureWitness) Sign(ctx context.Context, tpl *Template, index uint3
 		if !contains(xpubs, keyID.XPub) {
 			continue
 		}
-		var path [][]byte
-		for _, p := range keyID.DerivationPath {
-			path = append(path, p)
+		path := make([]([]byte), len(keyID.DerivationPath))
+		for i, p := range keyID.DerivationPath {
+			path[i] = p
 		}
 		sigBytes, err := signFn(ctx, keyID.XPub, path, h)
 		if err != nil {


core/txbuilder/witness.go
```diff
@@ -128,9 +128,9 @@ func (sw *SignatureWitness) Sign(ctx context.Context, tpl *Template, index int,
 		if !contains(xpubs, keyID.XPub) {
 			continue
 		}
-		var path [][]byte
-		for _, p := range keyID.DerivationPath {
-			path = append(path, p)
+		path := make([]([]byte), len(keyID.DerivationPath))
```
bobg: Why not `make([][]byte, len(keyID.DerivationPath))`?

It's just possible that `make([][]byte, 0, len(keyID.DerivationPath))` is a hair better in performance, because then the Go runtime doesn't have to initialize that space to all `nil`s. (You would then revert the subsequent lines to be the same as before the diff.)

kr:
For easy comparison, you can use [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp) to produce output like this:

```
benchmark                         old ns/op     new ns/op     delta
BenchmarkTransferWithBlocks-4     31896144      25861102      -18.92%

benchmark                         old bytes     new bytes     delta
BenchmarkTransferWithBlocks-4     61624         67099         +8.88%
```

kr:
LGTM (sorry I neglected to comment before)

nit: imperative subject

PR #347, opened 2016-12-19
core/txbuilder: remove maxTime Build parameter

jbowens:
Remove the maxTime parameter from the Build method on the Action
interface. Instead, provide a `MaxTime` method on the `TemplateBuilder`.

Also, changes `RestrictMinTimeMS` to `RestrictMinTime` so that both
min and max times operate on time structs.


jbowens:
PTAL

Issue #346, opened 2016-12-16
Access token list API does not return pagination information


dominic:
In a core with > 100 tokens, the `next` key in the response from `/list-access-tokens` for the first page of results is:

```
next: {page_size: 0, timeout: 0, after: "", type: "client"}
```

`after: ""` cannot be used to build a query for the next page.

jeffomatic:
Resolve in #356

PR #350, opened 2017-01-03
core: respect page_size parameter in API endpoints that return pages

dominic:


dominic:
PTAL

jeffomatic:
1. Is this related to #346?
2. Is there any other specific issue that is unblocked by this change? It's a good change, but it's one that should probably be pushed out to all of our paginated endpoints.

dominic:
This is unrelated to #346, where - in that case - the `page_size` param is handled correctly for access tokens, but the `after` object isn't generated correctly to request a new page. I discovered this issue while working on the Node SDK, but didn't want to bury the fix in the SDK pull request.

dominic:
Digging deeper, I am seeing that many query endpoints also fail to respect this field, i.e. https://github.com/chain/chain/blob/main/core/query.go#L219

Should we extend this PR to include them all?

jeffomatic:
> Should we extend this PR to include them all?

That sounds good--as we discussed IRL, the get-balances endpoint will be an exception. Please remember to also update the API spec YAML file as well...bug me if you need help with that.

dominic:
@jeffomatic PTAL again

dominic:
@jeffomatic updated

PR #351, opened 2017-01-03
sdk/node: new SDK

dominic:


dominic:
Still WIP, and sorting out this CLA issue.

jeffomatic:
In the example files, I notice that we use both `(foo) =>` and `function (foo)` style for lambdas. I don't think it makes a difference, so it'd be nice to be consistent. I think I prefer `(foo) =>`.

sdk/node/test/callbacks.js
```diff
@@ -0,0 +1,420 @@
...
+        assert.equal(batchResponse.successes[0], null)
+        assert.deepEqual([batchResponse.errors[1], batchResponse.errors[2]], [null, null])
+        next()
+      }),
+
+
+      // Basic issuance
+
+      (next) => async.waterfall([
+        cb => client.transactions.build(function(builder){
```
jeffomatic: Can we do `build(builder => ...`?

dominic: The promises test will have the same problem.

sdk/node/src/connection.js
```diff
@@ -5,6 +5,53 @@ const { fetch } = require('fetch-ponyfill')()
 const errors = require('./errors')
 const btoa = require('btoa')
 
+const blacklistAttributes = [
+  'asset_tags',
+  'asset_definition',
+  'account_tags',
+  'reference_data',
```
jeffomatic: Should probably add:

- `tags`: used in assets.create, accounts.create
- `next`, `after`: pagination cursor in response and request, respectively

We could probably refer to the Swagger spec for any more instances of arbitrary objects.

sdk/node/src/connection.js
```diff
@@ -34,6 +81,8 @@ class Connection {
       body = {}
     }
 
+    const snakeBody = snakeize(body) // Ssssssssssss
```
jeffomatic: Love this comment, but I would like to expand on it to explain why this is happening.

sdk/node/src/connection.js
```diff
@@ -111,6 +160,8 @@ class Connection {
             requestId: requestId
           }
         )
+      }).then((body) => {
+        return camelize(body)
```
jeffomatic: Comment here to indicate why this is happening.

sdk/node/testSrc/callbacks.js
```diff
@@ -11,7 +11,7 @@ const balanceByAssetAlias = (cb) => {
     let res = {}
 
     balances.items.forEach((item) => {
-      res[item.sum_by['asset_alias']] = item.amount
+      res[item.sumBy['assetAlias']] = item.amount
```
jeffomatic: I realize this is old code, but it's probably ok to do `item.sumBy.assetAlias` here.

sdk/node/testSrc/callbacks.js
```diff
@@ -413,7 +413,7 @@ describe('Callback style', () => {
         next()
       }),
 
-      cb => done()
+      () => done()
```
jeffomatic: you can just pass `done` here without the lambda

sdk/node/testSrc/promises.js
```diff
@@ -13,7 +13,7 @@ const balanceByAssetAlias = (balances) => {
   return Promise.resolve(balances)
   .then((balance) => {
     balance.items.forEach((item) => {
-      res[item.sum_by['asset_alias']] = item.amount
+      res[item.sumBy['assetAlias']] = item.amount
```
jeffomatic: `item.sumBy.assetAlias`

dominic:
PTAL

PR #352, opened 2017-01-04
core/generator: collect pool txs on Generator

jbowens:
Collect pending transactions on the generator struct instead of a
separate type. Since Generator now implements the `txbuilder.Submitter`,
we can incrementally build blocks as new transactions are submitted.
The protocol/mempool type still exists for tests.

Future work:
* use generator.Generator in core/ tests where possible
* move protocol/mempool type into protocol/prottest
* incrementally build blocks as transactions are submitted

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -45,7 +45,6 @@ import (
 	"chain/net/http/limit"
 	"chain/protocol"
 	"chain/protocol/bc"
-	"chain/protocol/mempool"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


kr:
:eyes:

kr:
Sweet. Is there a chance we might get rid of mempool entirely some day? Like if every place uses a Generator?

jbowens:
> Is there a chance we might get rid of mempool entirely some day? Like if every place uses a Generator?

Yeah, I think that's right

PR #353, opened 2017-01-05
core/config: rename block_xpub column to block_pub

tessr:
This fulfills one of my old todos and prepares for grpc (`chainjson.HexBytes` is not compatible with protobufs).

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -16,6 +16,7 @@ import (
 	"chain/crypto/ed25519"
 	"chain/database/pg"
 	"chain/database/sql"
+	chainjson "chain/encoding/json"
 	"chain/errors"
 	"chain/log"
 	"chain/protocol"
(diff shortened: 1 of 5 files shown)


core/config/config.go
```diff
@@ -44,23 +43,23 @@ type Config struct {
...
 	Quorum               int
 	MaxIssuanceWindow    time.Duration
 }
 
 type BlockSigner struct {
-	AccessToken string             `json:"access_token"`
-	Pubkey      chainjson.HexBytes `json:"pubkey"`
-	URL         string             `json:"url"`
+	AccessToken string `json:"access_token"`
+	Pubkey      []byte `json:"pubkey"`
```
erykwalder: Until we have grpc, we should probably leave these as hex bytes. I think otherwise requests will have to use base64.

tessr: Ah, I wonder if that's what's making the dashboard tests complain.

tessr: (Hmm, turns out, no, they were genuine lint errors. Regardless, I'll turn the pubkeys back into hex bytes.)

tessr:
This no longer changes the type. PTAL

erykwalder:
LGTM

Issue #169, opened 2016-11-09
Failing to add key to SDK signer produces misleading error message


jeffomatic:
Reproducible on 3579bf1992f423b9f9f246f9f15c8497ee3b3d5f

If you attempt to sign for a spending transaction without first adding the relevant key to the SDK signer object, the signer will not add a signature to the transaction. During the submit step, you are not told that signatures are missing; instead, you will receive a CH736 "Transaction not final" error. While this error is technically correct, it is not the right one to deliver to the user. The error should be that the signature is **missing**, and possibly hint that an xpub needs to be added to the client-side signing object.

The following Ruby script can recreate the problem:

```ruby
$:.unshift ENV['CHAIN'] + '/sdk/ruby/lib'

require 'chain'

c = Chain::Client.new
s = Chain::HSMSigner.new

asset_key = c.mock_hsm.keys.create
s.add_key(asset_key, c.mock_hsm.signer_conn)

account_key = c.mock_hsm.keys.create
# Don't add the key to the signer
#s.add_key(account_key, c.mock_hsm.signer_conn)

asset = c.assets.create(root_xpubs: [asset_key.xpub], quorum: 1)
acc = c.accounts.create(root_xpubs: [account_key.xpub], quorum: 1)

c.transactions.submit(s.sign(c.transactions.build { |b|
  b.issue asset_id: asset.id, amount: 100
  b.control_with_account account_id: acc.id, asset_id: asset.id, amount: 100
}))

c.transactions.submit(s.sign(c.transactions.build { |b|
  b.spend_from_account account_id: acc.id, asset_id: asset.id, amount: 1
  b.retire asset_id: asset.id, amount: 1
}))
```

The immediate fix for this should be to provide a better error message.

Some suggestions for longer-term fixes:

1. The SDK signing interface's `sign` method could return the number of signatures added. Based on the semantics of the multi-HSM signing interface, it is not an error to add zero signatures, but the user could check.
2. Move away from the multi-HSM signing interface, which is solving a future problem (having multiple HSMs in the same application) by making the 99% common case harder and more bug-prone. Adding xpubs to the signer provides for HSM routing that does not actually occur in practice.  Up front, we can provide sign/sign-batch methods directly from the Mock HSM, and remove the artificial constraint that the client must add xpubs to the signer API. This would remove the ["Load key" step](https://chain.com/docs/core/build-applications/keys#load-key) in our setup workflow, which is a significant efficiency.

tessr:
[...]

PR #359, opened 2017-01-06
core: more useful error message when missing signature

ameets:
This partially addresses #169


jeffomatic:
I think we have to perform an actual check for signatures. The user-facing error message for `txbuilder.ErrNoTxSighashCommitment` is correct--it is indicating that among the inputs in the transaction, there is no signature that commits to the entire transaction. This will trigger if there are signatures present, but only signatures that commit to individual outputs.

The problem seems to be that transactions with no signatures at all will fail the check for a txhash commitment by default--if you have zero signatures, you can't have a signature that commits to the txhash. I think we need another piece of logic that verifies the presence of signatures (or some other evidence that the user signed at all) before we move onto the check for the txhash commitment.

jeffomatic:
It seems like this may be superseded by #373. If so, we can probably go ahead and close this PR?

ameets:
Closing - superseded by [https://github.com/chain/chain/pull/373](#373)

PR #360, opened 2017-01-06
core: store info vars on config

tessr:
`/info` no longer returns JSON for `version`, `buildDate` and `buildCommit` (another pre-req for GRPC).


erykwalder:
LGTM

PR #362, opened 2017-01-06
core: use []byte instead of chain/encoding/json.Map

tessr:
Another GRPC requirement.


tessr:
Haha, jk, this isn't a prereq, this is a... co-req? Anyways, it will have to go in alongside the GRPC patch.

PR #364, opened 2017-01-09
core: remove pointers for gRPC

tessr:


core/core.go
```diff
@@ -70,15 +70,15 @@ func (h *Handler) info(ctx context.Context) (map[string]interface{}, error) {
 
 func (h *Handler) leaderInfo(ctx context.Context) (map[string]interface{}, error) {
 	var (
-		generatorHeight  *uint64
-		generatorFetched *time.Time
+		generatorHeight  uint64
```
kr: Are the SDK or dashboard currently looking for json `null` in this field? Do we need to update clients to look for `0` instead?

tessr: Is it even possible to be at height 0?

kr: That field would appear as 0 under the same circumstances as it would show up as null before this patch.

Based on the comments below, it happens when we're having trouble connecting to the generator. Looks like this patch should also update the last sentence of the comment on line 91:

```
// fetchTime might be the zero time if we're having trouble connecting
// to the remote generator. Only set the height & time if we have it.
// The dashboard will handle nulls correctly.
```

tessr: @dominic has given me his blessing from the dashboard perspective

dominic: If `0` truly means "we don't know", I'd update these eventually:

- https://github.com/chain/chain/blob/stop-pointing/dashboard/src/features/core/reducers.js#L39
- https://github.com/chain/chain/blob/stop-pointing/dashboard/src/features/core/reducers.js#L68
- https://github.com/chain/chain/blob/stop-pointing/dashboard/src/features/core/reducers.js#L131

However, I don't believe Dashboard will break if it gets zero back instead of null. The worst consequence is that replication lag will be "green" instead of "red" for `0`.

tessr: I'll update those in this PR as well, then.

tessr: All fixed 💃

jbowens:
lgtm

PR #365, opened 2017-01-09
protocol/mempool: delete package

jbowens:
Delete protocol/mempool and replace all of its uses with
core/generator.

For now, this relies on a new generator.Generator method PendingTxs that
returns the current pool. If we refactor the TestRecovery we might be
able to remove the PendingTxs method and add a MakeBlock method to
Generator to replace prottest.MakeBlock.


bobg:
:eyes:

core/generator/generator.go
```diff
@@ -59,6 +59,17 @@ func New(
 	}
 }
 
+// PendingTxs returns all of the pendings txs that will be
+// included in the generator's next block.
+func (g *Generator) PendingTxs() []*bc.Tx {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	txs := make([]*bc.Tx, len(g.pool))
```
bobg: `return g.pool[:]` ?

bobg:
LGTM

kr:
Ah, sorry, I didn't hit the "Comment" button yesterday.

The only thing that makes me uneasy about this is that it means a bunch of our tests have to depend on this other high-level thing (Generator) even if they're in an unrelated package. We already have the problem where most of our tests are really integration tests and it's very hard for us to write good unit tests, so that's not a new thing. But this would make it a little more severe.

Still, I think this is a good change. Maybe the real answer to my concern is to rewrite some of those other tests so they don't even need a Submitter at all. (So depend on even less high-level import-the-world sorts of things.) Hmm.

core/generator/generator.go
jbowens: Since `g.pool` can still be modified (and even sorted in topo order), I think this needs to be an explicit copy of the slice.

PR #367, opened 2017-01-09
core/config: make MaxIssuanceWindow a json.Duration

tessr:
When MaxIssuanceWindow was a time.Duration it was difficult to configure through an API call; now it accepts values like "12h." 

This will change when we switch to gRPC, but for now, this is nice.


tessr:
PTAL @erykwalder

erykwalder:
LGTM

PR #368, opened 2017-01-10
core/txbuilder: build returns tx data along with template

tessr:
More prep for gRPC--we used to pass the txdata struct around inside of the txtemplate because we could rely on the marshal function of txdata.


tessr:
PTAL @erykwalder

erykwalder:
LGTM

PR #369, opened 2017-01-10
core/txbuilder: turn ints into uint32s on signing path

tessr:
gRPC requires that we specify the int type; on the signing path we chose uint32s.

### DIFF
--- a/core/txbuilder/types.go
+++ b/core/txbuilder/types.go
@@ -33,7 +33,7 @@ func (t *Template) Hash(idx uint32) bc.Hash {
 	if t.sigHasher == nil {
 		t.sigHasher = bc.NewSigHasher(t.Transaction)
 	}
-	return t.sigHasher.Hash(int(idx))
+	return t.sigHasher.Hash(idx)
 }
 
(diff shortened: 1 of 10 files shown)


core/txbuilder/types.go
```diff
@@ -29,24 +29,24 @@ type Template struct {
 	sigHasher *bc.SigHasher
 }
 
-func (t *Template) Hash(idx int) bc.Hash {
+func (t *Template) Hash(idx uint32) bc.Hash {
 	if t.sigHasher == nil {
 		t.sigHasher = bc.NewSigHasher(t.Transaction)
 	}
-	return t.sigHasher.Hash(idx)
+	return t.sigHasher.Hash(int(idx))
```
bobg: This is not a safe conversion. Maybe update `SigHasher.Hash` to take a `uint32`?

tessr: Eeek, thanks, I'll look into it.

tessr:
This is all fixed up 💃

tessr:
I hope

bobg:
LGTM

PR #370, opened 2017-01-10
core: use fixtures for asset/account creation

erykwalder:
Consolidates account and asset creation to the coretest package. This
helps reduce future updates. Some calls to these functions in tests need
behavior not defined by coretest, and have been left alone.


jbowens:
Nice

PR #371, opened 2017-01-10
all: revise Ruby requirement from 2 to 2.1

### LINKED ISSUE #358: Ruby SDK requires Ruby 2.1
There's [a line](https://github.com/chain/chain/blob/main/sdk/ruby/lib/chain/access_token.rb#L31) in the ruby sdk that uses [required keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments#required-keyword-arguments) which appear in ruby 2.1. 

We should either specify ruby 2.1 or remove this syntax.

jeffomatic:
The Ruby SDK makes use of language features only available in
Ruby 2.1, so the documentation should reflect this.

Resolves #358.

### DIFF
--- a/docs/core/get-started/sdk.md
+++ b/docs/core/get-started/sdk.md
@@ -13,7 +13,7 @@ The Java SDK is available as a self-contained JAR file, with dependencies compil
 
 ## Ruby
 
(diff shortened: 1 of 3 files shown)


jeffomatic:
PTAL

dominic:
@jeffomatic We should add http://guides.rubygems.org/specification-reference/#required_ruby_version= to the gemspec to prevent installation issues as well.

jeffomatic:
@dominic Nice catch, updated.

dominic:
Approved 👍

PR #372, opened 2017-01-11
core/txdb: refactor TestInsertBlock

jbowens:
Refactor TestInsertBlock to use GetBlock instead of using its own
helper function to retrieve the block by its hash.


bobg:
LGTM

PR #373, opened 2017-01-11
core/txbuilder: better error messages for missing sigs

ameets:


tessr:
Two notes on PR style:

1. If a PR isn't ready for review, it's not a bad idea to comment "WIP" (work in progress) somewhere on the PR. 
2. When PRs get merged, the commit on main takes the name of the pull request. For history's sake, we like to have commits begin with the name of the package driving the commit. (In this case, it might be something like "core/txbuilder: better error messages when txns are missing sigs") 

🎈

ameets:
In `checkTxSighashCommitment` in finalize.go, it appears that the validations around `program` and `args` applies only to spendInputs based on the final error checking outside the for loop:

`if !allIssuances {
        return ErrNoTxSighashCommitment
  }

return nil`

If that's the intended behavior, would it make sense to skip validation checks for any issuanceInput in that fn?

i.e.
`case *bc.IssuanceInput:
			continue`

bobg:
I don't think so, because the function can succeed (return `nil`) within the `for` loop even when the current input is an issuance: https://github.com/chain/chain/blob/2c1b576cbd91fbe27dbc9c6469afd5a032f308a5/core/txbuilder/finalize.go#L96

ameets:
Thanks for the offline explanation @bobg - I didn't quite get that issuanceInputs would/could include a program (not just spendInputs). Makes sense now.

core/errors.go
```diff
@@ -128,6 +128,8 @@ var (
 		txbuilder.ErrBadWitnessComponent:   errorInfo{400, "CH733", "Invalid witness component"},
 		txbuilder.ErrRejected:              errorInfo{400, "CH735", "Transaction rejected"},
 		txbuilder.ErrNoTxSighashCommitment: errorInfo{400, "CH736", "Transaction is not final, additional actions still allowed"},
+		txbuilder.ErrTxSignatureFailure:    errorInfo{400, "CH737", "Transaction signature missing, possibly missing signature key"},
+		txbuilder.ErrNoTxSighashAttempt:    errorInfo{400, "CH738", "Transaction was not attempted"},
```
erykwalder: Somehow indicating that signing was what was not attempted in this error would be good.

core/txbuilder/finalize.go
```diff
@@ -57,10 +57,13 @@ func FinalizeTx(ctx context.Context, c *protocol.Chain, s Submitter, tx *bc.Tx)
 // least one input to commit to the complete transaction (what you get
 // when you build a transaction with allow_additional_actions=false).
 var ErrNoTxSighashCommitment = errors.New("no commitment to tx sighash")
+var ErrNoTxSighashAttempt = errors.New("no tx sighash attempted")
```
erykwalder: typically with a lot of back to back var declarations we group them like so:
```
var (
   ErrNoTxSighashCommitment = ...
   ErrNoTxSighashAttempt = ...
   ErrTxSignatureFailure = ...
)
```

core/txbuilder/finalize.go
```diff
@@ -57,10 +57,13 @@ func FinalizeTx(ctx context.Context, c *protocol.Chain, s Submitter, tx *bc.Tx)
...
+var ErrTxSignatureFailure = errors.New("tx signature was attempted but failed")
 
 func checkTxSighashCommitment(tx *bc.Tx) error {
```
erykwalder: It may be worth separating the new checks into a function that is called separately, before `checkTxSighashCommitment`. The reasoning being that the two added error checks apply to _all_ inputs, whereas `checkTxSighashCommitment` stops at the first valid input. Perhaps the added signature checks should happen for all inputs.

jeffomatic:
Just checking in on the status of this PR...is anyone hoping to return to this anytime soon?

ameets:
@jeffomatic Yup - target is Feb 8.

tessr:
I'm picking this up.

tessr:
Alright, I've responded to feedback and tweaked some things. The original script from #169 now returns:

```
 Code: CH737 
Message: Transaction signature missing, client may be missing signature key 
Request-ID: ed250f5175a39faf197d
```

PTAL

jbowens:
👀

jbowens:
lgtm

PR #374, opened 2017-01-11
core/signers: change xpubs column type to bytea

erykwalder:
Rather than dealing with string types and converting back and forth
between hex, the raw bytes can be used directly.

### DIFF
--- a/core/signers/signers.go
+++ b/core/signers/signers.go
@@ -2,6 +2,7 @@
 package signers
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"encoding/binary"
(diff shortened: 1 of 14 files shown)


core/signers/signers.go
```diff
@@ -75,27 +76,27 @@ func Path(s *Signer, ks keySpace, itemIndexes ...uint64) [][]byte {
...
 
 // Create creates and stores a Signer in the database
-func Create(ctx context.Context, db pg.DB, typ string, xpubs []string, quorum int, clientToken string) (*Signer, error) {
+func Create(ctx context.Context, db pg.DB, typ string, xpubs []chainkd.XPub, quorum int, clientToken string) (*Signer, error) {
 	if len(xpubs) == 0 {
 		return nil, errors.Wrap(ErrNoXPubs)
 	}
 
-	sort.Strings(xpubs) // this transforms the input slice
+	sort.Sort(sortKeys(xpubs)) // this transforms the input slice
```
jbowens: Is it ok if we're sorting by a different criteria here? (byte array vs hex representation)

jbowens:
lgtm

core/signers/signers.go
erykwalder: unless you can think of examples otherwise, I think since numbers are "less than" letters, it should be functionally equivalent. By byte is probably the more accurate sort that we want anyhow.

PR #375, opened 2017-01-11
core/txbuilder: replace string keys with XPubs

erykwalder:
Rather than working with strings, these functions can work with the keys
themselves, and leave conversion to the request layer.


bobg:
LGTM
🙌

PR #349, opened 2016-12-20
protocol/bc: reorganize and refactor

bobg:
Reorganize and refactor `TxData`, `TxInput`, `TxOutput`, and related types in anticipation of future changes. Fix serialization errors and update test vectors.

Closes https://github.com/chain/chain/issues/330.
Closes https://github.com/chain/chain/issues/335.

PR #376, opened 2017-01-11
encoding/blockchain: Encapsulate "varstrlist" and "extensible string" patterns

bobg:
Use the new functions throughout `protocol/bc`.

### DIFF
--- a/encoding/blockchain/blockchain.go
+++ b/encoding/blockchain/blockchain.go
@@ -4,12 +4,13 @@ package blockchain
 
 import (
 	"bytes"
-	"chain/encoding/bufpool"
 	"encoding/binary"
 	"errors"
 	"io"
 	"math"
 	"sync"
+
+	"chain/encoding/bufpool"
 )
 
 var bufPool = sync.Pool{New: func() interface{} { return new([9]byte) }}
--- a/encoding/blockchain/blockchain.go
+++ b/encoding/blockchain/blockchain.go
@@ -3,6 +3,8 @@
 package blockchain
 
 import (
+	"bytes"
+	"chain/encoding/bufpool"
 	"encoding/binary"
 	"errors"
 	"io"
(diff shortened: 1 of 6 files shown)


bobg:
Carves off part of https://github.com/chain/chain/pull/349 for easier review.

bobg:
PTAL

kr:
:eyes:

protocol/bc/txinput.go
```diff
@@ -297,15 +274,16 @@ func (t *TxInput) readFrom(r io.Reader, txVersion uint64) (err error) {
 // assumes w has sticky errors
 func (t *TxInput) writeTo(w io.Writer, serflags uint8) {
 	blockchain.WriteVarint63(w, t.AssetVersion) // TODO(bobg): check and return error
-	buf := bufpool.Get()
-	defer bufpool.Put(buf)
-	t.WriteInputCommitment(buf)
-	blockchain.WriteVarstr31(w, buf.Bytes())
+	blockchain.WriteExtensibleString(w, func(w io.Writer) error {
```
kr: Can this be

```
blockchain.WriteExtensibleString(w, t.WriteInputCommitment)
```

and same for writeInputWitness below?

encoding/blockchain/blockchain.go
```diff
@@ -91,6 +94,78 @@ func WriteVarstr31(w io.Writer, str []byte) (int, error) {
...
+			return nil, n, err
+		}
+		result = append(result, s)
+	}
+	return result, n, nil
+}
+
+// WriteExtensibleString sends the output of the given function,
+// together with a varint31 length prefix, to w.
+func WriteExtensibleString(w io.Writer, f func(io.Writer) error) (int, error) {
```
kr: Can/should we document that WriteExtensibleString always provides a *bytes.Buffer to f, and that the signature of f is this way mainly so that it matches io.WriterTo? That way, f can safely ignore errors from its Writer (which is guaranteed to be a *bytes.Buffer, which never returns an error from Write). That would let us clean up some (or maybe just one) of the call sites.

protocol/bc/txinput.go
bobg: Yes. I had to revise those functions to return errors first, but nice.

encoding/blockchain/blockchain.go
bobg: Eh, I don't find that suggestion especially compelling. Eliminating two or three error checks in a Go program feels like spitting into the ocean.

bobg:
Revised, PTAL

kr:
LGTM

PR #377, opened 2017-01-11
core/txdb: delete older snapshots

jbowens:
After saving a new snapshot, delete older snapshots.

We could change the snapshots table to be a singleton table, but the
/get-snapshot RPC uses content-addressable snapshots. There would be
a race condition between when a client looks up the height of the
most recent snapshot and when they retrieve that snapshot.

There might be other motivations for keeping more snapshots
than just the most recent snapshot?

### DIFF
--- a/core/txdb/snapshot.go
+++ b/core/txdb/snapshot.go
@@ -75,14 +75,14 @@ func storeStateSnapshot(ctx context.Context, db pg.DB, snapshot *state.Snapshot,
 
 	const insertQ = `
 		INSERT INTO snapshots (height, data) VALUES($1, $2)
-		ON CONFLICT (height) DO UPDATE SET data = $2
+		ON CONFLICT (height) DO UPDATE SET data = $2, created_at = NOW()
 	`
 	_, err = db.Exec(ctx, insertQ, blockHeight, b)
 	if err != nil {
 		return errors.Wrap(err, "writing state snapshot to database")
 	}
 
-	const deleteQ = `DELETE FROM snapshots WHERE timestamp < NOW() - INTERVAL '24 hours'`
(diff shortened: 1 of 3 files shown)


jbowens:
Maybe we should do this by time instead of by blocks? We'd need to add a new column to snapshots but it'd make a lot more sense to delete by age in time instead of blocks.

core/txdb/snapshot.go
```diff
@@ -77,9 +80,14 @@ func storeStateSnapshot(ctx context.Context, db pg.DB, snapshot *state.Snapshot,
...
 		ON CONFLICT (height) DO UPDATE SET data = $2
 	`
-
 	_, err = db.Exec(ctx, insertQ, blockHeight, b)
-	return errors.Wrap(err, "writing state snapshot to database")
+	if err != nil {
+		return errors.Wrap(err, "writing state snapshot to database")
+	}
+
+	const deleteQ = `DELETE FROM snapshots WHERE timestamp < $1`
```
kr: Would it be safer to do the time arithmetic in sql? Doing the subtraction in Go, if a cored has a misconfigured clock, it can accidentally delete all the snapshots. A cored clock that badly out of sync would cause other problems, for sure, but at least maybe we can avoid this one.

jbowens:
> Looking ahead, maybe snapshots would be a good early candidate for storing locally on Chain Core nodes? Especially since it's a very large object (and maybe we can do with just one or two of them at a time), I can imagine writing it directly to a file rather than using some fancy storage engine.

Agreed

jbowens:
updated, PTAL

jbowens:
PTAL

jbowens:
PTAL

erykwalder:
Only suggestion might be to use the column name "created_at" instead of "timestamp". That way there's also no confusion between type and field.

erykwalder:
LGTM

PR #378, opened 2017-01-11
core: use bytea for hash columns

erykwalder:
Replaces all the hash columns in the database that use text with ones
that use bytea.


jbowens:
lgtm

PR #380, opened 2017-01-13
core/account: introduce receivers

jbowens:
Add a create-account-receiver endpoint to create expiring control
programs. By default, a receiver expires in 30 days. Control programs
created through the create-control-program rpc do not expire.
Control programs created automatically during transaction building
are created with the transaction's max time as an expiration.

The expiration is only considered to be elapsed once the Core ingests a
block with a timestamp greater than the control program's expiration.
This allows us to use a transaction's max time as the control program's
expiry without worrying about clock skew. Once the expiration has
elapsed, the control programs are deleted and forgotten.

If any assets are sent to an expired control program, the Core will not
index them (so they will not be spendable) or annotate them as belonging
to any account.

This PR adds a new server RPC but does not expose it as an interface
in the SDK yet.

### DIFF
--- a/core/account/accounts.go
+++ b/core/account/accounts.go
@@ -239,13 +239,10 @@ func (m *Manager) insertAccountControlProgram(ctx context.Context, progs ...*con
 		keyIndexes = append(keyIndexes, int64(p.keyIndex))
 		controlProgs = append(controlProgs, p.controlProgram)
 		change = append(change, p.change)
-
-		var expiry stdsql.NullString
-		if !p.expiresAt.IsZero() {
-			expiry.String = p.expiresAt.Format(time.RFC3339)
-			expiry.Valid = true
-		}
-		expirations = append(expirations, expiry)
+		expirations = append(expirations, stdsql.NullString{
+			String: p.expiresAt.Format(time.RFC3339),
+			Valid:  !p.expiresAt.IsZero(),
+		})
 	}
 
 	_, err := m.db.Exec(ctx, q, accountIDs, keyIndexes, controlProgs, change, pq.Array(expirations))
--- a/core/account/accounts.go
+++ b/core/account/accounts.go
@@ -177,9 +177,10 @@ type controlProgram struct {
 	keyIndex       uint64
 	controlProgram []byte
 	change         bool
+	expiresAt      time.Time
 }
 
-func (m *Manager) createControlProgram(ctx context.Context, accountID string, change bool) (*controlProgram, error) {
+func (m *Manager) createControlProgram(ctx context.Context, accountID string, change bool, expiresAt time.Time) (*controlProgram, error) {
 	account, err := m.findByID(ctx, accountID)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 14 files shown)


jbowens:
@devongundry, @jeffomatic — this okay from a product perspective?

jeffomatic:
@jbowens I guess this is to prevent unbounded expansion of the control program index for accounts?

jbowens:
> I guess this is to prevent unbounded expansion of the control program index for accounts?

Yeah, that's right.

jeffomatic:
I think I'd need a chance to review this with @devongundry and the protocol team before I can have an informed opinion here.

kr:
This definitely impacts the product design in other ways, for example any "payment request" protocol we design should be sure to include the expiry time so that a potential payer doesn't try to send money after the cutoff. But it's important that we establish this convention from the outset, so that people don't get used to the idea that they can just hang on to control programs indefinitely.

I'm happy to join in whatever meeting you folks have if it would help. This has been on our list of things to do for a long time; I'm glad we're finally getting to it.

jbowens:
!PTAL, on hold until we have acceptance programs

jbowens:
Is this actually blocked on acceptance programs? If we want to get interface changes in sooner rather than later for impending deadlines, maybe we should implement expiry first. Then additional safety mechanisms (acceptance programs or whatever else) just need to make it into the protocol before new networks are deployed.

jbowens:
@kr please take a new look :)

core/account/accounts.go
```diff
@@ -222,23 +223,32 @@ func (m *Manager) CreateControlProgram(ctx context.Context, accountID string, ch
...
 		accountIDs = append(accountIDs, p.accountID)
 		keyIndexes = append(keyIndexes, int64(p.keyIndex))
 		controlProgs = append(controlProgs, p.controlProgram)
 		change = append(change, p.change)
+
+		var expiry stdsql.NullString
+		if !p.expiresAt.IsZero() {
+			expiry.String = p.expiresAt.Format(time.RFC3339)
+			expiry.Valid = true
+		}
```
kr: minor, but we could avoid a branch by writing:

```go
expirations = append(expirations, stdsql.NullString{
	String: p.expiresAt.Format(time.RFC3339),
	Valid:  !p.expiresAt.IsZero(),
})
```

kr:
LGTM with one minor suggestion

jeffomatic:
Might want to update language of commit message, I think this commit does affect new endpoints?

jbowens:
> Might want to update language of commit message, I think this commit does affect new endpoints?

I think I already updated the PR description? That's what's used for the commit message on landing.

jeffomatic:
I was referring to this line:

> This PR does not expose the new endpoint or interfaces in the SDK yet.

kr:
That line seems accurate to me. Maybe we have a slight mismatch in nomenclature?

jeffomatic:
Yeah I think I see the confusion now. I was reading "This PR does not expose the new endpoint" as the *server* endpoint, which I think [gets exposed](https://github.com/chain/chain/pull/380/files#diff-e3eba3817c54994b15d0337fe489bb42R124) in this PR. But I guess the "in the SDK" part refers to the whole statement?

kr:
It seems like the wording could stand to be clarified. 😁  I took a stab at it.

jeffomatic:
Sorry, totally did not intend to be a grammar jerk, I had misread it.

Issue #382, opened 2017-01-13
Issuance tx fails to balance if asset definition is empty


oleganza:
Works for non-empty asset definitions though, as reported by @chrisgarvin. Will need to prepare a unit test to see exact issue.

oleganza:
Addressed by #386

PR #386, opened 2017-01-13
core/asset: store raw asset definition to ensure stability of AssetID

oleganza:
This addresses https://github.com/chain/chain/issues/382

We also store VMVersion of the issuance program explicitly in the database.

### DIFF
--- a/core/asset/asset.go
+++ b/core/asset/asset.go
@@ -341,7 +341,7 @@ func assetQuery(ctx context.Context, db pg.DB, pred string, args ...interface{})
 // The empty asset def is an empty byte slice.
 func serializeAssetDef(def map[string]interface{}) ([]byte, error) {
 	if def == nil {
-		return nil, nil
+		return []byte{}, nil
 	}
 	return json.MarshalIndent(def, "", "  ")
 }
--- a/core/asset/asset.go
+++ b/core/asset/asset.go
@@ -64,16 +64,39 @@ func (reg *Registry) IndexAssets(indexer Saver) {
 type Asset struct {
 	AssetID          bc.AssetID
 	Alias            *string
-	Definition       map[string]interface{}
 	IssuanceProgram  []byte
 	InitialBlockHash bc.Hash
 	Signer           *signers.Signer
 	Tags             map[string]interface{}
+	rawDefinition    []byte
+	definition       map[string]interface{}
 	sortID           string
 }
 
-func (asset *Asset) SerializedDefinition() ([]byte, error) {
-	return serializeAssetDef(asset.Definition)
+func (asset *Asset) Definition() (map[string]interface{}, error) {
+	if asset.definition == nil {
+		if len(asset.rawDefinition) > 0 {
+			err := json.Unmarshal(asset.rawDefinition, &asset.definition)
+			if err != nil {
+				return nil, errors.Wrap(err)
+			}
+		}
+	}
+	return asset.definition, nil
+}
+
+func (asset *Asset) RawDefinition() []byte {
+	return asset.rawDefinition
+}
+
+func (asset *Asset) SetDefinition(def map[string]interface{}) error {
+	rawdef, err := serializeAssetDef(def)
+	if err != nil {
+		return err
+	}
+	asset.definition = def
+	asset.rawDefinition = rawdef
+	return nil
 }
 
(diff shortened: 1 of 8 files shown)


core/asset/asset.go
```diff
@@ -64,16 +64,40 @@ func (reg *Registry) IndexAssets(indexer Saver) {
...
+	rawDefinition    []byte
+	definition       map[string]interface{}
 	sortID           string
 }
 
-func (asset *Asset) SerializedDefinition() ([]byte, error) {
-	return serializeAssetDef(asset.Definition)
+func (asset *Asset) Definition() (map[string]interface{}, error) {
+	if asset.definition == nil {
+		if len(asset.rawDefinition) > 0 {
```
kr: nit: instead of nested ifs it would be a bit clearer imo to combine them

```
if asset.definition == nil && len(asset.rawDefinition) > 0 {
```

core/asset/block.go
```diff
@@ -32,11 +32,16 @@ func (reg *Registry) indexAnnotatedAsset(ctx context.Context, a *Asset) error {
...
-		"definition":       a.Definition,
+		"raw_definition":   json.HexBytes(a.RawDefinition()),
+		"vm_version":       a.VMVersion,
 		"issuance_program": json.HexBytes(a.IssuanceProgram),
 		"tags":             a.Tags,
 		"is_local":         "no",
 	}
+	adef, err := a.Definition()
+	if err == nil {
+		m["definition"] = adef
```
kr: Setting this conditionally seems like a user-visible change. Under what circumstances would it be omitted (when a.Definition returns an error)? Should we fill in a default value in that case? How would that case have manifested before this patch?

cc @jeffomatic

core/asset/builder.go
```diff
@@ -53,7 +53,7 @@ func (a *issueAction) Build(ctx context.Context, builder *txbuilder.TemplateBuil
 		return err
 	}
 
-	assetdef, err := asset.SerializedDefinition()
+	assetdef := asset.RawDefinition()
 	if err != nil {
```
kr: should we delete this if statement?

core/schema.sql
```diff
@@ -2,11 +2,12 @@
...
 --
 
--- Dumped from database version 9.5.0
--- Dumped by pg_dump version 9.5.0
+-- Dumped from database version 9.6.0
+-- Dumped by pg_dump version 9.6.0
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
```
kr: I think this statement will make pg 9.5 unhappy. idle_in_transaction_session_timeout was added in 9.6.

https://www.postgresql.org/docs/9.6/static/release-9-6.html

So you should prob just delete this line by hand. (sorry)

core/asset/asset.go
oleganza: fixed.

core/asset/builder.go
oleganza: fixed.

core/schema.sql
oleganza: fixed.

core/asset/block.go
oleganza: The previous behaviour as of 6c24fcc504aad981577940d1169d5ec974333b48 (when adef was in the issuance program as json string) would fail to parse non-json definitions and not index such assets. Current behaviour indexes even non-json definitions and returns to the clients both original hex-encoded raw string ("raw_definition"), and return nil for the structured form ("definition").

Note that current Cores never allow creating non-JSON definitions in the first place, so current applications would be unaffected.

jeffomatic: Sorry for the late response here. @oleganza's description fits my understanding--if the definition is not valid JSON, the create-asset API call will fail outright.

Also, following Oleg, I think the sensible default is `nil`. It doesn't really differentiate from the case where the asset definition is empty, but it's better than the alternatives I can think of.

This conversation does suggest that asset defs as exposed by the SDK may be a bit too high-level; previous incarnations of the API exposed byte arrays and left the client to perform JSON serialization/deserialization. On the other hand, tx annotation based on the JSON value was probably what drove the change. Exposing `raw_definition` seems like a good compromise.

PR #388, opened 2017-01-13
docs/protocol: add acceptance program

danrobinson:
This adds a fourth use of VM programs, "acceptance programs". Acceptance programs are included in outputs alongside control programs, are run at the time that output is included in a transaction (rather than when it is spent), and must be satisfied by the transaction in which they are included.

Control programs can use the new `ACCEPTANCEPROGRAM` opcode to confirm, at the time the output is spent, which acceptance program was included in it. This means that if someone changes the acceptance program that is associated with a control program, the output will become unspendable.

This allows control programs to indirectly enforce conditions on transactions that spend to them. For example, the acceptance program can set an expiration time after which it cannot be included in an output, or it can require a signature by the recipient or some other party. The control program can then commit to that acceptance program.

### DIFF
--- a/docs/protocol/specifications/vm1.md
+++ b/docs/protocol/specifications/vm1.md
@@ -1254,13 +1254,24 @@ Pushes the block timestamp in milliseconds on the data stack.
 
 Fails if executed in the [transaction context](#transaction-context).
 
+#### ACCEPTANCEPROGRAM
+
+Code  | Stack Diagram   | Cost
+------|-----------------|-----------------------------------------------------
+0xcf  | (∅ → program)     | 1; [standard memory cost](#standard-memory-cost)
+
+Pushes the [acceptance program](data.md#acceptance-program) from the output being spent.
+
+Fails if the current input is not an [issuance input](data.md#transaction-input-commitment).
+
+Fails if executed in the [block context](#block-context).
 
 
 ### Expansion opcodes
 
 Code  | Stack Diagram   | Cost
 ------|-----------------|-----------------------------------------------------
-0x50, 0x61, 0x62, 0x65, 0x66, 0x67, 0x68, 0x8a, 0x8d, 0x8e, 0xa6, 0xa7, 0xa9, 0xab, 0xb0..0xbf, 0xca, 0xcd..0xcf, 0xd0..0xff  | (∅ → ∅)     | 1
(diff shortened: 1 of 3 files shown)


docs/protocol/specifications/vm1.md
```diff
@@ -1254,13 +1256,24 @@ Pushes the block timestamp in milliseconds on the data stack.
...
+
+Fails if executed in the [block context](#block-context).
 
 
 ### Expansion opcodes
 
 Code  | Stack Diagram   | Cost
 ------|-----------------|-----------------------------------------------------
-0x50, 0x61, 0x62, 0x65, 0x66, 0x67, 0x68, 0x8a, 0x8d, 0x8e, 0xa6, 0xa7, 0xa9, 0xab, 0xb0..0xbf, 0xca, 0xcd..0xcf, 0xd0..0xff  | (∅ → ∅)     | 1
+0x50, 0x61, 0x62, 0x65, 0x66, 0x67, 0x68, 0x8a, 0x8d, 0x8e, 0xa6, 0xa7, 0xa9, 0xab, 0xb0..0xbf, 0xca, 0xd0..0xff  | (∅ → ∅)     | 1
```
kr: Do we need to keep `0xcd..0xce` in this list?

oleganza:
I don't think this is the right way forward. If we are to solve a bigger problem of end-to-end safety (in addition to security), then most transactions will be 2-party trades (or "swaps"). Even the one-way transfers will instead involve a "receipt" asset being returned to the payor.

Example:

1. Customer comes to a shop and wants to buy a pair of shoes.
2. Merchant generates address that times out in 10 minutes.
3. Merchant signs a single issuance entry: 
    1. entry issues a receipt with signature program that checks `before(now+10mins)` and `checkoutput($100 USD -> address)`.
4. Merchant transfers it to the customer's wallet.
5. Wallet verifies that receipt is of known AssetID (e.g. X.509-signed) and matches the product and price being sold.
6. Wallet adds payment inputs, output to collect the receipt and an output that sends $100 to the merchant's address.
7. If customer ignores the request and more than 10 mins pass, merchant can safely retire `address` since the input will not be usable anymore. Merchant does not need to remember issuance input at any point in time.

Why having receipt? Because then shop's owner can verify that all returns and disputes involve authentic receipts and not some insecure pieces of paper. All asset movements are perfectly accounted for.

docs/protocol/specifications/vm1.md
danrobinson: They're actually already used; that's a mistake in the existing spec.

jbowens:
should we close this out?

PR #389, opened 2017-01-14
core/assets: createAsset response encodes IssuanceProgram in hex

oleganza:
Current implementation uses default base64 encoding that's inconsistent with use elsewhere (e.g. https://github.com/chain/chain/blob/27e2c5af0195d71aba8837e7e04fadf2ffbd2069/core/asset/block.go#L36). This change ensures hex-encoding for the issuance program in the "createAsset" response.


erykwalder:
LGTM

Might be good to link to a static file in the commit though:
https://github.com/chain/chain/blob/27e2c5af0195d71aba8837e7e04fadf2ffbd2069/core/asset/block.go#L36

PR #391, opened 2017-01-17
sdk/java: use list of objects for filter params

erykwalder:
Some filter parameters cannot be represented as strings.


jbowens:
lgtm

PR #392, opened 2017-01-17
net/http/limit: extract limit functionality

erykwalder:
Rather than having the limit functionality contained in handler, it is
moved to a separate struct. This is helpful for allowing other protocols
to use limiters.


bobg:
LGTM

PR #395, opened 2017-01-17
protocol/bc: move AssetAmount into asset.go

tessr:
Part of #349

### DIFF
--- a/protocol/bc/asset.go
+++ b/protocol/bc/asset.go
@@ -2,6 +2,7 @@ package bc
 
 import (
 	"database/sql/driver"
+	"io"
 
 	"chain/crypto/sha3pool"
 	"chain/encoding/blockchain"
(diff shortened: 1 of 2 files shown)


protocol/bc/asset.go
```diff
@@ -32,3 +33,28 @@ func ComputeAssetID(issuanceProgram []byte, initialHash [32]byte, vmVersion uint
 	h.Read(assetID[:])
 	return assetID
 }
+
+type AssetAmount struct {
+	AssetID AssetID `json:"asset_id"`
+	Amount  uint64  `json:"amount"`
+}
+
+// assumes r has sticky errors
```
bobg: This comment can be removed. (It could have before, too.)

bobg:
LGTM with one minor suggestion above

PR #396, opened 2017-01-17
core/query: leave unmarshaling to request layer

erykwalder:
Rather than unmarshaling from the database into a struct or map
and then doing conversion at the request layer before sending a
response to the client, just leave the unmarshaling to the request
layer, where it can guarantee ordering.

### DIFF
--- a/core/accounts.go
+++ b/core/accounts.go
@@ -6,22 +6,23 @@ import (
 
 	"chain/core/signers"
 	"chain/crypto/ed25519/chainkd"
+	"chain/encoding/json"
 	"chain/net/http/reqid"
 )
 
 // This type enforces JSON field ordering in API output.
 type accountResponse struct {
-	ID     interface{} `json:"id"`
-	Alias  interface{} `json:"alias"`
-	Keys   interface{} `json:"keys"`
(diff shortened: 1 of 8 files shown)


kr:
:eyes:

jbowens:
somewhat related, what do you think about replacing the `map[string]interface{}` bag of props with explicit structs for annotated objects, like `query.AnnotatedTx`, `query.AnnotatedInput`, etc.

I started trying it out on a wip branch, because I think it would make the work to remove jsonb easier.

erykwalder:
Would the query package know all of the potential annotations that an object would receive?

jbowens:
> Would the query package know all of the potential annotations that an object would receive?

Yeah, it would need to anyways if we're going to convert the tables from jsonb to tables with schemas.

kr:
LGTM

My only feedback is I found it difficult to get my bearings just from the description above. Maybe being just a bit more explicit would be nice, like:

> Rather than unmarshaling from the database into a struct or map
and then doing conversion at the request layer before sending a
response to the client, just leave the unmarshaling to the request
layer, where it can guarantee ordering.

PR #397, opened 2017-01-17
protocol/bc: move OutputCommitment into its own file

tessr:
Part of #349. 

Note that this simply splits out `OutputCommitment` without making some of the other changes to the output commitment that are included in 349.

### DIFF
--- a/protocol/bc/outputcommitment.go
+++ b/protocol/bc/outputcommitment.go
@@ -5,6 +5,7 @@ import (
 	"io"
 
 	"chain/encoding/blockchain"
+	"chain/encoding/bufpool"
 	"chain/errors"
 )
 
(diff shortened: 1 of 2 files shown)


protocol/bc/outputcommitment.go
```diff
@@ -0,0 +1,66 @@
...
+// OutputCommitment contains the commitment data for a transaction
+// output (which also appears in the spend input of that output).
+type OutputCommitment struct {
+	AssetAmount
+	VMVersion      uint64
+	ControlProgram []byte
+}
+
+func (oc *OutputCommitment) writeTo(w io.Writer, assetVersion uint64) (err error) {
+	b := bufpool.Get()
```
bobg: The entire body of this function, except for the `return` at the end, should be replaced with a call to `WriteExtensibleString` (analogous to the `ReadExtensibleString` call in `readFrom`, below).

The callback to `WriteExtensibleString` should still wrap its nested writes within an `if assetVersion == 1` test.

tessr: I ended up leaving OutputCommitment alone in this “code moving” PR - I find sometimes it’s hard to figure out what’s going on when you move code into a new file _and_ make changes, and I definitely tied myself up a bit when I was trying to move this code and rewrite it at the same time. But let me know if it seems gratuitous to have this PR on its own without further changes.

bobg: No, sorry, you're right, you said as much in the PR description. My bad.

bobg:
LGTM

protocol/bc/outputcommitment.go
tessr: Thanks :)

PR #398, opened 2017-01-18
protocol/bc: return errors from WitnessHash functions

tessr:
This trickles down to `CalcMerkleRoot` as well. Part of #349.


bobg:
style nit: please rename PR to "return errors from WitnessHash functions"

otherwise LGTM

PR #399, opened 2017-01-18
protocol/bc: new files for TypedInputs

tessr:
Part of #349. In 349, there are other changes to both `IssuanceInput` and `SpendInput`, but I don't make those changes here.


bobg:
LGTM

PR #401, opened 2017-01-18
all: don't require GOPATH to be set

kr:
Starting in Go 1.8, it is possible to install and use Go
without setting GOPATH. Several of our shell scripts and
instructions assume that GOPATH is set (almost always to
determine the location of installed executables), but
this assumption will become invalid soon. (And I think
it's unreasonable to impose setting GOPATH as an extra
requirement just for our environment.) We can fix this
by calling 'go env GOPATH', which prints the computed
value of GOPATH (either the string the user set, or the
default, which happens to be $HOME/go in Go 1.8).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -33,14 +33,14 @@ Set the `CHAIN` environment variable, in `.profile` in your home
 directory, to point to the root of the Chain source code repo:
 
 ```
-export CHAIN=$GOPATH/src/chain
+export CHAIN=$(go env GOPATH)/src/chain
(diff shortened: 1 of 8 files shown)


kr:
I am not sure about the Dockerfile changes, so I would appreciate particular review attention there.

kr:
PTAL

erykwalder:
I was able to successfully build the dockerfile

erykwalder:
LGTM

PR #403, opened 2017-01-18
docker/ci: fix CHAIN env variable

erykwalder:
Unfortunately, it turns out that docker's ENV instruction cannot run
commands. As such, in this one area it is better to use simple variables
rather than `go env GOPATH`.


erykwalder:
@kr PTAL. My bad. I should have checked that the ENV var was set properly earlier.

jbowens:
lgtm

PR #404, opened 2017-01-18
protocol/bc: return errors when writing transactions

tessr:
Part of #349. Wraps up some todos.

### DIFF
--- a/protocol/bc/block.go
+++ b/protocol/bc/block.go
@@ -106,7 +106,6 @@ func (b *Block) WriteTo(w io.Writer) (int64, error) {
 	return ew.Written(), ew.Err()
 }
 
-// assumes w has sticky errors
 func (b *Block) writeTo(w io.Writer, serflags uint8) {
 	b.BlockHeader.writeTo(w, serflags)
 	if serflags&SerBlockTransactions == SerBlockTransactions {
--- a/protocol/bc/block.go
+++ b/protocol/bc/block.go
@@ -106,6 +106,7 @@ func (b *Block) WriteTo(w io.Writer) (int64, error) {
 	return ew.Written(), ew.Err()
 }
 
+// assumes w has sticky errors
 func (b *Block) writeTo(w io.Writer, serflags uint8) {
 	b.BlockHeader.writeTo(w, serflags)
 	if serflags&SerBlockTransactions == SerBlockTransactions {
--- a/protocol/bc/hash.go
+++ b/protocol/bc/hash.go
@@ -96,12 +96,13 @@ func ParseHash(s string) (h Hash, err error) {
 	return h, errors.Wrap(err, "decode hex")
 }
 
-func writeFastHash(w io.Writer, d []byte) {
+func WriteFastHash(w io.Writer, d []byte) error {
 	if len(d) == 0 {
-		blockchain.WriteVarstr31(w, nil)
-		return
+		_, err := blockchain.WriteVarstr31(w, nil)
+		return err
 	}
 	var h [32]byte
 	sha3pool.Sum256(h[:], d)
-	blockchain.WriteVarstr31(w, h[:])
+	_, err := blockchain.WriteVarstr31(w, h[:])
+	return err
 }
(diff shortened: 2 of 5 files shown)


protocol/bc/hash.go
```diff
@@ -96,12 +96,13 @@ func ParseHash(s string) (h Hash, err error) {
 	return h, errors.Wrap(err, "decode hex")
 }
 
-func writeFastHash(w io.Writer, d []byte) {
+func WriteFastHash(w io.Writer, d []byte) error {
```
erykwalder: does this need to be exported?

protocol/bc/transaction.go
```diff
@@ -271,39 +271,67 @@ func (tx *TxData) MarshalText() ([]byte, error) {
...
 	ew := errors.NewWriter(w)
-	tx.writeTo(ew, serRequired)
+	err := tx.writeTo(ew, serRequired)
+	if err != nil {
+		return ew.Written(), ew.Err()
+	}
 	return ew.Written(), ew.Err()
 }
 
 // assumes w has sticky errors
```
erykwalder: all comments like this are no longer valid/necessary

protocol/bc/transaction.go
```diff
@@ -271,39 +271,67 @@ func (tx *TxData) MarshalText() ([]byte, error) {
...
+		err = ti.writeTo(w, serflags)
+		if err != nil {
+			return errors.Wrapf(err, "writing tx input %d", i)
+		}
 	}
 
-	blockchain.WriteVarint31(w, uint64(len(tx.Outputs))) // TODO(bobg): check and return error
-	for _, to := range tx.Outputs {
-		to.writeTo(w, serflags)
+	blockchain.WriteVarint31(w, uint64(len(tx.Outputs)))
```
erykwalder: Need to capture `err` here

protocol/bc/hash.go
tessr: Oh no, I don't think so

protocol/bc/transaction.go
tessr: thanks 👍

protocol/bc/transaction.go
```diff
@@ -323,10 +350,10 @@ func (p *Outpoint) WriteTo(w io.Writer) (int64, error) {
 }
 
 // assumes w has sticky errors
```
erykwalder: this assumption is no longer the case

protocol/bc/txinput.go
```diff
@@ -199,13 +199,30 @@ func (t *TxInput) readFrom(r io.Reader, txVersion uint64) (err error) {
 }
 
 // assumes w has sticky errors
```
erykwalder: same here

protocol/bc/txoutput.go
```diff
@@ -55,11 +55,28 @@ func (to *TxOutput) readFrom(r io.Reader, txVersion uint64) (err error) {
 }
 
 // assumes r has sticky errors
```
erykwalder: same here

tessr:
the formerly-sticky errors's comments have been un-stuck

protocol/bc/block.go
```diff
@@ -106,7 +106,6 @@ func (b *Block) WriteTo(w io.Writer) (int64, error) {
 	return ew.Written(), ew.Err()
 }
 
-// assumes w has sticky errors
```
erykwalder: This one doesn't check errors, so would assume it has sticky errors. We should either leave comment or check errors.

tessr: I got a bit overzealous 😐  fixed

tessr: well, fixed the comment

tessr: (#349 doesn't touch this, so I for now I will leave this alone too)

erykwalder:
LGTM and sorry for all the back and forth

tessr:
Haha, no worries. Sorry for being sloppy with comments.

PR #407, opened 2017-01-19
core/query: define structs for annotated txs

jbowens:
Define explicit struct for annotated transactions. Annotators operate on these structs instead of the less structured `map[string]interface{}`. Defining explicit structs has the limitation that the core/query package needs to know about all possible top-level attributes. But, that will already be the case when we migrate away from relying on Postgres's jsonb type.


erykwalder:
:eyes:

core/account/annotate.go
```diff
@@ -3,61 +3,46 @@ package account
...
-			if !ok {
-				log.Error(ctx, errors.Wrap(fmt.Errorf("expected map, got %T", m)))
+	for _, tx := range txs {
+		for _, in := range tx.Inputs {
+			if len(in.ControlProgram) == 0 {
 				continue
 			}
-			if asMap["control_program"] == nil {
-				// Issuance inputs don't have control_programs
+			if controlProgramSet[string(in.ControlProgram)] {
```
jbowens: I think that happens down below? 
```go
	inputs[string(in.ControlProgram)] = append(inputs[string(in.ControlProgram)], in)
```
or do I misunderstand?

core/asset/annotate.go
```diff
@@ -2,81 +2,68 @@ package asset
...
-				continue
-			}
-			assetIDStr, ok := out["asset_id"].(string)
-			if !ok {
-				log.Error(ctx, errors.Wrap(fmt.Errorf("bad asset_id type %T", out["asset_id"])))
-				continue
-			}
-			assetIDStrMap[assetIDStr] = true
+		for _, out := range tx.Outputs {
+			assetIDStrMap[hex.EncodeToString(out.AssetID)] = out.AssetID
```
jbowens: Good call

core/asset/annotate_test.go
```diff
@@ -16,95 +20,60 @@ func TestAnnotateTxs(t *testing.T) {
...
 			},
 		},
 	}
-
 	err = reg.AnnotateTxs(ctx, txs)
 	if err != nil {
 		t.Fatal(err)
 	}
 	if !reflect.DeepEqual(txs, want) {
+		spew.Dump(txs)
```
jbowens: oops, yeah. left these in from debugging. i'll incorp into the test error

core/account/annotate.go
erykwalder: Wouldn't the continue skip that?

jbowens: Oh, gotcha now. Good catch

erykwalder:
LGTM

PR #414, opened 2017-01-19
protocol/bc: refactor common witness serialization

tessr:
Part of #349. Preparing for transactions with non-empty common witness


bobg:
LGTM

PR #416, opened 2017-01-20
core/asset: drop 'mutable_definition' column

jbowens:
Drop the `mutable_definition` column from the assets table. We don't
allow mutable asset definitions. This column is never read and is always
set to the default.


kr:
LGTM

(pg_dump noise :cry: tho)

jbowens:
is it time for me to update Postgres?

tessr:
@jbowens gee ya think

jbowens:
okay, I updated and got rid of the noise, since @tessr bullied me into it 😢

tessr:
TIL being mean works!!!!!!!

PR #417, opened 2017-01-20
protocol hardfork: fixed-length OutputID replacing Outpoint

oleganza:
# REPLACED BY https://github.com/chain/chain/pull/421

**Problem**

Outpoint is a variable-length structure `<txid>:<index>` which is 33-40 bytes long (33 bytes for most transactions). It is used by transaction inputs to identify exact output in the UTXO set ("Assets Merkle Tree"). The tree leafs contain `SHA3(output)` which allows save space and requires transactions to carry redundant copies of spent outputs to perform validation (otherwise nodes would have to store the entire outputs instead of their hashes — over 2x more data, and the ratio is much bigger in protocol 2). Also, for HSM-friendliness the TXSIGHASH must contain a redundant output's hash: `SHA3(txid || input index || SHA3(output))`.

**Solution:**

We define two new terms:

* `OutputID = SHA3(TxHash || OutputIndex)`
* `UnspentID = SHA3(OutputID || SHA3(OutputCommitment))`

How are these used:

1. Transaction input contains **OutputID** to identify the output being spent. This is a unique identifier of the output.
2. Transaction input uses **second serialization flag** to indicate if it contains the entire previous Output Commitment, or its hash (instead of empty place).
3. UTXO set becomes a **proper set** containing **UnspentIDs** instead of `{Outpoint -> SHA3(OutputCommitment)}`.

When a node validates a transaction, it computes `UnspentID` using provided `OutputID` and previous `OutputCommitment`. If the given unspent ID is present in the UTXO set, then previous output is proved to be both authentic and available for spending.

**Upsides:**

1. The outputID is constant-size and shorter: 32 bytes instead of 33-40 bytes. This simplifies merkle tree design, transaction data structure and all pieces of software that need to handle outpoints.
2. All outputs (via unspentIDs) in the transaction are randomized across the Assets Merkle Tree instead of being crammed inside a common subpath `<txid>||...`.
3. Inputs automatically commit directly to the spent outputs, so TXSIGHASH does not need to do that and can be simplified to `SHA3(txid || input index)`. HSM is able to verify which output this input commits to without having access to the entire parent transaction.
4. We keep the term _outpoint_ to mean a pair `(txid, index)`, but is internal to Chain Core to support random access to UTXOs. Validation protocol no longer uses outpoints.
5. UTXO takes 2x less RAM because it only contains unpent IDs (32 bytes) instead of a key-value pair (64+ bytes).
6. When we get to _tx entries_ design, we'll generalize the idea of OutputID to EntryID, so that any entry can have a unique identifier.

**Downsides:**

1. OutputID no longer indicates the transaction ID which makes it impossible to navigate the chain of transactions without also having a mapping `outpoint -> txid:index`. UTXO tree is not enough as it's only reflecting the latest state of the chain and throws away spent outpoints. Note that in order to navigate the transactions in practice one still needs the mapping `txid -> tx`, so maintaining one more index might not be a significant increase in complexity. Chain is doing this indexing already and we keep that mapping.
2. Chain Core no longer returns (txid,position) pair for annotated txinputs (called `spent_output:{transaction_id:String,position:Int}`), but instead returns output_id (`spent_output_id:String`). To maintain full compatibility, we'd need to make an additional request to locate the previous output's txid and position, but I'm not sure any application actually relies on such historical data. For spending (locating unspents), we fully maintain compatibility with clients using (txid,position) pairs.

# REPLACED BY https://github.com/chain/chain/pull/421


oleganza:
PTAL @kr @jbowens @tessr  @erykwalder @dominic @jeffomatic @danrobinson @devongundry

core/account/builder.go
```diff
@@ -93,11 +93,10 @@ func (a *spendAction) Build(ctx context.Context, b *txbuilder.TemplateBuilder) e
 	return nil
 }
 
-func (m *Manager) NewSpendUTXOAction(outpoint bc.Outpoint) txbuilder.Action {
+func (m *Manager) NewSpendUTXOAction(outputid bc.OutputID) txbuilder.Action {
```
jbowens: nit: s/outputid/outputID/

core/account/indexer.go
```diff
@@ -89,24 +99,23 @@ func (m *Manager) indexAccountUTXOs(ctx context.Context, b *bc.Block) error {
 	}
 
 	// Delete consumed account UTXOs.
-	deltxhash, delindex := prevoutDBKeys(b.Transactions...)
+	deloutputids := prevoutDBKeys(b.Transactions...)
```
jbowens: nit: s/deloutputids/delOutputIDs/

core/query/index.go
```diff
@@ -108,17 +108,16 @@ func (ind *Indexer) insertAnnotatedOutputs(ctx context.Context, b *bc.Block, ann
...
 	)
 
 	for pos, tx := range b.Transactions {
 		for _, in := range tx.Inputs {
 			if !in.IsIssuance() {
-				outpoint := in.Outpoint()
-				prevoutHashes = append(prevoutHashes, outpoint.Hash[:])
-				prevoutIndexes = append(prevoutIndexes, outpoint.Index)
+				oid := in.OutputID()
+				prevoutOIDs = append(prevoutOIDs, oid[:])
```
jbowens: nit: `OID` might get confused with Postgres's [object identifiers](https://www.postgresql.org/docs/8.1/static/datatype-oid.html). Maybe rename to `prevoutIDs`?

core/schema.sql
```diff
@@ -7,6 +7,7 @@
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
```
jbowens: this is Postgres 9.6 only. should probably delete this since some of us are still on 9.5

protocol/bc/outpoint.go
```diff
@@ -0,0 +1,39 @@
+package bc
+
+import (
+	"chain/encoding/blockchain"
+	"io"
+	"strconv"
```
jbowens: nit: import grouping

oleganza:
Thanks, @jbowens. Addressed all nits.

protocol/state/outputs.go
```diff
@@ -27,19 +30,23 @@ func NewOutput(o bc.TxOutput, p bc.Outpoint) *Output {
...
 }
 
 // OutputKey returns the key of an output in the state tree.
-func OutputKey(o bc.Outpoint) (bkey []byte) {
-	var b bytes.Buffer
-	w := errors.NewWriter(&b) // used to satisfy interfaces
-	o.WriteTo(w)
-	return b.Bytes()
+func OutputKey(o bc.UnspentID) (bkey []byte) {
+	// TODO(oleg): check if we no longer need this buffer writing.
```
kr: You're right, we don't need it. The buffer writing was just to serialize the outpoint. Since an UnspentID's internal representation is the actual bytes we want to put in the tree, we can just return them as you're doing here.

kr: In fact, maybe we should inline (get rid of) this whole function.

protocol/vm/introspection.go
```diff
@@ -193,7 +193,7 @@ func opIndex(vm *virtualMachine) error {
 	return vm.pushInt64(int64(vm.inputIndex), true)
 }
 
-func opOutpoint(vm *virtualMachine) error {
+func opOutputid(vm *virtualMachine) error {
```
kr: nit: opOutputID

kr:
> wording should probably be tweaked

Actually this might just consist of replacing the one word "proposal" with "solution" or something along those lines.

app.json
```diff
@@ -3,16 +3,21 @@
...
+        "version": "9.5"
+      }
     }
   ],
   "env": {
     "GO_INSTALL_PACKAGE_SPEC": "./cmd/...",
     "CMD": "cored"
   }
-}
+}
```
oleganza: yup, reverted.

core/account/builder.go
```diff
@@ -109,27 +108,34 @@ func (m *Manager) DecodeSpendUTXOAction(data []byte) (txbuilder.Action, error) {
...
+		if a.TxOut == nil {
+			return txbuilder.MissingFieldsError("position")
+		}
+		outid = bc.ComputeOutputID(*a.TxHash, *a.TxOut)
+	} else {
+		if a.OutputID == nil {
+			return txbuilder.MissingFieldsError("output_id")
+		}
+		outid = *a.OutputID
+	}
```
oleganza: That's harder to parse what's being meant here... But I get the point, how about this flat list of 3 options: https://github.com/chain/chain/pull/417/commits/36ad7c07b41505885c772102e72406a36ea4d402

core/account/indexer.go
```diff
@@ -51,8 +51,14 @@ func (m *Manager) indexAnnotatedAccount(ctx context.Context, a *Account) error {
 	})
 }
 
-type output struct {
+type outputWithOutpoint struct {
```
oleganza: Good call. Renamed to `rawOutput` and `accountOutput`: 31e2a038b99d51df50b6c26d292a1db49e72d7c2

core/migrate/data.go
```diff
@@ -107,4 +107,12 @@ var migrations = []migration{
 	{Name: "2017-01-19.0.asset.drop-mutable-flag.sql", SQL: `
 		ALTER TABLE assets DROP COLUMN definition_mutable;
 	`},
+	{Name: "2017-01-20.0.core.add-output-id-to-outputs.sql", SQL: `
+		ALTER TABLE annotated_outputs ADD COLUMN output_id bytea NOT NULL;
+		ALTER TABLE ONLY annotated_outputs ADD CONSTRAINT annotated_outputs_unique_output_id UNIQUE (output_id);
```
oleganza: Thanks. Addressed: 3e96e6b

core/migrate/data.go
```diff
@@ -107,4 +107,12 @@ var migrations = []migration{
...
+		ALTER TABLE account_utxos ADD COLUMN output_id bytea NOT NULL;
+		ALTER TABLE ONLY account_utxos ADD CONSTRAINT account_utxos_unique_output_id UNIQUE (output_id);
+		ALTER TABLE account_utxos ADD COLUMN unspent_id bytea NOT NULL;
+		ALTER TABLE ONLY account_utxos ADD CONSTRAINT account_utxos_unique_unspent_id UNIQUE (unspent_id);
```
oleganza: Thanks. Addressed: 3e96e6b

protocol/bc/outputid.go
```diff
@@ -0,0 +1,60 @@
...
+	h := sha3pool.Get256()
+	defer sha3pool.Put256(h)
+	h.Write(oid[:])
+	h.Write(outputHash[:])
+	h.Read(uid[:])
+	return uid
+}
+
+// WriteTo writes p to w.
+// It assumes w has sticky errors.
```
oleganza: done.

protocol/vm/introspection.go
oleganza: fixed.

protocol/state/outputs.go
oleganza: Removed. And added convenience `Bytes()->[]byte` methods to avoid creating local var all the time to do a slice. Is it a good idea? 7327600

protocol/bc/outputid.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"database/sql/driver"
+	"io"
+
+	"chain/crypto/sha3pool"
+	"chain/encoding/blockchain"
+)
+
+// OutputID identifies previous transaction output in transaction inputs.
+type OutputID Hash
```
oleganza: Thanks, I didn't know about this feature. I've also tried to use `outid.Bytes()` instead of `tmp := outid.Hash; tmp[:]` when I need a byteslice. Is that the right way? 9042b59937f3f507bed9864049bacf8ba1724c60

oleganza:
Replaced by https://github.com/chain/chain/pull/421

protocol/bc/outputid.go
kr: > I've also tried to use outid.Bytes() instead of tmp := outid.Hash; tmp[:] when I need a byteslice. Is that the right way?

I see nothing wrong with writing `outid.Hash[:]`, but I guess when the expression `outid` is actually a function call, it can't be written without a temporary variable. The `Bytes` method seems reasonable to me! But I wouldn't get dogmatic about using it everywhere; if `outid.Hash[:]` in one line is also an option, that's fine too IMO.

PR #419, opened 2017-01-20
core/query: define annotated accounts & assets

jbowens:
Add typed structs for annotated accounts and annotated assets instead of
using a freeform `map[string]interface{}`. This will be helpful when
replacing the jsonb `data` columns with a structured schema.

Finishing up the work from #407.

### DIFF
--- a/core/account/indexer.go
+++ b/core/account/indexer.go
@@ -27,11 +27,7 @@ type Saver interface {
 	SaveAnnotatedAccount(context.Context, *query.AnnotatedAccount) error
 }
 
-func (m *Manager) indexAnnotatedAccount(ctx context.Context, a *Account) error {
-	if m.indexer == nil {
-		return nil
-	}
-
+func Annotated(a *Account) (*query.AnnotatedAccount, error) {
 	aa := &query.AnnotatedAccount{
 		ID:     a.ID,
 		Alias:  a.Alias,
@@ -40,7 +36,7 @@ func (m *Manager) indexAnnotatedAccount(ctx context.Context, a *Account) error {
 
 	tags, err := json.Marshal(a.Tags)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	rawTags := json.RawMessage(tags)
 	aa.Tags = &rawTags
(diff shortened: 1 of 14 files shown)


jbowens:
Go 1.8 will clean some of this up because we can define the required json fields as `json.RawMessage` instead of a `*json.RawMessage`.

core/query/annotated.go
```diff
@@ -66,6 +67,39 @@ type SpentOutput struct {
...
+}
+
+type AccountKey struct {
+	RootXPub              chainkd.XPub         `json:"root_xpub"`
+	AccountXPub           chainkd.XPub         `json:"account_xpub"`
+	AccountDerivationPath []chainjson.HexBytes `json:"account_derivation_path"`
+}
+
+type AnnotatedAsset struct {
+	ID              chainjson.HexBytes `json:"id"`
```
erykwalder: could use `bc.AssetID`, I think?

core/account/indexer.go
```diff
@@ -22,33 +24,41 @@ const PinName = "account"
...
+	SaveAnnotatedAccount(context.Context, *query.AnnotatedAccount) error
 }
 
 func (m *Manager) indexAnnotatedAccount(ctx context.Context, a *Account) error {
 	if m.indexer == nil {
 		return nil
 	}
-	var keys []map[string]interface{}
+
+	aa := &query.AnnotatedAccount{
```
erykwalder: Maybe the part of this that creates the AnnotatedAccount could be extracted and used in core as well? Same for AnnotatedAsset?

erykwalder:
Feedbacked

core/query/annotated.go
jbowens: I think I originally avoided the `bc` types in this package for consistency between fields, so that all of the hex-encoded byte arrays on the annotated structs use the same type. Looking at it again, that might not be helpful/useful/clear. I'll switch em over to using the `bc` types where possible and see how it looks.

erykwalder:
LGTM

PR #420, opened 2017-01-21
testutil: add func DeepEqual, treats empty as equal to nil

bobg:
The new function treats `nil` as equal to zero-length slices and maps, and is otherwise mostly the same as `reflect.DeepEqual`. This slightly more relaxed definition of equality should reduce the time wasted on hard-to-track-down encoding errors that sometimes convert e.g. `[]byte{}` to `nil` and vice versa.

This PR also replaces uses of `reflect.DeepEqual` in `core` and `protocol` with `testutil.DeepEqual`.

### DIFF
--- a/testutil/deepequal.go
+++ b/testutil/deepequal.go
@@ -2,10 +2,9 @@ package testutil
 
 import "reflect"
 
-// DeepEqual is similar to reflect.DeepEqual, but treats
-// nil and []T{} as equal.
-// (Note, it's also a more naive implementation that doesn't detect
-// cycles).
+// DeepEqual is similar to reflect.DeepEqual, but treats nil as equal
+// to empty maps and slices.
+// (It's also a more naive implementation that doesn't detect cycles).
 func DeepEqual(x, y interface{}) bool {
 	vx := reflect.ValueOf(x)
 	vy := reflect.ValueOf(y)
@@ -13,7 +12,7 @@ func DeepEqual(x, y interface{}) bool {
 }
 
 func deepValueEqual(x, y reflect.Value) bool {
-	if isNilish(x) && isNilish(y) {
+	if isEmpty(x) && isEmpty(y) {
 		return true
 	}
 	if !x.IsValid() {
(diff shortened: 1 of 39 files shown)


bobg:
PTAL

kr:
:eyes:

testutil/deepequal.go
```diff
@@ -0,0 +1,129 @@
...
+// (Note, it's also a more naive implementation that doesn't detect
+// cycles).
+func DeepEqual(x, y interface{}) bool {
+	vx := reflect.ValueOf(x)
+	vy := reflect.ValueOf(y)
+	return deepValueEqual(vx, vy)
+}
+
+func deepValueEqual(x, y reflect.Value) bool {
+	if isNilish(x) && isNilish(y) {
```
kr: I think if we do this after the validity check then we can get rid of the extra check for IsValid inside isNilish.

testutil/deepequal.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+		return true
+
+	case reflect.Func:
+		return x.IsNil() && y.IsNil()
+	}
+	return false
+}
+
+func isNilish(v reflect.Value) bool {
```
kr: Would you consider using isEmpty as the name for this? "Empty" is already an established term in Go, and it means what we're trying to do here. A nil slice or map is empty for the purpose of reads.

testutil/deepequal.go
```diff
@@ -0,0 +1,129 @@
+package testutil
+
+import "reflect"
+
+// DeepEqual is similar to reflect.DeepEqual, but treats
+// nil and []T{} as equal.
```
kr: This is true for maps as well right? Maybe something like "…but treats empty maps and slices as equal to nil" to use conventional nomenclature.

kr:
Oh also, this is awesome. 😍 I forgot to put that in my previous comment. Very nice.

testutil/deepequal.go
bobg: No alas; that breaks the all-important case of `nil` (which is not "valid") vs. `[]byte{}` (which is).

kr:
LGTM

Issue #335, opened 2016-12-12
Transaction witness hash computed incorrectly


bobg:
The spec includes the hash of the transaction common witness fields when computing the transaction witness hash, but the implementation does not.

(The fact that the common witness fields are presently empty doesn't mean that we can omit the empty-string's hash from the computation.)

Issue #330, opened 2016-12-09
Transaction signature hash computed incorrectly


bobg:
The spec requires transaction signature hashes to be made from, among other things, an [output commitment hash](https://github.com/chain/chain/blob/main/docs/protocol/specifications/data.md#transaction-signature-hash). The current implementation hashes [an extensible string containing the output commitment](https://github.com/chain/chain/blob/3a4a6b9ab9a6c910990280bb5df60de2e7ece19a/protocol/bc/txoutput.go#L124), which is wrong - it should be the bare concatenation of output commitment fields that gets hashed.

PR #423, opened 2017-01-23
protocol/bc: fix tx sighash and witness hash bugs

bobg:
Splits `OutputCommitment.writeTo` into explicit `writeExtensibleString` and `writeContents` methods for clarity.

Closes https://github.com/chain/chain/issues/330 - sighash was wrongly including the output commitment's extensible-string prefix length.
Closes https://github.com/chain/chain/issues/335 - witness hash was wrongly omitting hash of (empty) witness.


bobg:
PTAL

bobg:
(Note, this splits out the bugfixing part of #349.)

tessr:
commit message nit (com-nit?): 

maybe add something like "remove wrapping extensible string from the output commitment hash and add transaction common witness fields in the transaction witness hash" to this commit message body so that a reader doesn't have to dive into those issues

protocol/bc/transaction.go
```diff
@@ -232,6 +239,23 @@ func (tx *Tx) WitnessHash() (hash Hash, err error) {
 	return hash, nil
 }
 
+func (tx *TxData) commonWitnessHash() (h Hash, err error) {
```
tessr: maybe add a comment that this is writing an empty common witness.

bobg: It is today, but it may not be in the future. Added the suggested comment to `writeCommonWitness` instead.

PR #425, opened 2017-01-24
WIP: dashboard: onboarding tutorial

chrisgarvin:


dominic:
One key addition would be persisting the tutorial state across app refreshes. You can add the `tutorial` key to https://github.com/chain/chain/blob/main/dashboard/src/utility/localStorage.js to make that sticky.

dashboard/src/features/shared/components/BaseNew.jsx
```diff
@@ -6,7 +6,11 @@ export const mapStateToProps = ( type ) => ( /* state */ ) => ({
 })
 
 export const mapDispatchToProps = (type) => (dispatch) => ({
-  submitForm: (data) => dispatch(actions[type].submitForm(data)),
+  submitForm: (data) => {
+    return dispatch(actions[type].submitForm(data)).then(() => {
+      dispatch(actions.tutorial.updateTutorial(data, type))
```
dominic: This function should return `data` so that the calling function can chain more `.then`s onto the result if desired.

dashboard/src/features/tutorial/components/Success/Success.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import styles from './Success.scss'
+import { Link } from 'react-router'
+
+class Success extends React.Component {
```
dominic: I think `Success` and `Description` are similar enough as to share code and styles with a little refactoring.

dashboard/src/features/tutorial/components/Success/Success.jsx
```diff
@@ -0,0 +1,52 @@
...
+            {this.props.dismiss &&
+              <div className={styles.skip}>
+                <a onClick={this.props.handleDismiss}>{this.props.dismiss}</a>
+              </div>
+            }
+          </div>
+          <div className={styles.content}>
+            <span className='glyphicon glyphicon-ok-sign'></span>
+            <div className={styles.text}>
+              {this.props.content.map(function (x, i){
```
dominic: Personal preference: `x` is a bit short for a variable name; maybe `contentLine` or `line` are good substitutions here.

dashboard/src/features/tutorial/components/Success/Success.jsx
```diff
@@ -0,0 +1,52 @@
...
+                let str = x['line']
+                if (x['type']) {
+                  if (x['type'] == 'account'){
```
dominic: You could flip this logic around a little bit. Something like:

```js
let replacement = userInput[x['type']]
if (x['index']) {
  replacement = replacement[x['index']]
}
str = str.replace(replacement['alias'])
```

would basically only go down the `index` path if it's present (and works if something other than accounts uses multiple values in the future).

dashboard/src/features/tutorial/components/Tutorial.jsx
```diff
@@ -0,0 +1,60 @@
...
+      <div>
+      {tutorialOpen && (tutorialTypes.includes(steps[tutorialStep]['component'])) &&
+          <TutorialComponent
+            userInput={userInput}
+            step={tutorialStep}
+            button={steps[tutorialStep]['button']}
+            title={steps[tutorialStep]['title']}
+            content={steps[tutorialStep]['content']}
+            dismiss={steps[tutorialStep]['dismiss']}
+            route={steps[tutorialStep]['route']}
```
dominic: If it doesn't cause problems, you can splat `steps[tutorialStep]` as component props:

```jsx
<TutorialComponent
  userInput={userInput}
  step={tutorialStep}
  {...steps[tutorialStep]}
  handleNext={this.props.showNextStep}
  handleDismiss={this.props.dismissTutorial}
/>
```

Not committed to that one way or another; it will add at least one unintended property, `component`, to the component too.

dominic:
I also think we will need to revisit some design stuff a little bit. It's a little hard to differentiate the tutorial. I prefer some of Zarya's original, more punchy designs in that direction, but there's a bunch of other options. That's a conversation we should have with Jeff and Devon as a group prolly.

screen shot 2017-01-24 at 5 39 25 pm

Issue #366, opened 2017-01-09
Spec deviation: asset version 1 output loses suffix data during serialization/hashing


oleganza:
Output commitments are extensible for all asset versions. It's only tx v1 that disallows use of extra space, but tx v2 might allow that even for existing asset version. Therefore the following code looks incorrect. When we read AV1 output, we should remember the suffix data, and enforce it being empty string only if tx version == 1, but otherwise we need to keep it around in the data structure. Same for AV2 etc: need to read full data and keep it around properly.

Immutable snippet: https://github.com/chain/chain/blob/359827001fc222136648cd34158f80cbbd19c03c/protocol/bc/txoutput.go#L119-L123

Snippet on main: https://github.com/chain/chain/blob/main/protocol/bc/txoutput.go#L119-L123

oleganza:
WDYT @bobg @kr? Do i understand the piece of code above correctly?

kr:
+1, this is a spec violation bug. There are probably several more like this.

In general, I believe we currently only read the exact fields we understand, and discard fields with future versions during deserialization. I think we do this pretty much across the board.

oleganza:
Ok, gotta review and fix all of these then.

bobg:
It would be good if the spec were clarified to say that unrecognized-but-allowed suffixes on input should be preserved for later output.

PR #426, opened 2017-01-24
encoding/blockchain: save and restore unconsumed extensible-string suffixes

bobg:
Remove the `all` parameter from `ReadExtensibleString`, which now always returns the unconsumed suffix. `WriteExtensibleString` now takes a suffix to always append.

Types with fields parsed from an extensible string now also store the unrecognized suffix to add back when writing.

Block commitment and witness types are moved to their own files, for improved clarity.

Closes https://github.com/chain/chain/issues/366 - unrecognized extensible string suffixes were being dropped

### DIFF
--- a/encoding/blockchain/blockchain.go
+++ b/encoding/blockchain/blockchain.go
@@ -7,6 +7,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"io/ioutil"
 	"math"
 	"sync"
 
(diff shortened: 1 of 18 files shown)


bobg:
PTAL

bobg:
PTAL

encoding/blockchain/blockchain.go
```diff
@@ -133,37 +134,39 @@ func ReadVarstrList(r io.Reader) ([][]byte, int, error) {
...
-var ErrLeftover = errors.New("extensible string partially unconsumed")
-
-// ReadExtensibleString reads a varint31 length prefix and then calls
-// the given function to consume that many bytes. If all is true, it
-// is an error for any bytes to be left over.
-func ReadExtensibleString(r io.Reader, all bool, f func(io.Reader) error) (int, error) {
+// ReadExtensibleString reads a varint31 length prefix and that many
+// bytes from r. It then calls the given function to consume those
+// bytes, returning any unconsumed suffix.
+func ReadExtensibleString(r io.Reader, f func(io.Reader) error) ([]byte, int, error) {
```
tessr: nit: named return parameters here might improve readability

bobg: Done, thanks.

tessr:
LGTM

bobg:
Note that even though this is labeled "blockchain reset," in practice only blockchains that were using extensible-string suffixes are affected, which it's almost certain none were.

PR #427, opened 2017-01-24
core/blocksigner: remove mockhsm dependency

jbowens:
Replace the direct dependency on the core/mockhsm package with a
`Signer` interface. In production, we can pass in a Signer
implementation that talks to signerd instead.

### DIFF
--- a/core/blocksigner/blocksigner.go
+++ b/core/blocksigner/blocksigner.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


jbowens:
Toyed with passing in a sign func instead but it seemed like a little too intense of a function signature

ameets:
@kr and I were talking about using a signer interface offline - this makes a lot of sense to me.

jbowens:
PTAL

PR #428, opened 2017-01-24
core/query: replace jsonb implementation

jbowens:
Reimplement transaction indexing and query filtering using
ordinary SQL tables with well-defined schemas instead of
jsonb blobs and the jsonb `@>` operator.

Environment expressions like `inputs(...)` or `outputs(...)` are
implemented via existence subqueries against the `annotated_inputs`
and `annotated_outputs` tables respectively.

With this implementation, we can now support true booleans and
filter queries like `asset_is_local`. The type checker will need to
be updated before that can be allowed. The boolean logic is
translated into SQL boolean expressions. Also, since we now have
a well-defined schema, the type checker can check types more
exhaustively by examining the types in the schema.

For now, booleans are still text. Because `'yes'` and `'no'` happen
[to be valid PostgreSQL boolean literal values](https://www.postgresql.org/docs/9.5/static/datatype-boolean.html), existing queries like
`asset_is_local = 'yes'` will continue to work without additional
changes. When we add support for true booleans, we'll need to
make sure not to break backwards compatibility.

Without any extra SQL indexes, queries do seem to be an order of
magnitude faster. More importantly, this opens up the door to defining
SQL indexes based on the filter queries that a user expects to use
in their application. Also, we'll be able to build separate SQL tables
to serve as balance query indexes, aggregating as much as possible
at indexing time.

After 864,000 two-output issuance transactions:
```
          Table           |    Size    | External Size
--------------------------+------------+---------------
 annotated_txs            | 2311847936 |     896270336
 annotated_outputs        |  791093248 |     285319168
 account_utxos            |  657170432 |     303267840
 annotated_inputs         |  550019072 |     133439488
 account_control_programs |  370368512 |     154337280
 blocks                   |  300761088 |     300326912
```

**Future work:**
- Updating the type checker to use the schema
- Updating language to allow boolean expressions
- Support defining indexes for tx, unspent outputs, etc.
- Support defining aggregated balance indexes


jbowens:
PTAL

jbowens:
I think there's still some ambiguity with queries that access fields of arbitrary json (asset defintion, tags, reference data). These expressions are converted to PostgreSQL as the [->> operator](https://www.postgresql.org/docs/9.5/static/functions-json.html). The `->>` operator will always return the JSON value as the PostgreSQL `text` type, even if the value in JSON is a number, object, boolean, etc.

We type check filter queries to prevent queries that compare mismatched types (ex. `'hello' = 5`), but when selecting from arbitrary json, the resulting value is untyped. This means a query like:

```
asset_definition.quota = 5
```
can fail in PostgreSQL with a mismatched types error:

```
jackson=# SELECT '{"quota": 5}'::jsonb->>'quota' == 5;
ERROR:  operator does not exist: text == integer
LINE 1: SELECT '{"quota": 5}'::jsonb->>'quota' == 5;
                                               ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

We could typecheck the AST and then use the known types to cast untyped values appropriately?

bobg:
Wow, man. Sweet.

Im'ma start reviewing this soon, but first I have to ask: any chance of splitting it into a series of smaller PRs?

erykwalder:
should've put eyes on this. I'm about 3/4 of the way through

jbowens:
> any chance of splitting it into a series of smaller PRs?

Yeah, I've been thinking about that. The one possibility that I can think of is adding the new columns, tables and indexing in one PR but don't make use of it. Then in a subsequent PR, drop the `data` jsonb columns and swap the filter-to-SQL implementations.

core/query/sql.go
```diff
@@ -0,0 +1,91 @@
...
+			"alias":  {Name: "alias", Type: filter.String, SQLType: filter.SQLText},
+			"quorum": {Name: "quorum", Type: filter.Integer, SQLType: filter.SQLInteger},
+			"tags":   {Name: "tags", Type: filter.Object, SQLType: filter.SQLJSONB},
+		},
+	}
+	outputsTable = &filter.SQLTable{
+		Name:  "annotated_outputs",
+		Alias: "out",
+		Columns: map[string]*filter.SQLColumn{
+			"id":               {Name: "id", Type: filter.String, SQLType: filter.SQLBytea},
```
erykwalder: `output_id`?

core/query/sql.go
```diff
@@ -0,0 +1,91 @@
...
+			"asset_tags":       {Name: "asset_tags", Type: filter.Object, SQLType: filter.SQLJSONB},
+			"asset_is_local":   {Name: "asset_local", Type: filter.String, SQLType: filter.SQLBool},
+			"amount":           {Name: "amount", Type: filter.Integer, SQLType: filter.SQLBigint},
+			"account_id":       {Name: "account_id", Type: filter.String, SQLType: filter.SQLText},
+			"account_alias":    {Name: "account_alias", Type: filter.String, SQLType: filter.SQLText},
+			"account_tags":     {Name: "account_tags", Type: filter.Object, SQLType: filter.SQLJSONB},
+			"issuance_program": {Name: "issuance_program", Type: filter.String, SQLType: filter.SQLBytea},
+			"reference_data":   {Name: "reference_data", Type: filter.Object, SQLType: filter.SQLJSONB},
+			"is_local":         {Name: "local", Type: filter.Bool, SQLType: filter.SQLBool},
+			// TODO(jackson): moar?
```
erykwalder: moar what?

erykwalder:
This is looking really good

core/query/accounts.go
```diff
@@ -13,16 +14,19 @@ import (
...
 
 	const q = `
-		INSERT INTO annotated_accounts (id, data) VALUES($1, $2)
-		ON CONFLICT (id) DO UPDATE SET data = $2
+		INSERT INTO annotated_accounts (id, alias, keys, quorum, tags)
+		VALUES($1, $2, $3::jsonb, $4, $5::jsonb)
+		ON CONFLICT (id) DO UPDATE SET tags = $5::jsonb
 	`
-	_, err = ind.db.Exec(ctx, q, account.ID, b)
+	_, err = ind.db.Exec(ctx, q, []byte(account.ID), alias, keysJSON,
```
jbowens: whoop

core/query/sql.go
jbowens: lol i think i thought i was missing some fields. pretty sure I got em all now

jbowens:
updated

erykwalder:
Looks good to me

PR #429, opened 2017-01-24
dashboard: remove actionCreator

dominic:


dominic:
PTAL

dashboard/src/features/core/actions.js
```diff
@@ -1,10 +1,9 @@
 import chain from '_chain'
 import { context } from 'utility/environment'
-import { actionCreator } from 'features/shared/actions'
 
-const updateInfo = actionCreator('UPDATE_CORE_INFO', param => ({ param }))
-const setClientToken = actionCreator('SET_CLIENT_TOKEN', token => ({ token }))
-const clearSession = actionCreator('USER_LOG_OUT')
+const updateInfo = (param) => ({type: 'UPDATE_CORE_INFO', param})
+const setClientToken = (param) => ({type: 'SET_CLIENT_TOKEN', param})
```
chrisgarvin: i think `param` should either be passed through as `token`, or the core reducer needs to return `action.param` instead of `action.token`

chrisgarvin:
LGTM

PR #435, opened 2017-01-25
core/txbuilder: add explicit retire action

jbowens:
Add an explicit retire action instead of using the `control_program`
action and specifying the OP_FAIL program. This provides more
flexibility as our implementation of retirement changes.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -100,6 +100,7 @@ func (h *Handler) init() {
 		"control_account":                h.Accounts.DecodeControlAction,
 		"control_program":                txbuilder.DecodeControlProgramAction,
 		"issue":                          h.Assets.DecodeIssueAction,
+		"retire":                         txbuilder.DecodeRetireAction,
 		"spend_account":                  h.Accounts.DecodeSpendAction,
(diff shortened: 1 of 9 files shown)


jbowens:
Resurrecting this. @jeffomatic should I remove the SDK changes or should we just bump the SDK version when this goes out?

jeffomatic:
1. By my understanding, this should be backward compatible with older SDKs, since the old `control_with_program` technique would still work, right?
2. We've added a Node SDK. @dominic or @chrisgarvin should be able to point you to the right spot.

dominic:
re Node SDK: https://github.com/chain/chain/blob/main/sdk/node/src/api/transactions.js#L119-L131

Let's not modify Dashboard here? We'll be converting it to the new SDK this week, so it can be updated alongside that.

jbowens:
> By my understanding, this should be backward compatible with older SDKs, since the old control_with_program technique would still work, right?

Yeah, just not newer SDKs with older coreds

jeffomatic:
> Yeah, just not newer SDKs with older coreds

Okay cool, this would probably end up as a minor version bump--you're supposed to use the same major.minor version for the core and SDKs.

erykwalder:
LGTM

Issue #439, opened 2017-01-26
Restore spent_output annotation to tx inputs


jeffomatic:
#421 removed the `spent_output` annotation for transaction inputs. We merged this change for the sake of moving forward with a big PR (nice work @oleganza!), but there are good reasons to retain that annotation--it provides easy access to transaction history (a feature of past use cases), and in any case, we should avoid the breaking change. I spoke with Oleg and he proposed filing this issue after landing the initial change.

jeffomatic:
Fixed in #440

PR #440, opened 2017-01-26
core/query: restore outpoint annotation in txinputs

oleganza:
This addresses https://github.com/chain/chain/issues/439

1. TxInputs have `spent_output{transaction_id,position}` struct side by side with `spent_output_id`.
2. Support for this struct is returned to Java and Ruby SDKs. NodeSDK is unchanged because it was implicitly forwarding JSON results as is to the user.
3. Annotation done by batch collection of a mapping `outputid -> outpoint` for all transactions in a block.
4. OutputID is not queried from DB, but instead computed from Outpoint. This adds extra hashing overhead per input, but reduces DB traffic by ≈50% (33 bytes vs 65 bytes per input). **Not sure if it's the right call, so please advise!**

PTAL @kr @jeffomatic @jbowens

### DIFF
--- a/core/query/index.go
+++ b/core/query/index.go
@@ -123,7 +123,8 @@ func (ind *Indexer) loadOutpoints(ctx context.Context, outputIDs pq.ByteaArray) 
 	`
 	results := make(map[bc.OutputID]bc.Outpoint)
 	err := pg.ForQueryRows(ctx, ind.db, q, outputIDs, func(txHash bc.Hash, outputIndex uint32) {
-		// We compute outid on the fly instead of receiving it from DB to save 40% of bandwidth.
+		// We compute outid on the fly instead of receiving it from DB to save 47% of bandwidth:
+		// DB is sending (hash256, int32) instead of (hash, int32, hash).
 		outid := bc.ComputeOutputID(txHash, outputIndex)
 		results[outid] = bc.Outpoint{
 			Hash:  txHash,
--- a/core/query/index.go
+++ b/core/query/index.go
@@ -115,17 +115,17 @@ func (ind *Indexer) insertAnnotatedTxs(ctx context.Context, b *bc.Block) ([]*Ann
 	return annotatedTxsDecoded, nil
 }
 
-func (ind *Indexer) loadOutpoints(ctx context.Context, outputIDs pq.ByteaArray) (map[bc.OutputID]*bc.Outpoint, error) {
+func (ind *Indexer) loadOutpoints(ctx context.Context, outputIDs pq.ByteaArray) (map[bc.OutputID]bc.Outpoint, error) {
 	const q = `
 		SELECT tx_hash, output_index
 		FROM annotated_outputs
 		WHERE output_id IN (SELECT unnest($1::bytea[]))
 	`
-	results := make(map[bc.OutputID]*bc.Outpoint)
+	results := make(map[bc.OutputID]bc.Outpoint)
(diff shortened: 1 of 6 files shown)


jbowens:
Oof. How sure are we that we want to keep this field? AFAIK, you can trace transaction history through the output IDs. It's also the only place in the annotated objects that uses nested json objects. It's inconsistent which makes it difficult to build alternative filter query implementations.

I know it would be a backwards incompatible change to remove it, but if there ever were a time to make such changes it's *before* production releases.

If we do need to keep it, I think we should do this in a batch lookup in the indexer. We'd need to do two separate passes, one to insert outputs and one to annotate inputs, but it'd avoid doing a query per TX per block.

jeffomatic:
@jbowens I would prefer to retain this annotation, since we are hoping to set the expectation that nothing will break with this release. However, you make valid points and I agree that it is a tricky situation, and I can keep the conversation going today. (At the very least, if we decide to retain the annotation, I will look into deprecating it.)

In the mean time, I think it makes sense to hedge our bets and prepare this PR as if we were going to land it. If there's an n+1 query issue, then let's try to get that out of the way. Also, if there's an A/B performance comparison, that would help inform the conversation.

jeffomatic:
Had some fruitful conversations with @kr and @devongundry on the matter, and we came to the following conclusion:

1. I think we have to bite the bullet and try to maintain backward-compatibility for the upcoming release, per our [versioning scheme](https://chain.com/docs/core/reference/versioning). It seems like there's an opportunity to use batching to mitigate the performance hit, and we can tolerate the remaining overhead, at least in the short term.
2. We'll deprecate `spent_output`, which will discourage current/future users from using it, and give us some headroom to streamline it out of the API over the next few months.

Hope that works!

oleganza:
The change to make a batch request for all inputs in a block should be pretty straightforward, but i'm not sure if we also want to introduce some async behavior or not which would be less straightforward.

Ah, also this PR in its current state does not restore accessors to `spent_output` in Ruby and Java. Will add those too.

jbowens:
👀

core/query/index.go
```diff
@@ -96,13 +108,34 @@ func (ind *Indexer) insertAnnotatedTxs(ctx context.Context, b *bc.Block) ([]*Ann
...
 	`
-	_, err := ind.db.Exec(ctx, insertQ, b.Height, positions, hashes, annotatedTxs)
+	_, err = ind.db.Exec(ctx, insertQ, b.Height, positions, hashes, annotatedTxs)
 	if err != nil {
 		return nil, errors.Wrap(err, "inserting annotated_txs to db")
 	}
 	return annotatedTxsDecoded, nil
 }
 
+func (ind *Indexer) loadOutpoints(ctx context.Context, outputIDs pq.ByteaArray) (map[bc.OutputID]*bc.Outpoint, error) {
```
jbowens: nit: since bc.Outpoint is not very large, maybe avoid the pointer? you can still check existence of an entry using

```go
outpoint, ok := outpoints[prevoutID]
if ok {
        // ...
}
```

oleganza: Actually, I had it as a value at first, but changed to check the absence. Thanks!

oleganza: Fixed and rebased: 8c1047348ed2fd9b7813c9bb9fd8ed89124b4650

jbowens:
lgtm. might need to update node sdk too?

oleganza:
@jbowens NodeSDK never wrapped that field in a function in the first place, so no changes are needed.

jeffomatic:
This is great, thank you @oleganza!

PR #441, opened 2017-01-26
docs/core/reference: change versioning rules

jeffomatic:
To simplify the release of new features and updates,, we are
relaxing the compatibility constraints so that only software
packages with identitcal major and minor versions are considered
mutually-compatible. This commit removes the "tolerance" language
from the versioning doc; compatibility is derived from strict
equality of major and minor versions.

This allows us, among other things, to introduce breaking changes
without undergoing a deprecation cycle. This not only includes
user-facing breaking changes, but also changes in the wire
protocol. For example, if the wire representation of a particular
API call changes from A to A', then the tolerance rule would force
us accept A or A' for a whole minor version.

This also resolves a subtle ambiguity in mutual compatability
between the client SDK and the server API. Even when applying a
standard deprecation process, we cannot guarantee both of the
following:

1) SDK 1.0 is compatible with server 1.1
2) SDK 1.1 is compatible with server 1.0

Using the wire protocol example above, even if server 1.1 supports
both the old (A) and new (A') API call, SDK 1.1 will have to
adopt A'. This will break compatibility between SDK 1.1 and server
1.0.

### DIFF
--- a/docs/core/reference/versioning.md
+++ b/docs/core/reference/versioning.md
@@ -15,7 +15,8 @@ Network versioning applies to the network API - the interface through which diff
 Signer versioning applies to the pieces of software that operate or interact with HSMs to facilitate transaction and block signing. These versions will change infrequently.
 
 ## Core Versioning
-The various packages of the Chain Core suite each use a three number versioning scheme - **x.y.z**. The first two numbers indicate compatibility between packages, with a tolerance of `+/-1`. For example, if you are a running a version of the Chain Core Mac app whose first two numbers are **1.1**, you can use a version of an SDK whose first two numbers are **1.0**, **1.1**, or **1.2**.
+
+The various packages of the Chain Core suite each use a three number versioning scheme - **x.y.z**. The first two numbers indicate compatibility between packages. For example, if you are a running a version of the Chain Core Mac app whose first two numbers are **1.1**, you can use a version of an SDK whose first two numbers are **1.1**.
 
 Sometimes we make bugfixes and minor feature updates to individual software packages. When this occurs, we update the third number in the version string for that package only. This number doesn't affect compatibility with other packages. For example, you can safely use version **1.1.1** of the Chain Core Mac App with version **1.1.3** of the Java SDK.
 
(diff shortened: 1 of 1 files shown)


jeffomatic:
@devongundry @ryandotsmith Would like to get blessings from both of you re: this change.

In fact, I would also like to go even further and keep the signer/HSM versioning the same as the other packages in Chain Core. As this document stands now, there is no way to reason about compatibility between the SDK and the signer (i.e., for signing transactions), because they exist in different version schemes.

devongundry:
I'm fine with this if @kr is

kr:
At first glance, this seems like it would make upgrading difficult. If I'm on version 1.1 and I want to upgrade to 1.2, what procedure should I follow? (Assuming, of course, that I don't want any downtime.)

jeffomatic:
@kr Great question! That could mean that requiring strict equality of versions is a non-starter. But then I'd like some other way of resolving some of the issues raised in this PR, namely:

1. Are we as a team committed to strict deprecate-then-break discipline? That would make upgrading possible, but it would impose some friction on our release process. Some more friction might actually be totally fine, but informal conversation suggests we want to preserve the right to break at any time. Our stated versioning promise should reflect whatever process we're committed to.
2. Assuming the answer to the above is "yes", then we still need to complicate the versioning language to account for the *directionality* of compatibility, such that newer versions of the SDK cannot be backward-compatible with older versions of the core.
3. Tangentially, I think signer versioning ought to be lumped into the versioning of the other Chain Core packages, since we need some way to account for compatibility between the SDK and the signer (which presumably shares the MockHSM interface).

kr:
As a point of reference, here's the language in beanstalkd's compatibility rules:

> As always, there will be no incompatible protocol changes until version 2.0. A client written for version 1.9 will work unmodified with any later 1.x release of beanstalkd.

https://kr.github.io/beanstalkd/2013/04/14/1.9-release-notes.html

Here's the Go 1 compatibility promise: https://golang.org/doc/go1compat

Go is a huge project, with a large surface area and a lot of code underneath; beanstalkd is a small project, with a small surface area and not much code (but also a very small team behind it, basically just me). Both projects have maintained compatibility according to their rules for years. Go 1 was released in 2012; beanstalkd 1 was released in 2008.

I'm sure we can do it too, especially if we only need to do it across a pair of "minor" versions at a time.

> Tangentially, I think signer versioning ought to be lumped into the versioning of the other Chain Core packages, since we need some way to account for compatibility between the SDK and the signer (which presumably shares the MockHSM interface).

This makes good sense to me.

jeffomatic:
Thanks @kr. My takeaway here is that we should double-down on backward-compatibility discipline. I'll close this issue for now, in favor of a future change that deals with the following:

- Re-categorizing the signer packages as part of the overall cored versioning scheme.
- Accounting for directionality of compatibility, e.g. new SDKs probably won't be compatible with older cores. What will help most here is hashing out some production upgrade scenarios that include all of our various components (core, SDK, signing server, HSM).

PR #442, opened 2017-01-26
docs/core/reference: clarify versioning language

jeffomatic:
This commit updates the official versioning language in two
significant ways:

- Core and signer versioning are no longer separate. Since core
  components interact directly with signer components, their
  interfaces should fall under the same versioning umbrella.
- Pre-existing language around version "tolerance" has been
  simplified to the guarantee that server packages are backward
  compatible with clients that are at most one minor version
  behind. This provides for a reasonable upgrade flow between
  the Chain Core server and client apps.

There are also minor language tweaks, mostly for brevity and
clarity.

### DIFF
--- a/docs/core/reference/versioning.md
+++ b/docs/core/reference/versioning.md
@@ -6,29 +6,39 @@ There are three distinct Chain versioning schemes:
 
 * **core** versioning
 * **network** versioning
-* **signer** versioning
 
-Core versioning applies to the main pieces of software that comprise, package, or directly interact with Chain Core (excluding those that facilitate transaction and block signing). For example, the Chain Core Mac app and the Java SDK.
+Core versioning applies to the main pieces of software that compose, package, or directly interact with Chain Core. This includes, for example, the Chain Core server, the Java SDK, and the Chain Core Mac app.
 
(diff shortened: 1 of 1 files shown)


jeffomatic:
@kr @devongundry @ryandotsmith Okay, I gave it another shot, PTAL

kr:
LGTM

It's interesting, because there are three relationships that affect the upgrade procedure: the SDK is a client of cored, the SDK is a client of signerd, and cored is a client of signerd. So to upgrade from 1.0 to 1.1, you must first update all signerds, second update all coreds, then third update all SDK clients. I don't know if we want to go over that explicitly in this document, or elsewhere, or just informally with customers. WDYT?

Another thought, for down the line when we're writing release notes. Repeating a succinct version of this in the release notes might make it easier to understand because it's concrete. Instead of a "for example" it'll be the real version strings. So release notes for an SDK might say "This SDK (version 1.1.3) can be used with any Chain Core server version 1.1.x or 1.2.x." or words to that effect.

jeffomatic:
@kr: Totally. I also imagine a hairy situation re: signerd, which acts as a "server" to cored and the SDK, but also as a "client" of the HSM firmware. My gut right now says that it's more confusing than useful to spell out the gnarly edge cases in a general doc such as this. So long as we feel like this scheme isn't internally inconsistent, I think this level of specificity is ok.

I very much like your idea of repeating the compatibility implications as much as possible (even ad nauseum, for safety's sake), and in the spirit of that, we may also want to spell out precise upgrade steps, especially if it involves more than just cored and an SDK.

PR #443, opened 2017-01-27
core/query/filter: unit test the scanner

jbowens:
Add unit tests for the filter scanner. Increase test coverage before
we add unicode support.


jbowens:
PTAL

kr:
LGTM

PR #444, opened 2017-01-27
dashboard: add integration tests

dominic:


dashboard/package.json
```diff
@@ -8,17 +8,20 @@
...
   "author": "Chain",
   "license": "",
   "dependencies": {
     "babel-polyfill": "^6.16.0",
     "bootstrap-sass": "~3.3.7",
+    "chain-sdk": "^1.0.2",
     "fetch-ponyfill": "^3.0.2",
     "lodash.pick": "^4.4.0",
     "lodash.uniq": "^4.5.0",
+    "lodash.snakecase": "^4.1.1",
```
jeffomatic: Maybe not for this PR, but at this point we should just get all of lodash.

jeffomatic:
As discussed in person, there might be an opportunity to us a helper function that merges CSS classes together with the enclosing component's `this.constructor.name`.

dashboard/src/utility/dashboardClasses.js
```diff
@@ -0,0 +1,9 @@
+import classNames from 'classnames'
+import snakeCase from 'lodash/snakecase'
+
+const dashboardClasses = (owner, ...args) => {
```
jeffomatic: I think this all pretty much looks good, but some aesthetic/bikesheddy things to consider briefly, before we land:

1. `dashboardClasses` is a bit ambiguous as a name. Strawman: `componentClassNames`, which implies that you should use it for the `classNames` prop for components. In any case, a descriptive comment for the function would also help legibility of that pattern.
2. There are a few things we could do to the component name: snake_case them, leave them alone, prefix them, or some combination of those. snake_case feels a bit neither-here-nor-there to me, because it is a Rails-idiomatic transformation of the original name; also, CSS class names are traditionally kebab-case. If we wanted to be super verbose and explicit, we could do something like `component-OriginalComponentName`. Here, the original component class is preserved without transformation, but there is a prefix, to prevent ambiguous collisions. Anyway, YMMV.
3. I don't think `owner` should ever be allowed to be null. In fact, we could even consider adding a pseudo-static check to make sure that `owner` derives from `React.Component`.

dominic: 1. :+1:
2. :+1: to `component-OriginalName`; that increases searchability a lot
3. :+1:

dashboard/src/utility/componentClassNames.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import classNames from 'classnames'
+
+const componentClassNames = (owner, ...args) => {
+  if (!React.Component.prototype.isPrototypeOf(owner))
+    throw new Error('Component must descend from React.Component')
+
+  const coreName = 'component-' + owner.constructor.name
```
jeffomatic: Do the corresponding tests need to be updated too?

dashboard/src/utility/componentClassNames.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import classNames from 'classnames'
+
+const componentClassNames = (owner, ...args) => {
+  if (!React.Component.prototype.isPrototypeOf(owner))
```
jeffomatic: Super nit: I would sleep better if we put braces around all multi-line if-blocks.

dominic: Getting a good night's sleep is really important.

dashboard/src/utility/componentClassNames.js
dominic: whoop.

PR #445, opened 2017-01-27
core/query/filter: add unicode support

jbowens:
Support UTF-8 characters in strings and identifiers. This does _not_ add
escaping for strings or identifiers so `'` and `\` are still illegal
characters in strings. Any non-alphanumeric characters are still
illegal in identifiers.

Token positions are reported using the byte offset of the beginning of
the token. To retrieve the code point offset, the consumer of the package
can use `utf8.RuneCount(src[:offset])`.

### DIFF
--- a/core/query/filter/parser_test.go
+++ b/core/query/filter/parser_test.go
@@ -19,6 +19,10 @@ func TestParseValid(t *testing.T) {
 			expr: valueExpr{typ: tokString, value: "'hello world'"},
 		},
 		{
+			p:    "'สวัสดีชาวโลก'",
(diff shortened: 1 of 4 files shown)


kr:
LGTM

core/query/filter/parser_test.go
```diff
@@ -19,6 +19,10 @@ func TestParseValid(t *testing.T) {
 			expr: valueExpr{typ: tokString, value: "'hello world'"},
 		},
 		{
+			p:    "'สวัสดีชาวโลก'",
```
bobg: Nice.

PR #447, opened 2017-01-27
docs: update deploy instructions

jeffomatic:
We're noticing that the `main` branch is not always the appropriate source for our online documentation, because it often contains updates to the source docs that reflect changes that haven't been released yet. Therefore, this commit proposes a new `docs-release` branch that is prepared ad hoc before documentation is uploaded to our public website.

### DIFF
--- a/bin/upload-docs
+++ b/bin/upload-docs
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 set -ef -o pipefail
 
-# TODO: when production is viable target, add commandline option to upload to
-# the production bucket.
 bucket="chain-staging.chain.com"
 if [ "$1" = "prod" ];
 then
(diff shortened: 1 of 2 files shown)


jeffomatic:
PTAL @dominic

dominic:
Thoughts on it just being `release` instead of `docs-release`? It would be really nice to have an entire repo state that is safe to checkout. I've seen other projects name this branch `stable` to indicate a safe clone point (versus debugging users who just happen to get the HEAD of main at any given day).

jeffomatic:
We could try that, but the reason I didn't make this more general is that I just haven't thought through the implications of something so widely scoped as "release". As in, I'm not sure what that means outside of the docs alone.

With `docs-release`, the question of what to put in that branch is pretty clear: what are the most recent official verisons of the docs, reflecting the publically-downloadable `cored` and SDKs? What would be the equivalent definition for `release`? Do we have a concept of a stable source build that is not simply the same thing as the last `cmd-cored-x.y.z` tag?

dominic:
LGTM

PR #448, opened 2017-01-27
protocol/bc: do away with witness hashes

bobg:
The merkle tree is now computed from transaction hashes. This is preparatory to landing the code in https://github.com/chain/chain/tree/txgraph


bobg:
PTAL

kr:
:eyes:

kr:
LGTM!

PR #449, opened 2017-01-27
core: disable some endpoints in production

jbowens:
Disable the reset and mock HSM endpoints in production. The reset code
won't even be included in the binary in production, but the API endpoint
should also be disabled.


jbowens:
PTALooksie

tessr:
I just found your request for a "looksie" so charming, you know

PR #450, opened 2017-01-28
protocol/bc: add marshaling function to BlockHeader

erykwalder:
This enables sending and receiving block headers in RPC requests.


kr:
LGTM

PR #453, opened 2017-01-31
core/config: add block hsm settings

erykwalder:
This allows the core to configure the remote HSM to be used for block
signing.

### DIFF
--- a/core/schema.sql
+++ b/core/schema.sql
@@ -7,7 +7,6 @@
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
-SET idle_in_transaction_session_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SET check_function_bodies = false;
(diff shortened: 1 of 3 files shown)


core/schema.sql
```diff
@@ -7,6 +7,7 @@
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
```
jbowens: this is a PostgreSQL 9.6 config. we might want to remove it if other people are still using 9.5. or we could just require everyone to use 9.6 now?

jbowens:
lgtm

core/schema.sql
tessr: updating postgres will require re-bundling the windows installer and (i think) mac app.

erykwalder:
updated

jbowens:
sry meant to approve before but forgot to click the button

PR #454, opened 2017-01-31
protocol/vm: simplify vm interface

bobg:
Instead of returning an `ok` bool and an error, fold the `!ok` case into the error by returning the new value `ErrFalseVMResult`.


bobg:
PTAL

bobg:
!PTAL, one adjustment coming

bobg:
PTAL

erykwalder:
LGTM

PR #455, opened 2017-01-31
core/query: drop outpoint index

jbowens:
The PostgreSQL index on `annotated_outputs.(tx_hash, outpoint_index)` existed for a single query that marked consumed outputs as spent. Ever since #421, we've used a query against the output ID for that. There's already an index on output ID.


erykwalder:
LGTM

PR #458, opened 2017-02-02
protocol/tx: add extensible tx hashing

bobg:
This new package implements a change to transaction hashing that is more flexible and futureproof. Transactions are converted to new data structures that are intended eventually to replace the types in `protocol/bc`. This transitional code is intended to protect against hash values changing when we reach that point.

The spec for these changes is in `docs/protocol/specifications/txgraph-draft.md`.

Still to come (in follow-on PRs): the code that integrates this new package with the existing code in `protocol/bc`; and integration of `txgraph-draft.md` into the spec proper.

PR #459, opened 2017-02-02
protocol/bc: replace hashing with the protocol/tx implementation

bobg:
This is a follow-on to https://github.com/chain/chain/pull/458 that replaces the implementation of transaction hashing in `protocol/bc` with the mechanism in `protocol/tx`.

This has required replacing `bc.TxData` objects in some contexts with `bc.Tx` objects.

There are also several places in which skeletal test data had to be more fully fleshed out to pass muster under the new hashing rules.


bobg:
Note: after https://github.com/chain/chain/pull/458 lands, I'll adjust this PR to merge onto `main` rather than onto `txgraph`.

protocol/tx/transaction.go
```diff
@@ -2,15 +2,20 @@ package tx
...
 import (
 	"chain/crypto/sha3pool"
+	"chain/errors"
 	"chain/protocol/bc"
 	"fmt"
 )
 
+func init() {
+	bc.TxHashesFunc = TxHashes
+}
```
jbowens: not necessarily now, but would it possible be able to invert the dependency and make `protocol/bc` depend on `protocol/tx`? This init feels a little fragile (and requires all of those `init` funcs in our tests). We'd need to export more of the `tx` package's types and move the `bc.Tx`-to-graph conversion into `bc` but neither seems bad?

tessr:
womp womp, git land closed this, and now I can't reopen it

PR #461, opened 2017-02-03
sdk/java: update pom.xml for Maven deploys

jeffomatic:
This commit adds all necessary tags for releasing the Java SDK
to Maven.

To deploy to Maven, ensure your $HOME/.m2/settings.xml contains
the tags described here:

https://gist.github.com/jeffomatic/e6b3c23799e9586698ca466e2765916b

Then run:

```
cd sdk/java
mvn clean deploy -DskipTests=true
```


jeffomatic:
PTAL

ryandotsmith:
When we get around to describing our SDK deployment methods in our README, we should add the commands in this PR message.

PR #463, opened 2017-02-03
protocol/bc: replace hashing with the protocol/tx implementation

tessr:
This supersedes #459, with a rebased version of the txgraph-integ branch. (I cut a new branch because I was afraid of messing up the rebase, which, I did.)


tessr:
Note that this does not address Jackson's concern from #459: https://github.com/chain/chain/pull/459#discussion_r99254552

core/account/builder.go
```diff
@@ -123,7 +123,11 @@ func (a *spendUTXOAction) Build(ctx context.Context, b *txbuilder.TemplateBuilde
 		outid = *a.OutputID
 	} else if a.TxHash != nil && a.TxOut != nil {
 		// This is compatibility layer - legacy apps can spend outputs via the raw <txid:index> pair.
-		outid = bc.ComputeOutputID(*a.TxHash, *a.TxOut)
+		q := `SELECT output_id FROM account_utxos WHERE tx_hash=$1 AND index=$2`
+		err := a.accounts.utxoDB.db.QueryRow(ctx, q, *a.TxHash, *a.TxOut).Scan(&outid)
+		if err != nil {
+			return err
```
jbowens: If the lookup fails to find a matching output, we might want to return `pg.ErrUserInputNotFound` like we do within the reserver: https://github.com/chain/chain/blob/main/core/account/reserve.go#L424

o/w, I think this would end up as a 500?

core/migrate/data.go
```diff
@@ -97,9 +97,9 @@ var migrations = []migration{
 		ALTER TABLE config ALTER COLUMN blockchain_id SET DATA TYPE bytea USING decode(blockchain_id, 'hex');
 	`},
 	{Name: "2017-01-13.0.core.asset-definition-bytea.sql", SQL: `
-		ALTER TABLE assets 
+		ALTER TABLE assets
 			ALTER COLUMN definition SET DATA TYPE text;
-		ALTER TABLE assets 
+		ALTER TABLE assets
```
jbowens: I don't think we can modify and of these older migrations. It'll change the SHA. If we really don't care about preserving data since it requires a blockchain reset anyways, we should just collapse all the migrations into a new single migration. We can do that in this PR or a new one.

core/migrate/data.go
```diff
@@ -117,13 +117,16 @@ var migrations = []migration{
 	{Name: "2017-01-25.0.account.cp-expiry.sql", SQL: `
 		ALTER TABLE account_control_programs ADD COLUMN expires_at timestamp with time zone;
 	`},
-	{Name: "2017-01-30.0.txdb.snapshots-timestamp.sql", SQL: `
+	{Name: "2017-01-30.1.txdb.snapshots-timestamp.sql", SQL: `
 		ALTER TABLE snapshots ADD COLUMN created_at timestamp without time zone DEFAULT now();
 	`},
-	{Name: "2017-01-30.1.core.add-block-hsm-config.sql", SQL: `
+	{Name: "2017-01-30.2.core.add-block-hsm-config.sql", SQL: `
```
jbowens: These filenames also shouldn't change I think

core/query/annotated.go
```diff
@@ -182,9 +182,15 @@ func buildAnnotatedInput(orig *bc.TxInput, outpoints map[bc.OutputID]bc.Outpoint
 	return in
 }
 
-func buildAnnotatedOutput(orig *bc.TxOutput, idx uint32, txhash bc.Hash) *AnnotatedOutput {
+func buildAnnotatedOutput(tx *bc.Tx, idx uint32) *AnnotatedOutput {
+	orig := tx.Outputs[idx]
+	referenceData := json.RawMessage(orig.ReferenceData)
+	if len(referenceData) == 0 {
+		referenceData = []byte(`{}`)
+	}
```
jbowens: I think this is dead code from the rebase. `referenceData` isn't used.

tessr: Good catch, thanks.

core/migrate/data.go
tessr: Blerp. 

I think I'm going to remove the schema changes from this PR altogether and open a new one with the very simple db change we're actually interested in below.

core/query/index.go
```diff
@@ -89,6 +89,8 @@ func (ind *Indexer) insertAnnotatedTxs(ctx context.Context, b *bc.Block) ([]*Ann
 
 	// Build the fully annotated transactions.
 	for pos, tx := range b.Transactions {
+		hashes = append(hashes, tx.ID[:])
+		positions = append(positions, uint32(pos))
```
jbowens: I think these are also an artifact of the rebase. `hashes` and `positions` are populated down below in the `for pos, tx := range annotatedTxs {` loop

core/migrate/data.go
tessr: (resolved offline, but for those following along at home.... we'll preserve this for now but plan on collapsing into a single migration before we do the next blockchain reset.)

jbowens:
Nailed it

PR #464, opened 2017-02-03
core/txdb: store snapshot as a set of hashes

jbowens:
The snapshot's state tree is now just a set of hashes instead of a kv
pair. When serializing the state tree as a protobuf, there's no longer a
need to store key and value separately.

Future work:
  * Update `protocol/patricia` to not duplicate key and value
    in-memory.


jbowens:
Nevermind, I think this is a little more subtle than I thought, and we shouldn't try to sneak it into this release. I forgot that we hash within the patricia package, so if we insert the same value as both the key and the value, the value will change when it's hashed within the patricia package.

PR #465, opened 2017-02-03
core/migrate: drop unspent ids from account_utxos

tessr:
This detangles the schema updates from #463. (It is also blocked by #463.) 

It also includes a heck of a hot of schema dumping noise because we're flipping between dumping with postgres 9.5 and 9.6


jbowens:
should we merge this separately or within #463?

bobg:
Cherry-picked into #463.

PR #468, opened 2017-02-03
protocol/tx: replace a call to io.CopyN with a fixed-length read and write

bobg:
```
BenchmarkHashEmptyTx-8        13705         5998          -56.23%
BenchmarkHashNonemptyTx-8     56714         28370         -49.98%
```


bobg:
PTAL

jbowens:
lgtm

PR #469, opened 2017-02-03
protocol/tx: set valueSource position to 0 when the source is not a mux entry

bobg:
Per discussion with @danrobinson.


bobg:
PTAL

PR #470, opened 2017-02-03
core/txdb: store state tree as a set

jbowens:
The snapshot's state tree is now just a set of hashes instead of a kv
pair. When serializing the state tree as a protobuf, there's no longer a
need to store key and value separately.

Future work:
  * Update `protocol/patricia` interface to not expose as a kv pair.


jbowens:
This isn't vital for 1.1rc because we'll just ignore the `hashes` protocol buffer field during decoding.

erykwalder:
LGTM

PR #471, opened 2017-02-04
docs/core/get-started: clean up language on SDK page

jeffomatic:
- Add Java Gradle instructions
- Move version compatibility language to the top of the doc
- Add table of contents
- Simplify verbiage

### DIFF
--- a/docs/core/get-started/sdk.md
+++ b/docs/core/get-started/sdk.md
@@ -1,10 +1,18 @@
 # Download Chain Core SDKs
 
-Chain Core SDKs enable your appplication to communicate with Chain Core.
+Client libraries for Chain Core are available for the following platforms:
+
+- [Java](#java)
+- [Node.js](#node-js)
+- [Ruby](#ruby)
+
+To ensure compatibility between your application and Chain Core, choose an SDK version whose major and minor components (`major.minor.x`) match your version of Chain Core.
 
 ## Java
 
-The Java SDK is available via Maven, on Sonatype's [Central Repository](http://central.sonatype.org/). Simply add the following to your `pom.xml`:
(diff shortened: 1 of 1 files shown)


jeffomatic:
PTAL

docs/core/get-started/sdk.md
```diff
@@ -16,24 +24,30 @@ The Java SDK is available via Maven, on Sonatype's [Central Repository](http://c
...
-You can also [download the JAR](../../java/chain-sdk-latest.jar) as a binary.
+**Gradle** users should add the following to `build.gradle`:
 
-## Ruby
+```
+compile 'com.chain:chain-sdk-java:1.0.1'
+```
 
-The Ruby SDK is available [via Rubygems](https://rubygems.org/gems/chain-sdk). Make sure to use the most recent version whose major and minor components (`major.minor.x`) match your version of Chain Core. Ruby 2.1 or greater is required.
+You can also [download the JAR](https://search.maven.org/remotecontent?filepath=com/chain/chain-sdk-java/1.0.1/chain-sdk-java-1.0.1.jar) as a binary.
```
dominic: Do we need this? I'm not sure if Java developers think of Maven as a download host.

jeffomatic: I'm in favor of keeping it because there are workflows outside of Maven/Gradle that use the jarfile directly. The fact that the jar is hosted on Maven is incidental. It also moves away from downloading the jarfile from the local core, which means we won't have to bundle it into the docs anymore.

FWIW, I'm kinda just copying the "Download" section of this: http://square.github.io/okhttp/

Lemme know if that addresses your concern.

dominic: Yup! Copy best practices errday.

PR #472, opened 2017-02-04
core/query: use SQL table to typecheck thoroughly

jbowens:
Use the SQL table definitions to typecheck filter predicates more
exhaustively. Also, when possible, infer the type of arbitrary
json values so that we can cast them to the correct PostgreSQL
value.

For example, if you run the filter:
```
asset_definition.priority = 10
```

The type checker will infer that `asset_definition.priority` is an integer
and the resulting SQL will be:
```sql
(tbl."asset_definition"->>"priority")::bigint = 10::bigint
```

This also allows for queries to access boolean values in reference data.
For example, this is now a valid query:
```
asset_definition.is_dollar
```
Booleans that are exposed in the JSON as "yes"/"no" still need to be
compared as strings until we have a backwards compatible way of
supporting both syntaxes for those fields.

### DIFF
--- a/core/query/filter/typecheck_test.go
+++ b/core/query/filter/typecheck_test.go
@@ -1,6 +1,9 @@
 package filter
 
-import "testing"
+import (
+	"chain/testutil"
+	"testing"
+)
 
 func TestTypeCheckInvalid(t *testing.T) {
 	testCases := []struct {
@@ -16,6 +19,8 @@ func TestTypeCheckInvalid(t *testing.T) {
 		{p: `wat(asset_id = 'a')`},
 		{p: `position(asset_id = 'a')`},
 		{p: `position.huh`},
+		{p: `ref.something = 'abc' OR ref.something = 123`},
+		{p: `ref.buyer.id = 'abc' OR ref.buyer = 'hello'`},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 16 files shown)


erykwalder:
:eyes:

core/query/filter/typecheck_test.go
```diff
@@ -1,15 +1,31 @@
...
+		{p: `1 = 'hello world'`, err: errors.New("= expects operands of matching types")},
+		{p: `inputs('hello')`, err: errors.New("inputs(...) body must have type bool")},
+		{p: `inputs(1=1).bar`, err: errors.New("selector `.` can only be used on objects")},
+		{p: `'hello'.foo`, err: errors.New("selector `.` can only be used on objects")},
+		{p: `inputs(amount = asset_id)`, err: errors.New("= expects operands of matching types")},
+		{p: `inputs(100 = asset_id)`, err: errors.New("= expects operands of matching types")},
+		{p: `inputs(wat)`, err: errors.New("invalid attribute: wat")},
+		{p: `wat(asset_id = 'a')`, err: errors.New("invalid environment `wat`")},
+		{p: `position(asset_id = 'a')`, err: errors.New("invalid environment `position`")},
+		{p: `('a' = 'a') = (1 = 1)`, err: errors.New("= expects integer or string operands")},
```
erykwalder: I know this check wasn't just added, but is that what we want? What if you want to see if two bool values are equal?

erykwalder:
LGTM, my only feedback was from something not in this PR

core/query/filter/typecheck_test.go
jbowens: Yeah, I think we will want to relax this soon, but it's tricky at the moment because the booleans that we expose on our own objects are exposed as strings. I think we should figure out what to do with those fields before relaxing it.

PR #474, opened 2017-02-04
bin: remove Java SDK jar from bundle-docs flow

jeffomatic:
Since the Java SDK is now hosted on Maven, there is no need to
include the release in Chain Core itself. This is a nice reduction
in the size of the Chain Core binary.


jeffomatic:
PTAL

PR #475, opened 2017-02-04
protocol/tx: return more specific types from newData and newSpend

bobg:
Also remove a stale TODO.

### DIFF
--- a/protocol/tx/data.go
+++ b/protocol/tx/data.go
@@ -29,15 +29,13 @@ func (d *data) Body() interface{} { return d.body }
 
 func (data) Ordinal() int { return -1 }
 
-func newData(hash bc.Hash) entry {
+func newData(hash bc.Hash) *data {
 	d := new(data)
 	d.body = hash
 	return d
 }
 
 func hashData(data []byte) (h bc.Hash) {
-	// TODO: Do we want domain separation here?  (E.g., a "data:"
-	// prefix.) If so, both the code and the spec need updating.
 	sha3pool.Sum256(h[:], data)
 	return
 }
--- a/protocol/tx/spend.go
+++ b/protocol/tx/spend.go
@@ -14,7 +14,7 @@ func (s *spend) Body() interface{} { return s.body }
 
 func (s spend) Ordinal() int { return s.ordinal }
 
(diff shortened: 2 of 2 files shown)


bobg:
PTAL
I am smol

protocol/tx/data.go
```diff
@@ -29,15 +29,13 @@ func (d *data) Body() interface{} { return d.body }
...
-func newData(hash bc.Hash) entry {
+func newData(hash bc.Hash) *data {
 	d := new(data)
 	d.body = hash
 	return d
 }
 
 func hashData(data []byte) (h bc.Hash) {
-	// TODO: Do we want domain separation here?  (E.g., a "data:"
-	// prefix.) If so, both the code and the spec need updating.
```
kr: Did we make a call on this? IMO we should do domain separation as a matter of habit, unless there's a really clear reason not to. cc @tonychain

tonychain: I'm definitely a fan of always doing domain separation. The overhead it adds is negligible and it's good security hygiene

jbowens: @kr — I think this was the discussion around it:
https://chainhq.slack.com/archives/txgraph/p1486079399000187

bobg:
PTAL

protocol/tx/data.go
kr: That the data entry itself serves as a domain separator seems plausible to me. I'll defer to @tonychain and @oleganza.

jbowens:
lgtm

protocol/tx/data.go
oleganza: Data entry _is_ the domain separator for the raw data. This prevents having the raw hash of the data to be reused in the txgraph structure. For the purposes of identifying the raw data using `OP_REFDATAHASH` we need a raw hash that represents the raw data.

tonychain: To back up a bit, the threat would be an attacker convincing you to sign a message in one context, and then reusing it in another.

So long as that threat is addressed, I'm happy. Sounds like it is?

oleganza: Sounds like it is, yup. Thanks for clarifying the threat model, @tonychain!

PR #476, opened 2017-02-06
core/txfeed: move txfeed query from core/query

jbowens:
Move the list txfeeds query logic from the `core/query` package to
the `core/txfeed` package. The txfeeds query doesn't use filters, so it
probably shouldn't be in that package. Also, the `core/query` dependency
on `core/txfeed` makes it difficult to refactor the `core/query` and
`core/query/filter` packages without introducing a circular dependency.


jbowens:
PTAL, need this to prevent a cyclic dependency when refactoring in #472.

erykwalder:
LGTM

PR #478, opened 2017-02-07
dashboard: use SDK for object creation

dominic:
Adds integration testing frameworks for object lists and creation


dashboard/package.json
```diff
@@ -48,6 +48,7 @@
     "babel-preset-es2015": "~6.9.0",
     "babel-preset-react": "~6.5.0",
     "babel-preset-react-hmre": "~1.1.1",
+    "babel-register": "^6.22.0",
```
jeffomatic: Would be good to have a good default re: whether to use `^` or `~`. My $0.02 would be to be as conservative as possible...if we did strict equality I would be okay with that, but it looks like `~` is the most common thing.

dashboard/src/components/Common/TextField/TextField.jsx
```diff
@@ -30,6 +30,7 @@ class TextField extends React.Component {
         {this.props.title && <FieldLabel className={styles.title}>{this.props.title}</FieldLabel>}
         <input className='form-control'
           type={this.state.type}
+          name={this.props.fieldProps.name}
```
jeffomatic: Is this necessary with the `{...fieldProps}` at the end of this component instance?

dashboard/src/utility/environment.js
```diff
@@ -1,6 +1,7 @@
 /* global process */
 
 import chain from '_chain'
+import chainSDK from 'chain-sdk'
```
jeffomatic: By convention: `chainSdk`

dashboard/package.json
```diff
@@ -8,8 +8,8 @@
     "build": "NODE_ENV=production webpack --config src/config/webpack.app.js",
     "lint": "eslint \"src/**\"",
     "start": "hjs-dev-server src/config/webpack.app.js",
-    "test": "wdio test/conf/wdio.local.js",
-    "test-browserstack": "wdio test/conf/wdio.browserstack.js",
+    "test": "wdio test/conf/wdio.local.js --suite base",
+    "testExtended": "EXTENDED=1 wdio test/conf/wdio.local.js",
```
jeffomatic: `--suite extended` would probably simplify running the extended tests a little.

dashboard/test/specs/keys.js
```diff
@@ -0,0 +1,49 @@
...
+    before(() => {
+      signer = new chain.HsmSigner()
+
+      return expect(ensureConfigured()).to.be.fulfilled
+        .then(() => expect(setUpObjects(signer)).to.be.fulfilled)
+        .then(() => browser.url('/mockhsms'))
+    })
+
+    it('does not display a welcome message', () => {
+      browser.isExisting('.component-EmptyList').should.equal(false)
```
jeffomatic: *ducking for cover*

Eyeballing it right now, the `component-` prefix sure is verbose, isn't it? I'm down to get rid of it if you are. UpperCamelCase seems obvious enough as an indicator of a component, and there shouldn't be collisions unless we adopt a CSS framework that uses that casing style.

dashboard/package.json
dominic: This actually has two side effects. Running all tests (aka "extended") by not specifying a suite, and configuring the tests to [_not_ run in parallel for the extended suite](https://github.com/chain/chain/pull/478/files#diff-1019fea1af46f77a3038852a0e2a0665R51).

Since the extended tests involve empty states and core resets, parallelizing them with normal tests can result in random failures when a single core is being subjected to multiple tests.

dashboard/test/helpers.js
```diff
@@ -0,0 +1,86 @@
+const chain = require('chain-sdk')
+const client = new chain.Client()
+
+global.sleep = (ms) => {
```
dominic: This isn't about promise fulfillment so much as giving `cored` time to start up; the promises will resolve or reject rather quickly as is. Rather, I want calls _after_ `client.config.configure({})` to wait a second before being attempted.

Something like https://github.com/magicdawn/promise.sleep matches this case better. How would you feel about a promise that resolves with setTimeout instead?

```js
const sleep = (ms) => new Promise(resolve => {
  setTimeout(() => resolve(), ms)
}) 
```

dashboard/src/components/Common/TextField/TextField.jsx
dominic: Ah, not if I include it the `TEXT_FIELD_PROPS` whitelist. Good catch.

dashboard/test/specs/keys.js
dominic: Yeah, I can't come up with a strong argument to keep it around.

dashboard/package.json
jeffomatic: Maybe I'm not understanding something here. I assumed that the following will run only the tests in the `extended` suite, which are also the tests that require `EXTENDED=1`:

```
EXTENDED=1 wdio test/conf/wdio.local.js --suite extended
```

dashboard/test/helpers.js
jeffomatic: Yeah that's great.

dashboard/package.json
dominic: I see your point. We don't need to define the `extended` suite at this point - I want `npm run testExtended` to run _everything_ it can get its hands on, in a way that passes safely.

PR #481, opened 2017-02-07
protocol/tx: Dereference the data entry when initializing a vmcontext refdata hash

bobg:


bobg:
PTAL

bobg:
PTAL

bobg:
Here is a :cookie: PTAL

protocol/tx/transaction.go
```diff
@@ -49,6 +49,18 @@ func TxHashes(oldTx *bc.TxData) (hashes *bc.TxHashes, err error) {
...
 	hashes.VMContexts = make([]*bc.VMContext, len(oldTx.Inputs))
 
+	getRefDataHash := func(id entryRef) bc.Hash {
+		dEntry, ok := entries[id]
+		if !ok {
+			return bc.EmptyStringHash
+		}
+		d, ok := dEntry.(*data)
+		if !ok {
+			return bc.EmptyStringHash
```
jbowens: this shouldn't be an error?

bobg: I suppose so, thanks. Fixed.

jbowens:
lgtm. it might be worth trying it without the `getRefDataHash` closure to avoid extra indirection? it might even be the same number of lines of code.

PR #484, opened 2017-02-07
cmd/cored: add signerd client

ameets:


cmd/cored/main.go
```diff
@@ -233,7 +234,20 @@ func launchConfiguredCore(ctx context.Context, db *sql.DB, conf *config.Config,
 		if err != nil {
 			chainlog.Fatal(ctx, chainlog.KeyError, err)
 		}
-		s := blocksigner.New(blockPub, mockHSM, db, c) // TODO(jackson): support real HSM
+
+		var hsm blocksigner.Signer = mockHSM
+		if conf.BlockHSMURL != "" {
+			hsm = &remoteHSM{Client: &rpc.Client{
+				BaseURL:      conf.BlockHSMURL,
+				AccessToken:  conf.BlockHSMAccessToken,
```
erykwalder: signerd only uses the password portion of the auth. seems like there are two options:

1) take a "access token" string of "username:password" when configuring (in this case ":password") and don't worry about modifying it
2) take a password when configuring, and convert to an access token string here

erykwalder: looking for outside opinions here

cmd/cored/main.go
```diff
@@ -381,6 +395,20 @@ type remoteSigner struct {
 	Key    ed25519.PublicKey
 }
 
+// remoteHSM is a client wrapper for an hsm that is used as a blocksigner.signer
```
kr: typo, `signer` should be capitalized

cmd/corectl/main.go
```diff
@@ -79,7 +79,10 @@ func configGenerator(db *sql.DB, args []string) {
 
 	var flags flag.FlagSet
 	maxIssuanceWindow := flags.Duration("w", 24*time.Hour, "the maximum issuance window `duration` for this generator")
-	isSigner := flags.Bool("s", false, "whether this core is a signer")
+	flagK := flags.String("k", "", "local `pubkey` for signing blocks")
```
bobg: Don't think backticks belong in these help strings.

cmd/corectl/main.go
```diff
@@ -79,7 +79,10 @@ func configGenerator(db *sql.DB, args []string) {
...
+	flagHSMURL := flags.String("hsm-url", "", "hsm `url` for signing blocks (mockhsm if empty)")
```
bobg: How about just `-url` and `-token`?

(Here and in `createToken`.)

cmd/corectl/main.go
```diff
@@ -88,8 +91,19 @@ func configGenerator(db *sql.DB, args []string) {
 	flags.Parse(args)
 	args = flags.Args()
 
+	//not a blocksigner
+	if *flagK == "" && *flagHSMURL != "" {
+		fatalln("error: flag -hsm-url has no effect without -k")
+	}
+
+	//hsm url and access token are
+	//ToDo (ameets): update when switching to x.509 authorization
```
bobg: Please use the format `TODO(ameets): ...` throughout.

cmd/corectl/main.go
ameets: I think it's confusing to just use `-url` is because there are two url arguments: the hsm url (as a flag) and a pubkey url (as an arg) `"usage: corectl config-generator [flags] [quorum] [pubkey url]..."`

Open to suggestions

cmd/corectl/main.go
ameets: I copied the help string format from configNongenerator. Not sure about the previous history - do we want to change the format throughout?

ameets:
PTAL

cmd/corectl/main.go
```diff
@@ -120,12 +133,15 @@ func configGenerator(db *sql.DB, args []string) {
 
 	conf := &config.Config{
 		IsGenerator: true,
-		IsSigner:    *isSigner,
 		Quorum:      quorum,
 		Signers:     signers,
 		MaxIssuanceWindow: chainjson.Duration{
 			Duration: *maxIssuanceWindow,
 		},
+		IsSigner:            *flagK != "",
```
erykwalder: Just to double check, previously if `IsSigner` was passed true without `BlockPub`, then `config.Configure` would generate a block key. Are we fine losing that option?

cmd/cored/main.go
```diff
@@ -233,7 +234,21 @@ func launchConfiguredCore(ctx context.Context, db *sql.DB, conf *config.Config,
 		if err != nil {
 			chainlog.Fatal(ctx, chainlog.KeyError, err)
 		}
-		s := blocksigner.New(blockPub, mockHSM, db, c) // TODO(jackson): support real HSM
+
+		var hsm blocksigner.Signer = mockHSM
+		if conf.BlockHSMURL != "" {
+			//TODO(ameets): these members may not all be needed - reevaluate
```
erykwalder: All of them at least add information to the user agent or header which can be helpful in debugging. I think it's fine to leave all of them and clear this todo.

cmd/corectl/main.go
ameets: I think so. The idea here was to standardize the format between config.Nongenerator and config.Generator

cmd/corectl/main.go
bobg: No strong opinion; your call.

cmd/corectl/main.go
bobg: OK, fine as-is.

cmd/corectl/main.go
```diff
@@ -88,8 +91,18 @@ func configGenerator(db *sql.DB, args []string) {
 	flags.Parse(args)
 	args = flags.Args()
 
+	//not a blocksigner
```
bobg: Sorry to be so nitpicky but I'd like to see a space after the `//` in all comments.

core/blocksigner/blocksigner.go
```diff
@@ -53,6 +53,7 @@ func New(pub ed25519.PublicKey, hsm Signer, db pg.DB, c *protocol.Chain) *BlockS
 func (s *BlockSigner) SignBlock(ctx context.Context, b *bc.Block) ([]byte, error) {
 	sig, err := s.hsm.Sign(ctx, s.Pub, &b.BlockHeader)
 	if err != nil {
+		//TODO(ameets): rethink error pkg to preserve full stack trace
```
bobg: I think this TODO should be removed. It's not about this PR or this line of code.

(Also note that judicious use of `errors.Wrap`/`Wrapf` is even better than full stack traces, since they let us preserve exactly the important information and leave out the chaff.)

cmd/cored/main.go
bobg: Slight preference for (2).

cmd/corectl/main.go
ameets: Will leave as is for now.

ameets:
@bobg @erykwalder another PTAL

bobg:
LGTM, suggest waiting for @erykwalder's LGTM too

erykwalder:
LGTM

cmd/corectl/main.go
kr: FYI the back ticks are a feature of https://godoc.org/flag#PrintDefaults:

> The listed type, here int, can be changed by placing a back-quoted name in the flag's usage string; the first such item in the message is taken to be a parameter name to show in the message and the back quotes are stripped from the message when displayed. For instance, given
>
> ```
> flag.String("I", "", "search `directory` for include files")
> ```
>
> the output will be
>
> ```
> -I directory
> 	search directory for include files.
> ```

PR #485, opened 2017-02-07
core/mockhsm: return mockhsm keys in reverse chronological order

dominic:


dominic:
PTAL

jeffomatic:
Need to change pagination criteria a few lines up as well.

PR #486, opened 2017-02-07
sdk/java: accept multiple URLs in Client

jbowens:
Allow the Java SDK Client to be configured with multiple cored or
signerd URLs for high availability in the absence of a load balancer.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -528,6 +528,7 @@ public class Client {
     private ConnectionPool pool;
 
     public Builder() {
+      this.urls = new ArrayList<URL>();
       this.setDefaults();
     }
 
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -555,7 +555,7 @@ public class Client {
      * Adds a base URL for the client to use.
      * @param url the URL of the Chain Core or HSM.
      */
-    public Builder addURL(URL url) throws BadURLException {
+    public Builder addURL(URL url) {
       this.urls.add(url);
       return this;
     }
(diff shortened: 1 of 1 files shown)


kr:
:eyes:

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -414,14 +455,15 @@ private Response checkError(Response response) throws ChainException {
...
-    } catch (URISyntaxException e) {
-      throw new MalformedURLException();
+  private void nextURL(int failedIndex) {
+    if (this.urls.size() == 1) {
+      return; // No point contending on the CAS if there's only one URL.
     }
+
+    // A request to the url at failedIndex just failed. Move to the next
+    // URL in the list.
+    int nextIndex = (failedIndex + 1) % this.urls.size();
```
kr: This looks like a data race on this.urls if addURL or setURL is called concurrently. Should the docs for those methods say they mustn't be called concurrently with any RPCs?

jbowens: Those methods are on the Builder whereas this one is on the Client. Once a Client is constructed, `urls` should be immutable.

kr:
LGTM

jeffomatic:
👀

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -491,36 +532,55 @@ public Builder() {
...
+        throw new BadURLException(e.getMessage());
+      }
+      return this;
+    }
+
+    /**
+     * Adds a base URL for the client to use.
+     * @param url the URL of the Chain Core or HSM.
+     */
+    public Builder addURL(URL url) throws BadURLException {
```
jeffomatic: I don't think this method throws BadURLException.

jbowens:
Tested it locally, and it worked okay, but I discovered #487.

Issue #487, opened 2017-02-08
core: 500 forwarding to leader


jbowens:
Ran into this error while testing #486.

```
72010274Z status=500 chaincode=CH000 path=/build-transaction error="json: cannot unmarshal array into Go value of type map[string]interface {}"
/Users/jackson/src/chain/core/rpc/rpc.go:67 - chain/core/rpc.(*Client).Call
/Users/jackson/src/chain/core/api.go:331 - chain/core.(*Handler).forwardToLeader
/Users/jackson/src/chain/core/transact.go:78 - chain/core.(*Handler).build
/Users/jackson/src/chain/core/api.go:127 - chain/core.(*Handler).(chain/core.build)-fm
/usr/local/go/src/runtime/asm_amd64.s:516 - runtime.call128
app=cored buildtag=dev processID=chain-mba.local-25974-d430288902dc23074f0e reqid=36285cb64c07e3db6e75 at=http.go:52 t=2017-02-08T00:57:40.975637505Z status=500 chaincode=CH000 path=/build-transaction error="json: cannot unmarshal array into Go value of type map[string]interface {}"
```

jbowens:
Fixed in #488.

PR #488, opened 2017-02-08
core: fix leader forwarding

jbowens:
Fix build-transaction leader-forwarding to use an `interface{}` instead
of a `map[string]interface{}` (fixing #487). Also, remove an unnecessary
extra pointer indirection in `forwardToLeader`.


jbowens:
PTAL

bobg:
LGTM

PR #491, opened 2017-02-08
core/query: consistently handle absent aliases

jbowens:
In most places throughout core/query we expect alias columns to
be non-null and the empty string if not provided. The only exception is
when the entity being aliased might not even exist (for ex,
`account_alias` on an output which might not even have an associated
account_id).

The code for querying accounts and aliases expected absent aliases to
be stored in the database as the empty string. The indexing code stored
them as NULLs. This commit fixes the indexing code and makes the alias
columns on the annotated_accounts and annotated_assets tables
non-nullable.


jbowens:
PTAL

kr:
LGTM

PR #492, opened 2017-02-08
sdk/java: add receiver interface

jeffomatic:
This commit introduces receivers to the Java SDK.

- Adds Receiver class, and deprecates ControlProgram class.
- Adds Transaction.Action.ControlWithReceiver class, and deprecates Transaction.Action.ControlWithProgram class.
- Creation of receivers has been moved under the Account namespace, to leave room for a future, generic create-receiver API call.
- For a simple cross-core payment workflow, we'll use JSON as a serialization format for receivers. To help with this, our standard JSON serialization utilties will be moved to a common location, com.chain.common.Utils.

### DIFF
--- a/sdk/java/src/main/java/com/chain/api/Account.java
+++ b/sdk/java/src/main/java/com/chain/api/Account.java
@@ -320,8 +320,7 @@ public class Account {
      * @throws HTTPException This exception is raised when errors occur making http requests.
      * @throws JSONException This exception is raised due to malformed json requests or responses.
      */
-    public Receiver create(Client client)
-        throws ChainException {
+    public Receiver create(Client client) throws ChainException {
       return client.singletonBatchRequest(
           "create-account-receiver", Arrays.asList(this), Receiver.class, APIException.class);
     }
@@ -339,8 +338,8 @@ public class Account {
    * @throws HTTPException This exception is raised when errors occur making http requests.
    * @throws JSONException This exception is raised due to malformed json requests or responses.
    */
-  public static BatchResponse<Receiver> createReceiverBatch(Client client, List<ReceiverBuilder> builders)
-      throws ChainException {
+  public static BatchResponse<Receiver> createReceiverBatch(
+      Client client, List<ReceiverBuilder> builders) throws ChainException {
     return client.batchRequest(
         "create-account-receiver", builders, Receiver.class, APIException.class);
   }
--- a/sdk/java/src/main/java/com/chain/api/Account.java
+++ b/sdk/java/src/main/java/com/chain/api/Account.java
@@ -271,8 +271,8 @@ public class Account {
      * Specifies the account under which the receiver is created. You must use
      * this method or @{link ReceiverBuilder#setAccountId}, but not both.
      *
-     * @param alias the unique alias of the account
-     * @return the updated ReceiverBuilder object
+     * @param alias The unique alias of the account.
+     * @return The updated ReceiverBuilder object.
      */
     public ReceiverBuilder setAccountAlias(String alias) {
       this.accountAlias = alias;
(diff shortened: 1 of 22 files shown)


jeffomatic:
This is WIP and pending the introduction of the control-with-receiver txbuilder action.

sdk/java/src/main/java/com/chain/api/Transaction.java
```diff
@@ -875,8 +877,50 @@ public ControlWithAccount setAmount(long amount) {
...
+        return this;
+      }
+
+      /**
+       * Specifies the receiver (as a JSON string) that is being paid to.
+       *
+       * @param serialized the JSON-serialized receiver being paid to
+       * @return the updated action object
+       */
+      public ControlWithReceiver setReceiver(String serialized) throws JSONException {
```
jbowens: Maybe we should throw a more generic exception than a `JSONException` to preserve the flexibility to serialize receivers in other ways (ex, protocol buffers) in the future.

On the same subject, would a different type for `serialized` allow more flexibility for binary serialization formats?

jeffomatic: > Maybe we should throw a more generic exception than a JSONException to preserve the flexibility to serialize receivers in other ways (ex, protocol buffers) in the future.

We could do:

1. A base class called SerializationException that JSONException derives from.
2. A totally different class, leaving JSONException to refer to JSON deserialization errors in API responses.

> On the same subject, would a different type for serialized allow more flexibility for binary serialization formats?

We can method-overload for different types in the future. Is string-ness enough to suggest a JSON serialization, or should we allow for different string serializations, such a protobuf hexbytes? If the latter, then perhaps it's better if I include a serialization hint in the method name, e.g. `setReceiverFromJSON`?

jeffomatic: Update: I'm gonna specify `Json` in the associated method names. It's the most future-proof solution, and arguably a bonus for legibility.

jeffomatic:
PTAL

sdk/java/src/main/java/com/chain/api/Account.java
```diff
@@ -253,4 +253,94 @@ public Builder setTags(Map<String, Object> tags) {
...
+
+    @SerializedName("expires_at")
+    public Date expiresAt;
+
+    /**
+     * Specifies the account under which the receiver is created. You must use
+     * this method or @{link ReceiverBuilder#setAccountId}, but not both.
+     *
+     * @param alias The unique alias of the account.
+     * @return The updated ReceiverBuilder object.
```
jeffomatic: Really great question, I was wondering about myself as I was literally going through and capitalizing first letters, adding articles, and adding periods. It's kind of all over the place, so I just started doing what made sense to me in this PR. It would be good to codify that at some point.

FWIW, what do you think your preference would be?

jbowens: Might be worth looking at this: http://www.oracle.com/technetwork/articles/java/index-137868.html

Couple things that I like from there:
* Use "this" instead of "the" when referring to an object created from the current class.
* re `@param`: . The description begins with a lowercase letter if it is a phrase (contains no verb), or an uppercase letter if it is a sentence. End the phrase with a period only if another phrase or sentence follows it.

Even the stdlib's javadocs aren't particularly opinionated. I do like the lowercase, phrases for simple `@param` tags like here: [java.util.concurrent.atomic.AtomicInteger](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicInteger.html) But, maybe full sentences shouldn't be prohibited for more complicated parameters.

jeffomatic:
Need to add tests for deprecated features back in

sdk/java/src/main/java/com/chain/api/Account.java
jeffomatic: Thanks, this is really useful

PR #494, opened 2017-02-08
core/txbuilder: add `control_receiver` action

jbowens:
Add a transaction-building action to control assets with a Receiver.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -99,6 +99,7 @@ func (h *Handler) init() {
 	h.actionDecoders = map[string]func(data []byte) (txbuilder.Action, error){
 		"control_account":                h.Accounts.DecodeControlAction,
 		"control_program":                txbuilder.DecodeControlProgramAction,
+		"control_with_receiver":          txbuilder.DecodeControlWithReceiverAction,
 		"issue":                          h.Assets.DecodeIssueAction,
 		"retire":                         txbuilder.DecodeRetireAction,
 		"spend_account":                  h.Accounts.DecodeSpendAction,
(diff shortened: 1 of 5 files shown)


core/api.go
```diff
@@ -99,6 +99,7 @@ func (h *Handler) init() {
 	h.actionDecoders = map[string]func(data []byte) (txbuilder.Action, error){
 		"control_account":                h.Accounts.DecodeControlAction,
 		"control_program":                txbuilder.DecodeControlProgramAction,
+		"control_with_receiver":          txbuilder.DecodeControlWithReceiverAction,
```
jbowens: @jeffomatic — are we sure we want this to be `control_with_receiver`? `control_receiver` would fit the existing action types more closely.

jeffomatic: Good question. I actually started (then paused) a PR that would allow the core to accept action labels with prepositions, e.g. "control_with_account", etc., in addition to the ones they accept now. The problem we have now is that the SDKs expose the form with prepositions to the user (because they are way more legible), and the wire protocol doesn't use them. Is that too bikesheddy?

jbowens: Yeah, I think the wire protocol and the SDK interface are very different domains. I don't think the wire protocol should be considered part of the product/public interface. It's an implementation detail.

There's still value in consistency within the wire protocol for the same reason we value consistency in code style. Consistency between the SDKs and wire protocol is nice if it's easy, but otherwise, not worth it IMO.

jeffomatic: After futzing with it, I decided that it's too much of a microchange to humanize the action labels here. "control_with_account" is preferable to "control_account", but the value prop is too limited to justify updating the wire protocol and SDKs. So `control_receiver` it is.

PR #495, opened 2017-02-08
core: wrap batch panics with stack trace

jbowens:
When recovering a panic in a batch endpoint, wrap the panic with the
stack trace. Otherwise, the reported panic doesn't include a stack trace
and can be difficult to debug. Also, bump the number of frames that we
collect.


kr:
LGTM

PR #496, opened 2017-02-08
errors: add new function Sub

kr:
Sub takes an existing error, which might have an associated stack trace or other metadata, and substitutes a new root error, which is sometimes necessary to return a certain error code to the client.

This change also updates all call sites I could find to use the new function.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -154,3 +154,19 @@ func Data(err error) map[string]interface{} {
 	wrapper, _ := err.(wrapperError)
 	return wrapper.data
 }
+
+// Sub returns an error with all the associated data from old,
+// including stack trace, detail, message, and data, but using
+// new as the root error. It also wraps the newly-created error
+// with the formatted error string from old.
+//
+// Use this when you need to substitute a new root error in place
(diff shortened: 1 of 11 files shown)


kr:
PTAL

errors/errors.go
```diff
@@ -154,3 +154,25 @@ func Data(err error) map[string]interface{} {
...
+// including stack trace, detail, message, and data, but using
+// new as the root error. It also wraps the newly-created error
+// with the formatted error string from old.
+//
+// Sub returns nil when either new or old is nil.
+//
+// Use this when you need to substitute a new root error in place
+// of an existing error that may already hold a stack trace
+// or other metadata.
+func Sub(new, old error) error {
```
jbowens: I don't mind but `new` might look a little weird with syntax highlighting.

jbowens:
lgtm besides the nit on the variable name

errors/errors.go
kr: Hmm, personally I prefer it as `new`; that seems like the right name to me. (Syntax highlighting ought to know that `new` is a valid variable name.)

PR #497, opened 2017-02-09
core/leader: run lead func synchronously

### LINKED ISSUE #489: core: nil pointer dereference in /build-transaction
I noticed this error when failing over between two cored processes with high transaction throughput. Shortly after the secondary becomes leader, these nil pointer dereferences happen. A little while later, all build-transaction calls succeed. It doesn't always happen on a failover. It seems like it's probably a race condition surrounding some leader-only state (maybe the state tree?).

I wish we reported a stack trace of where the panic was.

```
app=cored buildtag=dev processID=chain-mba.local-28703-d06d338b073853927021 reqid=c9adf2bf2996e67cc980 at=http.go:52 t=2017-02-08T01:33:48.007350845Z subreqid=782d7ec96a950d4a654f status=500 chaincode=CH000 path=/build-transaction error="runtime error: invalid memory address or nil pointer dereference"
app=cored buildtag=dev processID=chain-mba.local-28703-d06d338b073853927021 reqid=f47f7464280f03e67a68 at=http.go:52 t=2017-02-08T01:33:48.014702926Z subreqid=3d069c6bd5ea0fdd782e status=500 chaincode=CH000 path=/build-transaction error="runtime error: invalid memory address or nil pointer dereference"
```

jbowens:
This refactors the `core/leader` package and how it's used by `cmd/cored`:
* the `lead` closure is run synchronously within the `leader.Run` loop
* the global `IsLeading` value isn't updated until the `lead` closure completes
* the`cmd/cored`'s `lead` closure calls `Recover` instead of pushing the responsibility to the `core/fetch` and `core/generator` packages
* the `leader.Run` loop responsible for invoking `lead` is separated from the loop responsible for detecting leadership changes

This ensures that all leadership-only routines and code paths are guaranteed to have a recovered, current blockchain state. It also ensures that if blockchain recovery takes longer than 10 seconds, a new leader won't be demoted prematurely.

In a followup PR, we'll be able to remove [this](https://github.com/chain/chain/blob/main/cmd/cored/main.go#L350-L368) goroutine from `cmd/cored`, move the pin creating into the `lead` closure and delete the `protocol.Chain.Ready` function.

Fixes #489.

### DIFF
--- a/core/leader/leader.go
+++ b/core/leader/leader.go
@@ -47,17 +47,18 @@ func Run(db *sql.DB, addr string, lead func(context.Context)) {
 	var leadCtx context.Context
 	var cancel func()
 	for leader := range leadershipChanges(ctx, l) {
-		isLeading.Store(leader)
-
 		if leader {
 			log.Messagef(ctx, "I am the core leader")
 			leadCtx, cancel = context.WithCancel(ctx)
-			go l.lead(leadCtx)
+			l.lead(leadCtx)
 		} else {
 			log.Messagef(ctx, "No longer core leader")
 			cancel()
 		}
+
+		isLeading.Store(leader)
 	}
+	panic("unreachable")
 }
 
 // leadershipChanges spawns a goroutine to check if this process
--- a/core/leader/leader.go
+++ b/core/leader/leader.go
@@ -18,7 +18,11 @@ var isLeading atomic.Value
 // IsLeading returns true if this process is
 // the core leader.
 func IsLeading() bool {
-	return isLeading.Load().(bool)
+	v := isLeading.Load()
+	if v == nil {
+		return false
+	}
+	return v.(bool)
 }
 
 // Run runs as a goroutine, trying once every five seconds to become
--- a/core/leader/leader.go
+++ b/core/leader/leader.go
@@ -69,6 +69,12 @@ func Run(db *sql.DB, addr string, lead func(context.Context)) {
 // is leader periodically. Every time the process becomes leader
 // or is demoted from being a leader, it sends a bool on the
 // returned channel.
+//
+// It provides the invariants:
+// * the first value sent on the channel is `true` when the
+//   process is first elected leader
+// * every value sent on the channel is the opposite of the
+//   previous value
 func leadershipChanges(ctx context.Context, l *leader) chan bool {
 	ch := make(chan bool)
 	go func() {
--- a/core/leader/leader.go
+++ b/core/leader/leader.go
@@ -4,7 +4,7 @@ package leader
 
 import (
 	"context"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"chain/database/pg"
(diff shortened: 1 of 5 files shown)


core/leader/leader.go
```diff
@@ -50,10 +44,45 @@ func Run(db *sql.DB, addr string, lead func(context.Context)) {
 	}
 	log.Messagef(ctx, "Using leaderKey: %q", l.key)
 
-	update(ctx, l)
-	for range time.Tick(5 * time.Second) {
-		update(ctx, l)
+	var leadCtx context.Context
+	var cancel func()
+	for leader := range leadershipChanges(ctx, l) {
```
kr: Sending a value over this channel raises the possibility that it might not strictly alternate. Would it be better to track that in this loop? like

```
leader := false
for range leadershipChanges(ctx, l) {
    leader = !leader
    if leader ...
}
```

Not sure, what do you think?

jbowens: Like if something else sent a value over the channel? Right now, the `leader` value is ferried directly from the source of truth, which is nice. Moving the `leader` value into this loop would make it clearly strictly alternate, but introduce the possibility that it might be on the wrong state.

What do you think about something like

```
var leadCtx context.Context
var cancel func()
for {
    <-becameLeaderCh
    log.Messagef(ctx, "I am the core leader")
    leadCtx, cancel = context.WithCancel(ctx)
    l.lead(leadCtx)
    isLeading.Store(true)

    <-demotedCh
    log.Messagef(ctx, "No longer core leader")
    cancel()
    isLeading.Store(false)
}
```

kr: Yeah I guess either way it depends on the leadershipChanges loop behaving correctly. There are two invariants it provides:

- each value is the opposite of the previous value
- the first value is true

My suggestion just moved us from depending on that first invariant to depending on the second one.

On second thought, I think it's ok to depend on either one of those invariants, but we should probably document them both in leadershipChanges. So IMO this code is fine as-is (with the addition of a brief comment).

The two-channel option also seems ok, but it has the possibility of deadlocking if either loop gets out of sync. An advantage of your original approach here is that it would be easy for this loop to enforce the invariant by also tracking its own bool variable, if we wanted to be that paranoid. (I don't think it's necessary, but I wouldn't object, either.)

jbowens:
updated with documentation of `leadershipChanges`'s invariants

core/leader/leader.go
```diff
@@ -69,6 +69,12 @@ func Run(db *sql.DB, addr string, lead func(context.Context)) {
 // is leader periodically. Every time the process becomes leader
 // or is demoted from being a leader, it sends a bool on the
 // returned channel.
+//
+// It provides the invariants:
+// * the first value sent on the channel is `true` when the
```
kr: Sorry to nitpick, but this could be interpreted as "the first value sent on the channel is true but only under certain conditions: when the process ...".

How about "The first value sent on the channel is true. (This will happen at the time the process is first elected leader.)"?

jbowens:
PTAL

kr:
LGTM

PR #499, opened 2017-02-09
core/coreunsafe: use config.Production

jbowens:
Use `config.Production` instead of the more fragile expvar to protect
against a production environment.


kr:
LGTM

kr:
I would also like to explore having the prod build constraint arrange it so that this package doesn't even get linked in. (Checked in wercker using `go tool nm` or some other mechanism.)

PR #500, opened 2017-02-09
errors: clarify documentation

kr:
Using these parameter names for Sub seems to make its
behavior clearer.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -155,10 +155,9 @@ func Data(err error) map[string]interface{} {
 	return wrapper.data
 }
 
-// Sub returns an error with all the associated data from err,
-// including stack trace, detail, message, and data, but using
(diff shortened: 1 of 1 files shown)


kr:
PTAL

This also addresses the potential syntax highlighting problem raised in https://github.com/chain/chain/pull/496#discussion_r100225050.

bobg:
Actually, `root` sounds like it's going to be the more deeply buried error. How about `newRoot` and `orig`? :bike: :house:

kr:
Let me try to adjust the wording to hopefully make it clearer. (I already wasn't super happy with the existing wording.)

kr:
Ok, updated. There is no "new" and "old" involved. It is just a root from one place and everything else from another place.

bobg:
LGTM

PR #502, opened 2017-02-09
sdk/ruby: fix path joining

jeffomatic:
The Connection class does not appropriately join URL path components
when both trailing and leading slashes occur in adjacent components.
This can occur when connecting to a real HSM server and attempting to
sign.

I chose a more brute-force approach for this fix, despite the existence
of something like URI::join. The latter has issues when adjacent paths
are both relative. The fix in this commit deals with this case
correctly.

### DIFF
--- a/sdk/ruby/lib/chain/connection.rb
+++ b/sdk/ruby/lib/chain/connection.rb
@@ -111,7 +111,10 @@ module Chain
           http.send "#{k}=", @opts[k]
         end
 
-        req = Net::HTTP::Post.new(@url.request_uri + path)
+        full_path = @url.request_uri.chomp('/')
+        full_path += (path[0] == '/') ? path : '/' + path
+
+        req = Net::HTTP::Post.new(full_path)
         req['Accept'] = 'application/json'
         req['Content-Type'] = 'application/json'
         req['User-Agent'] = 'chain-sdk-ruby/' + Chain::VERSION
--- a/sdk/ruby/spec/unit/connection_spec.rb
+++ b/sdk/ruby/spec/unit/connection_spec.rb
@@ -1,0 +1,21 @@
+require 'chain'
+require 'webmock'
+
+include WebMock::API
+WebMock.enable!
+
+describe Chain::Connection do
+
+  example 'works with mixtures of relative and absolute paths' do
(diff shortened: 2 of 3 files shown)


sdk/ruby/lib/chain/connection.rb
```diff
@@ -111,7 +111,10 @@ def _single_request(path, body)
           http.send "#{k}=", @opts[k]
         end
 
-        req = Net::HTTP::Post.new(@url.request_uri + path)
+        full_path = @url.request_uri.chomp('/')
+        full_path += (path[0] == '/') ? path : '/' + path
```
tonychain: Personally I'd just do:

```ruby
full_path = @url.request_uri.sub(/^\/+/, "/")
```

jeffomatic: I don't know if consecutive slashes are possible in this case, since `@url` is a parsed URI object.

jeffomatic: How about I do your trick, but preserving the original string addition.

jeffomatic: Hmm n/m, it just gets a bit messier.

jeffomatic:
PTAL

tonychain:
Yay tests

sdk/ruby/spec/unit/connection_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  example 'works with mixtures of relative and absolute paths' do
+    stub_request(:any, 'foo.test/bar').to_return(body: '{}', headers: {'Chain-Request-ID' => 'test'})
+    stub_request(:any, 'foo.test/bar/baz').to_return(body: '{}', headers: {'Chain-Request-ID' => 'test'})
+
+    expect {
+      Chain::Connection.new(url: 'http://foo.test').request('bar')
+      Chain::Connection.new(url: 'http://foo.test').request('/bar')
+      Chain::Connection.new(url: 'http://foo.test/bar').request('baz')
+      Chain::Connection.new(url: 'http://foo.test/bar').request('/baz')
```
kr: Should we also have tests for

```
Chain::Connection.new(url: 'http://foo.test/').request('bar')
Chain::Connection.new(url: 'http://foo.test/').request('/bar')
```

kr: (Note, done in #1038.)

PR #504, opened 2017-02-10
docs/enterprise/thales: update machine image loading

ryandotsmith:
@erykwalder and @kr discovered a setup command that will install the machine image in such a way that it doesn't need to be reloaded in the event of an HSM reboot.

Related: chain/chainprv#2453


erykwalder:
LGTM. Will users typically have more than one module? I think we just have one.

ryandotsmith:
@erykwalder For production environments where there are multiple HSMs, there will be multiple modules.

PR #505, opened 2017-02-10
protocol: remove Chain.Ready

jbowens:
With the changes from #497, we can now move the block processor pin
creation into the `leader.Run` closure after the blockchain is
recovered.


erykwalder:
LGTM

PR #506, opened 2017-02-10
sdk/node: refactor test directories

jeffomatic:
The goal here is to make sure Mocha options are available to all invocations of Mocha, not just via `npm run test`.


dominic:
I think it's a little confusing to have `testSrc` with test files, and `test` with a single config in it. I'd like to propose two different potential paths:

1. Move all test files to `/test`, and compile tests to `/testsCompiled`. The opts file would then be:
    
    ```
    testsCompiled
    -t 60000
    ```

    We'd have to update the `.gitignore` and `package.json` pretest steps accordingly.

2. Move `mocha.opts` to `testSrc` and change the `package.json` test command to:

    ```
    "test": "mocha --opts testSrc/mocha.opts",
    ```

jeffomatic:
The point of this PR is to allow any invocation of Mocha (even outside of package.json) to use our defaults, so that eliminates option 2. Option 1 is okay, but it raises an inconsistency in the namescheme--the library source itself is in `src` and is compiles to `dist`, and the test source is in `test` and compiled into `testCompiled`.

One question I'll ask is whether we should simply run the tests on the uncompiled, pre-Babel files, not not bother compiling the tests, which don't get distributed to the browser anyway.

PR #507, opened 2017-02-10
sdk/node: add receivers interface

jeffomatic:
- Add createReceiver and createReceiverBatch
- Deprecate createControlProgram
- Add controlWithReceiver txbuilder action
- Deprecate controlWithProgram txbuilder action

### DIFF
--- a/sdk/node/testSrc/callbacks.js
+++ b/sdk/node/testSrc/callbacks.js
@@ -421,7 +421,7 @@ describe('Callback style', () => {
         next()
       }),
 
-      // Control program creation
+      // Control program creation (deprecated)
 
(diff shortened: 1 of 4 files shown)


sdk/node/testSrc/callbacks.js
```diff
@@ -365,6 +365,62 @@ describe('Callback style', () => {
...
+        }, cb),
+        (issuance, cb) => signer.sign(issuance, cb),
+        (signed, cb) => client.transactions.submit(signed, cb),
+      ], (err, submitted) => {
+        assert(err === null)
+        assert(submitted.id)
+        next()
+      }),
+
       // Control program creation
```
dominic: Mark these tests as deprecated too?

jeffomatic: Nice find

PR #509, opened 2017-02-10
core/asset: ignore non-JSON asset definitions

jbowens:
Chain Core won't allow assets to be created with non-JSON asset
definitions, but a buggy Chain Core or other clients on a network
may allow it. Instead of crashing the block processor, the error
should be ignored.


kr:
LGTM

PR #512, opened 2017-02-11
dashboard: onboarding tutorial

chrisgarvin:


dashboard/src/features/app/reducers.js
```diff
@@ -22,6 +22,14 @@ export const flashMessages = (state = new Map(), action) => {
       }
     }
 
+    case 'UPDATE_TUTORIAL': {
+      if(tutorial.isShowing && tutorial.currentStep.objectType == action.object){
```
jeffomatic: whitespace

dashboard/src/features/tutorial/components/Tutorial.jsx
```diff
@@ -1,35 +1,30 @@
...
-            handleDismiss={this.props.dismissTutorial}
-          />
-        }
-    </div>
+      {tutorialOpen && (tutorialTypes.includes(this.props.currentStep['component'])) &&
+        <TutorialComponent
+          userInput={userInput}
+          {...this.props.currentStep}
+          handleNext={() => this.props.showNextStep(tutorialRoute)}/>}
+      </div>
```
jeffomatic: indentation

dashboard/src/features/tutorial/components/TutorialForm/TutorialForm.jsx
```diff
@@ -2,49 +2,88 @@ import React from 'react'
...
+    this.state = { showFixed: false }
+
+    // We must bind in the constructor so that we have a reference to the bound
+    // function to remove from the window event listener later.
+    this.handleScroll = this.handleScroll.bind(this)
+  }
+
+  handleScroll(event) {
+    const scrollTop = event.srcElement.scrollingElement.scrollTop
+    this.setState({showFixed: scrollTop > 140})
```
jeffomatic: Where does the number `140` come from?

dashboard/src/features/tutorial/components/TutorialForm/TutorialForm.jsx
```diff
@@ -2,49 +2,88 @@ import React from 'react'
...
+                  <tr>
                     <td className={styles.listBullet}>{i+1}</td>
-                    <td>{str}</td>
+                    <td>{title}</td>
                   </tr>
                 ]
                 if (contentLine['description']) {
-                  rows.push (<tr className={styles.listItemDescription}>
+                  let descriptionResult = []
+                  contentLine['description'].forEach(function (descriptionLine){
```
jeffomatic: I think we can use the `() =>` convention here for lambdas.

dashboard/src/features/tutorial/components/TutorialForm/TutorialForm.jsx
```diff
@@ -2,49 +2,88 @@ import React from 'react'
...
+                      let replacement = userInput[descriptionLine['type']] || descriptionLine['type']
+                      if ('index' in descriptionLine){
+                        replacement = replacement[descriptionLine['index']]
+                      }
+
+                      if (replacement.hasOwnProperty('alias')) {
+                        replacement = replacement['alias'] || ''
+                      }
+
+                      description.split('STRING').forEach(function (item, i, arr){
```
jeffomatic: I think we can use the `() =>` convention here for lambdas.

dashboard/src/features/tutorial/components/TutorialModal/TutorialModal.jsx
```diff
@@ -0,0 +1,63 @@
...
+import { connect } from 'react-redux'
+
+const mapStateToProps = (state) => ({
+  tutorialRoute: state.tutorial.route,
+  currentStep: state.tutorial.currentStep,
+  showTutorial: state.routing.locationBeforeTransitions.pathname.startsWith(state.tutorial.route)
+})
+
+const mapDispatchToProps = ( dispatch ) => ({
+  dismissTutorial: () => dispatch(actions.dismissTutorial)
```
jeffomatic: You may want to double-check with @dominic, but I feel like one of the takeaways from our refactoring conversations is that we should be moving away from trivial action variables like `dismissTutorial`, and simply inline an action object such as `{type: 'DISMISS_TUTORIAL'}`. The rationale there is that you can search for the action string right away; as the code stands, you first have to look up `dismissTutorial` in order to figure out what reducer ultimately deals with this action.

dashboard/src/features/tutorial/reducers.js
```diff
@@ -1,7 +1,11 @@
 import { combineReducers } from 'redux'
+import steps from './steps.json'
 
 export const step = (state = 0, action) => {
-  if (action.type == 'TUTORIAL_NEXT_STEP' || action.type == 'UPDATE_TUTORIAL'){
+  if (action.type == 'TUTORIAL_NEXT_STEP'){
+    return state + 1
+  }
+  else if (action.type == 'UPDATE_TUTORIAL' && steps[state].objectType == action.object) {
```
jeffomatic: The `else` here is technically unnecessary. If we still want to have the `else`, though, it should appear on the line above, next to the previous closing brace.

dashboard/src/features/tutorial/reducers.js
```diff
@@ -14,8 +18,21 @@ export const isShowing = (state = true, action) => {
   return state
 }
 
-export const userInputs = (state = { accounts: [] }, action) => {
-  if (action.type == 'UPDATE_TUTORIAL'){
+export const route = (currentStep) => (state = 'transactions', action) => {
+  if (action.type == 'TUTORIAL_NEXT_STEP'){
+    return action.route
+  }
+  else if (action.type == 'UPDATE_TUTORIAL' && currentStep.objectType == action.object){
```
jeffomatic: See above re: `else`.

dashboard/src/features/tutorial/reducers.js
```diff
@@ -14,8 +18,21 @@ export const isShowing = (state = true, action) => {
...
+    return action.object + 's'
+  }
+  else if (action.type == 'DISMISS_TUTORIAL'){
+    return 'transactions'
+  }
+  return state
+}
+
+export const userInputs = (currentStep) => (state = { accounts: [] }, action) => {
+  if (action.type == 'UPDATE_TUTORIAL' && currentStep.objectType == action.object){
```
jeffomatic: whitespace

dashboard/src/features/tutorial/reducers.js
```diff
@@ -33,8 +50,22 @@ export const userInputs = (state = { accounts: [] }, action) => {
...
-export default combineReducers({
-  step,
-  isShowing,
-  userInputs
-})
+export default (state = {}, action) => {
+  const tutorialRoute = state.route
+  const tutorialUserInputs = state.userInputs
+
+  delete state.currentStep // combineReducers logs error because currentStep set outside of function
```
jeffomatic: Can you clarify what's going on here? Deleting items out of the old state object in a reducer shouldn't do anything. `newState` should completely replace this value.

dashboard/src/reducers.js
```diff
@@ -59,5 +59,8 @@ const makeRootReducer = () => (state, action) => {
     tutorial,
     unspent,
   })(state, action)
+  finalState.app = app(state.app, action, finalState.tutorial)
```
jeffomatic: I wonder if we can try to disentangle this state a little bit. It seems like the app reducer cares about state under the `app` key, but the `tutorial` key as well, which is a bit convoluted. Is there a reason why `tutorial` isn't a subkey under `app`?

dashboard/src/features/tutorial/components/TutorialForm/TutorialForm.jsx
dominic: This is a 100% magic Dominic constant.

dashboard/src/features/app/reducers.js
```diff
@@ -104,8 +104,8 @@ export const dropdownState = (state = '', action) => {
   return state
 }
 
-export default combineReducers({
+export default (state, action) => combineReducers({
```
jeffomatic: I don't think this change ends up doing anything, right?

dashboard/src/features/tutorial/components/TutorialForm/TutorialForm.jsx
jeffomatic: Can we add a comment that explains the magic number here?

dashboard/src/reducers.js
```diff
@@ -43,7 +43,7 @@ const makeRootReducer = () => (state, action) => {
     window.location.href = '/'
   }
 
-  return combineReducers({
+  const finalState = combineReducers({
```
jeffomatic: Doesn't look like this change does anything.

chrisgarvin: Ah yes, you're correct. In the end we got rid of any of the changes made within this reducer. Will revert to the way it previously existed.

PR #514, opened 2017-02-12
dashboard: remove interim API library

dominic:
Dashboard now uses the official Chain Node.js SDK


jeffomatic:
Will be posting screenshots of things I notice as I QA this branch.

jeffomatic:
Account/asset keys are not displayed correctly

screen shot 2017-02-15 at 6 00 35 pm

jeffomatic:
Some transaction details are appearing as null

screen shot 2017-02-15 at 6 03 19 pm

jeffomatic:
tx inputs have a "control program" field, which seems incorrect

jeffomatic:
Outputs in tx display have transaction ID, which is redundant, and the values are null

screen shot 2017-02-15 at 6 08 12 pm

jeffomatic:
Client token and network token links are broken

jeffomatic:
After clicking Core Status: Delete All Data, the subsequent Core Config page flickers

jeffomatic:
Error due to not selecting a pre-existing HSM key is a little unclear about the cause

screen shot 2017-02-15 at 6 22 58 pm

jeffomatic:
Resetting core does not delete MockHSM keys

jeffomatic:
Some outputs have null as output ID. This was the result of spending an unspent output and retiring it. Unsure if tx structure is related.

screen shot 2017-02-15 at 6 33 47 pm

dominic:
> Error due to not selecting a pre-existing HSM key is a little unclear about the cause

That's on main too - we should extract that into an isssue.

jeffomatic:
pagination doesn't work, still uses `context`

PR #520, opened 2017-02-13
cmd/cored: add version cmd line flag

ameets:


erykwalder:
Note to reviewers: this will change the log prefixes.

```
log.SetPrefix("cored-" + buildTag + ": ")
```

```
chainlog.SetPrefix(append([]interface{}{"app", "cored", "buildtag", buildTag, "processID", processID}, race...)...)
```

cmd/cored/main.go
```diff
@@ -116,6 +117,17 @@ func init() {
 }
 
 func main() {
+	v := flag.Bool("version", false, "prints cored version")
+	flag.Parse()
+
+	if *v {
+		fmt.Printf("cored version: %s production=%t\n", config.Version, config.Production)
```
kr: :bike::house: can I request `"cored %s production=%t"` as the format string?

kr: my rationale:

- less stutter (the command already says `-version`)
- closer to the [GNU style](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion) (for whatever that's worth)
- looks nicer to my eye

kr: actually, can we put the `production=%t` on a separate line? it would also be nice to put the buildCommit and buildDate as well, since they're available.

I want to use this format:

```
cored 1.1
production=true
build-commit=abc123
build-date=2017-02-13T21:28:52Z
```
or

```
cored 1.1+changes
production=false
build-commit=?
build-date=?
```

(I'll do the same in signerd.)

erykwalder:
updated, PTAL

kr:
LGTM

PR #522, opened 2017-02-13
core/txbuilder: check txsighash for issuance-only txs, too

### LINKED ISSUE #169: Failing to add key to SDK signer produces misleading error message

tessr:
Should resolve #169 for real this time.

### DIFF
--- a/core/txbuilder/finalize.go
+++ b/core/txbuilder/finalize.go
@@ -69,7 +69,6 @@ var (
 )
 
 func checkTxSighashCommitment(tx *bc.Tx) error {
-	allIssuances := true
 	var lastError error
 
 	for i, inp := range tx.Inputs {
@@ -77,7 +76,6 @@ func checkTxSighashCommitment(tx *bc.Tx) error {
 		switch t := inp.TypedInput.(type) {
 		case *bc.SpendInput:
 			args = t.Arguments
-			allIssuances = false
 		case *bc.IssuanceInput:
 			args = t.Arguments
 		}
@@ -114,11 +112,7 @@ func checkTxSighashCommitment(tx *bc.Tx) error {
 		return nil
 	}
 
-	if !allIssuances {
-		return lastError
-	}
-
-	return nil
+	return lastError
 }
 
 // RemoteGenerator implements the Submitter interface and submits the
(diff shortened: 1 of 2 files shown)


bobg:
:eyes:

bobg:
LGTM (through narrowed eyes)

PR #524, opened 2017-02-13
cmd/cored: disable mockhsm in prod

ameets:


core/api.go
```diff
@@ -123,6 +123,13 @@ func (h *Handler) init() {
 	m := http.NewServeMux()
 	m.Handle("/", alwaysError(errNotFound))
 
+	if h.MockHSM != nil {
+		m.Handle("/mockhsm/create-key", devOnly(needConfig(h.mockhsmCreateKey)))
+		m.Handle("/mockhsm/list-keys", devOnly(needConfig(h.mockhsmListKeys)))
+		m.Handle("/mockhsm/delkey", devOnly(needConfig(h.mockhsmDelKey)))
+		m.Handle("/mockhsm/sign-transaction", devOnly(needConfig(h.mockhsmSignTemplates)))
+	}
```
jbowens: do we want to 404 or do we want to error if mockhsm endpoints are hit in production? That `devOnly` function wraps the handler and ensures that it'll return a relevant error if called in production.

erykwalder:
updated, PTAL

bobg:
This PR seems to be obsoleted by https://github.com/chain/chain/pull/521

PR #525, opened 2017-02-14
core: initialize handler before first request

erykwalder:
Rather than initializing the http handler on the first request, this
makes the creation of the handler an explicit step.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -177,11 +177,11 @@ func runServer() {
 		h = launchConfiguredCore(ctx, db, conf, processID)
 	} else {
 		chainlog.Messagef(ctx, "Launching as unconfigured Core.")
-		h = &core.API{
+		h = core.Handler(&core.API{
 			DB:           db,
 			AltAuth:      authLoopbackInDev,
 			AccessTokens: &accesstoken.CredentialStore{DB: db},
-		}
+		})
 	}
 
 	secureheader.DefaultConfig.PermitClearLoopback = true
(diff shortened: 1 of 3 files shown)


erykwalder:
I know this seems big, but most of it is just naming

bobg:
:eyes:

bobg:
How would you feel about reorganizing this PR into separate commits, one that's purely renaming, one that makes initialization a separate step, and one that changes the action decoders?

erykwalder:
Will do

erykwalder:
rebased

cmd/cored/main.go
```diff
@@ -177,11 +177,11 @@ func runServer() {
 		h = launchConfiguredCore(ctx, db, conf, processID)
 	} else {
 		chainlog.Messagef(ctx, "Launching as unconfigured Core.")
-		h = &core.API{
+		h = core.Handler(&core.API{
```
bobg: The fact that `launchConfiguredCore` also contains a call to `core.Handler` caused a momentary hiccup in my understanding of the flow here. It might be a little clearer if this `else` branch called a parallel function, `launchUnconfiguredCore`, rather than calling `core.Handler` directly. A non-PR-blocking suggestion.

bobg:
LGTM (but see comment)

erykwalder:
updated w/ feedback. Please take another look :)

cmd/cored/main.go
```diff
@@ -176,12 +177,7 @@ func runServer() {
 	if conf != nil {
```
bobg: Hm, now I wonder whether this `if/else` shouldn't be encapsulated inside a single `launchCore` function, and maybe it should be called `newHandler`? In any case, still your call, still LGTM.

bobg:
One new comment, still LGTM

PR #526, opened 2017-02-14
core: rename Handler to API

erykwalder:


bobg:
LGTM, and thanks for rejiggering https://github.com/chain/chain/pull/525

PR #527, opened 2017-02-14
core: make action decoders a lookup function

erykwalder:
Rather than having a map that is initialized at the first request,
action decoders should be available on the API at all times. This
changes the map into a function.


bobg:
LGTM

PR #528, opened 2017-02-14
cmd/cored: remove mockhsm dependency in prod

erykwalder:
When the prod build tag is used, mockhsm should not be included in the
final binary.


bobg:
LGTM

PR #529, opened 2017-02-14
cmd/corectl: check key length in generator config

erykwalder:
Satisfies an old todo about checking key lengths. The mockhsm can now
generate bare ed25519 keys, so the length should be enforced.


bobg:
LGTM

PR #530, opened 2017-02-14
cmd/corectl: disable key generation in prod

erykwalder:
This disables the create-block-keypair command when the prod build tag
is used. Additionally, mockhsm is not compiled into the binary when the
prod build tag is used.


bobg:
LGTM

PR #531, opened 2017-02-15
core/txfeed: fix not-found status code for Find()

jeffomatic:
When retrieving a transaction feed by alias, an empty response
should return a CH002 error, rather than a CH000. This is resolved
by substituting sql.ErrNoRows with pg.ErrUserInputNotFound.

### DIFF
--- a/core/txfeed/txfeed.go
+++ b/core/txfeed/txfeed.go
@@ -130,6 +130,9 @@ func (t *Tracker) Find(ctx context.Context, id, alias string) (*TxFeed, error) {
 	)
 
 	err := t.DB.QueryRow(ctx, q.String(), id).Scan(&feed.ID, &sqlAlias, &feed.Filter, &feed.After)
+	if err == sql.ErrNoRows {
+		return nil, errors.WithDetailf(pg.ErrUserInputNotFound, "alias: %s", alias)
+	}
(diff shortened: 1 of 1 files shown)


jeffomatic:
PTAL

core/txfeed/txfeed.go
```diff
@@ -130,6 +130,9 @@ func (t *Tracker) Find(ctx context.Context, id, alias string) (*TxFeed, error) {
 	)
 
 	err := t.DB.QueryRow(ctx, q.String(), id).Scan(&feed.ID, &sqlAlias, &feed.Filter, &feed.After)
+	if err == sql.ErrNoRows {
+		return nil, errors.WithDetailf(pg.ErrUserInputNotFound, "alias: %s", alias)
```
kr: We have a new function! https://godoc.org/github.com/chain/chain/errors#Sub

I suggest:

```
err = errors.Sub(pg.ErrUserInputNotFound, err)
err = errors.WithDetailf(err, "alias: %s", alias)
return nil, err
```

This way if err already has a stack trace or detail or whatever, they'll all be preserved.

jeffomatic:
added call to `errors.Sub`, PTAL

kr:
LGTM

PR #534, opened 2017-02-15
protocol/bc: switch to txgraph-based hashing for block headers

bobg:
Yes, `protocol/tx` is a weird package name for something having to do with blocks. But it's temporary (as are the anti-circular-reference variables and all the `init_test.go` files.)


kr:
LGTM

PR #535, opened 2017-02-15
protocol/state: replace state.Output with bc.OutputID

bobg:
Turns out we no longer need to carry around the `TxOutput`.

The indexer was using the `TxOutput` in a `state.Output`, but it can get that same info elsewhere, and now does.


bobg:
PTAL

bobg:
PTAL

kr:
:eyes:

kr:
LGTM

PR #537, opened 2017-02-16
core/query: index retired outputs

erykwalder:
Since the annotated_outputs table is also used for lookups on the
annotated_txs table, retired outputs should be indexed.


jbowens:
what should we do about the timespan column for retired outputs? gotta be careful that they won't be used for /list-unspent-outputs and /list-balances queries. maybe mark them as spent as we're inserting them?

erykwalder:
good idea

erykwalder:
updated, PTAL

jbowens:
lgtm

PR #538, opened 2017-02-16
protocol/tx: assign output id for all types

erykwalder:
All outputs should have an output id assigned, whether they are standard
outputs or retirement outputs. Since the header.body.Results field
contains only outputs, there should be no need for further filtering.

### DIFF
--- a/protocol/bc/txhashes.go
+++ b/protocol/bc/txhashes.go
@@ -3,9 +3,11 @@ package bc
 type (
 	// TxHashes holds data needed for validation and state updates.
 	TxHashes struct {
-		ID        Hash
-		OutputIDs []Hash // each OutputID is also the corresponding UnspentID
-		Issuances []struct {
+		ID Hash
+		// contains OutputIDs and retirment hashes.
+		// each OutputID is also the corresponding UnspentID
+		ResultHashes []Hash
+		Issuances    []struct {
 			ID           Hash
 			ExpirationMS uint64
 		}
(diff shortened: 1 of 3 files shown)


erykwalder:
@bobg @tessr PTAL

protocol/tx/transaction.go
```diff
@@ -29,10 +29,7 @@ func TxHashes(oldTx *bc.TxData) (hashes *bc.TxHashes, err error) {
 	// OutputIDs
 	hashes.OutputIDs = make([]bc.Hash, len(header.body.Results))
 	for i, resultHash := range header.body.Results {
-		result := entries[resultHash]
-		if _, ok := result.(*output); ok {
-			hashes.OutputIDs[i] = bc.Hash(resultHash)
-		}
+		hashes.OutputIDs[i] = bc.Hash(resultHash)
```
bobg: Since `hashes.OutputIDs` is no longer just outputs but all results, please rename it to `ResultHashes`.

erykwalder:
updated, PTAL

protocol/bc/txhashes.go
```diff
@@ -3,9 +3,11 @@ package bc
 type (
 	// TxHashes holds data needed for validation and state updates.
 	TxHashes struct {
-		ID        Hash
-		OutputIDs []Hash // each OutputID is also the corresponding UnspentID
-		Issuances []struct {
+		ID Hash
+		// contains OutputIDs and retirment hashes.
```
bobg: spelling - retirement

bobg:
LGTM with one spelling nit

PR #540, opened 2017-02-16
docs/protocol: remove transaction witness hash

danrobinson:
This removes the transaction witness hash—which is no longer committed to the blockchain and no longer needs to be computed—from the spec.


danrobinson:
PTAL @oleganza @kr @bobg

oleganza:
LGTM.

PR #542, opened 2017-02-16
sdk/ruby: deprecate superfluous tx input fields

jeffomatic:
Neither of control_program and input_witness should be a property
of Transaction::Input.


jeffomatic:
PTAL

PR #544, opened 2017-02-17
core/query: make annotated output's txid truly omittable

jeffomatic:
AnnotatedOutput'd TransactionID field is a bc.Hash, which does not have
a zero-value that will trigger the omitempty directive when serialized
to JSON. We use pointers instead to get the desired behavior.

This commit is a stopgap measure to prevent some weird behavior in the
Node SDK; namely, transaction output objects are currently returned as
strings of zeros. In the future, we can re-evaluate the API spec and decide
whether we should allow the transaction ID field to be empty in any context.


jeffomatic:
PTAL

kr:
LGTM

PR #545, opened 2017-02-17
protocol/patricia: remove unused parameter

kr:
There's no need for insert, delete, and lookup to be
methods on Tree. The *Tree is completely unused in the
body of those three functions (other than in the
recursive calls).


kr:
PTAL

bobg:
LGTM

PR #546, opened 2017-02-17
protocol/patricia: remove error case in Delete

kr:
Delete used to return ErrPrefix in one case: when the
requested key exactly matches an interior node. There
were three problems with this:

First, it didn't cover all cases where the requested key
is a prefix of an existing item. Consider the following
tree of three items (001, 010, and 011):

    0             (interior node)
    +-- 001       (leaf node)
    +-- 01        (interior node)
        +--  010  (leaf node)
        +--  011  (leaf node)

Attempting to delete 01 produced ErrPrefix, but
attempting to delete 00 was a no-op, returning a nil
error. Both of those keys are prefixes of other keys in
the set.

Second, there was a bug in the recursive call. It
returned a nil node on error. This returned all the way
up the stack and caused the root node to be replaced
with nil. Thus, attempting to delete a nonexistent key
corresponding to an interior node would actually delete
the entire tree.

Third, and most broadly, attempting to delete a prefix
of an existing key (whether or not it matches an
interior node) shouldn't be an error. It should be a
no-op, just like trying to delete any other key that's
not in the set.

This patch addresses these issues by removing the error
condition and treating that case as a no-op.

There's no significant difference in performance in the
existing benchmarks.

benchmark                      old ns/op     new ns/op     delta
BenchmarkInserts-4             27923330      27744174      -0.64%
BenchmarkInsertsRootHash-4     37095589      37046587      -0.13%

benchmark                      old allocs     new allocs     delta
BenchmarkInserts-4             161756         161755         -0.00%
BenchmarkInsertsRootHash-4     171776         171776         +0.00%

benchmark                      old bytes     new bytes     delta
BenchmarkInserts-4             11642978      11642854      -0.00%
BenchmarkInsertsRootHash-4     11970751      11970739      -0.00%

### DIFF
--- a/core/txdb/snapshot_test.go
+++ b/core/txdb/snapshot_test.go
@@ -65,10 +65,7 @@ func TestReadWriteStateSnapshot(t *testing.T) {
 			}
 		}
 		for _, key := range changeset.deletes {
-			err := snapshot.Tree.Delete(key[:])
-			if err != nil {
-				t.Fatal(err)
-			}
+			snapshot.Tree.Delete(key[:])
 		}
 
 		err := storeStateSnapshot(ctx, dbtx, snapshot, uint64(i))
--- a/protocol/patricia/patricia.go
+++ b/protocol/patricia/patricia.go
@@ -19,7 +19,7 @@ import (
 	"chain/protocol/bc"
 )
 
-// ErrPrefix is returned from Insert or Delete if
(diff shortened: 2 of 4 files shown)


kr:
PTAL

kr:
Sorry about the formatting of the benchmark comparison. It looks okay in the actual commit message.

PR #548, opened 2017-02-18
protocol/patricia: un-export ErrPrefix

kr:
No other package uses the value of ErrPrefix, so we can
reduce the surface area of this package by not exporting it.


kr:
PTAL

PR #550, opened 2017-02-21
protocol/tx: store refdatahashes where pointers to data entries used to go

bobg:


bobg:
PTAL

kr:
:eyes:

protocol/tx/header.go
```diff
@@ -1,10 +1,12 @@
 package tx
 
+import "chain/protocol/bc"
+
 type header struct {
 	body struct {
 		Version              uint64
 		Results              []entryRef
-		Data                 entryRef
+		RefDataHash          bc.Hash
```
kr: The name RefDataHash seems redundant for these fields. We can see that it's a hash right in the declaration. (The validator actually doesn't care that it's a hash of anything -- it's really just 32 arbitrary bytes.) So I would prefer RefData or even just Data.

bobg: OK, done.

PR #552, opened 2017-02-21
cmd/cored: use latestVersion constant only for dev builds

kr:
This slightly simplifies the logic for initializing the
version string on startup. (It shouldn't be complicated!)
Notably, this means that the release version string is
derived entirely from the build tag, and it doesn't use
the latestVersion constant at all -- this way, release
version strings are consistently derived from a single
source of truth, the git tag.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -86,13 +86,10 @@ var (
 
 func init() {
 	var version string
-	if strings.HasPrefix(buildTag, "cmd.cored-") {
-		// build tag with cmd.cored- prefix indicates official release
-		version = latestVersion
-	} else if buildTag != "?" {
-		version = latestVersion + "-" + buildTag
+	if buildTag != "?" {
+		version = strings.TrimPrefix(buildTag, "cmd.cored-")
 	} else {
-		// -dev suffix indicates intermediate, non-release build
+		// +changes suffix indicates non-release build
 		version = latestVersion + "+changes"
 	}
 


kr:
PTAL especially @jeffomatic, what do you think? I feel like the code might be a bit easier to follow this way.

jeffomatic:
Just my two cents:

1. I do have a slight hesitation: this removes a forcing function for updating the `latestVersion` constant. In the current scheme, we must update the constant in order for the version to materialize correctly in the API response. This makes the version change concrete in the commit history, which IIRC was the main reason to use a constant instead of relying on Git tags as the source of version truth. If we land this PR, we lose the forcing function, and the only motivation for updating `latestVersion` is to ensure that intermediate dev builds show the right number, which makes it easy to forget about.

    As a secondary concern, we use hardcoded versions as the source of version truth in other packages such as the SDKs, and it would be nice to remain consistent with that.

    A tertiary concern is that custom build tags that don't include a version string will not report the nearest release version via the API. I think we hardly do this (it's tertiary!), but you never know.

2. If we do land this PR, it's probably still worthwhile to retain a comment that explains why `cmd.cored-` prefixes are stripped; namely, the prefixes represent tags marking formal releases.

kr:
Yeah I see those points.

> I do have a slight hesitation: this removes a forcing function for updating the `latestVersion` constant. In the current scheme, we must update the constant in order for the version to materialize correctly in the API response. This makes the version change concrete in the commit history, which IIRC was the main reason to use a constant instead of relying on Git tags as the source of version truth. If we land this PR, we lose the forcing function, and the only motivation for updating `latestVersion` is to ensure that intermediate dev builds show the right number, which makes it easy to forget about.

Understood. This actually came up earlier today. I needed to cut a 1.1rc4 release under time pressure, so I tried to make a tag and build the release. Under the current scheme, we actually have the opposite problem, which I got bit by today: I made a tag and built the binary, but the version string came out wrong, because it was derived from the hard-coded constant rather than the build tag. It's lucky that I ran the resulting binary to double-check the version string, or I would have shipped a broken release (a cored with an incorrect version string). Storing the canonical version string in only one place (wherever it may be) would solve that problem.

I don't fully understand why "makes the version change concrete in the commit history" — as opposed to merely putting it in the git repository — should be a goal. Both memcached and beanstalkd have been using git tags for this for almost ten years and it works pretty well. Sure, we *can* put the version string in source code… but it's more work that way.

> As a secondary concern, we use hardcoded versions as the source of version truth in other packages such as the SDKs, and it would be nice to remain consistent with that.

I agree this would be nice to do, all else being equal, but it's lower on my list. Some languages might have constraints that force us to store the version string in source code, but that doesn't mean we should do it in all languages if we can find a better option for some of them.

> A tertiary concern is that custom build tags that don't include a version string will not report the nearest release version via the API. I think we hardly do this (it's tertiary!), but you never know.

In that case, I would say it's the responsibility of such a "custom build tag" to include any necessary version information. It's hard to really reason about this scenario because it seems hypothetical.

The hard-coded latestVersion const is a last-ditch effort to present something meaningful when all else fails. At least it's better than nothing, for the case where the user has compiled with 'go install'. IIRC that was the reason we introduced latestVersion. I think it'd be better to keep its scope limited to solving that problem, and use git tags for the actual release version string (since, IMO, they are better suited to it). In fact, to address the 'bare go install' case, perhaps even more thoroughly than the current scheme, maybe we should consider having chainbot generate some sort of unique ID for every single commit. That would remove the need to manually update latestVersion entirely (and provide better precision than the current scheme). I'll do a little bit of investigation on that option before I continue with this PR.

kr:
Closing this in favor of #760.

PR #556, opened 2017-02-22
sdk/windows: poll pg_isready for successful exit code

dominic:


dominic:
@jeffomatic PTAL

Will build a new Windows release in the morning to test this.

jeffomatic:
If not now, then eventually, we should try to figure out how to set the locale using `initdb` so that logfiles are written in English. I just ran the default `initdb` command from a Chinese-language Windows install, and it uses UTF8 encoding, but uses the system language (`Chinese (Simplified)_China.936`). It'll help future debugging if we don't have to run the logfile through Google Translate.

PR #558, opened 2017-02-22
	cmd/slashland: generate a "rev id" for each commit on main

kr:
Right now, when someone clones the repo and builds using 'go install chain/cmd/cored', the resulting binary will have some minimal version information: the latestVersion constant containing the version string of the last release we made. This comes in handy when someone is reporting trouble and we ask them to run `cored -version` to investigate. But this is only a little better than nothing; there are typically hundreds of commits between releases. Adding a unique string to every single revision (on main) will give us a much more precise idea of what code is being run.

We generate source files for our four main languages: Go, Java, JavaScript, and Ruby. It won't be difficult to add others if we need them.

This PR only begins recording the revision ID. We'll incorporate it into the version output in followup work.

### DIFF
--- a/cmd/slashland/main.go
+++ b/cmd/slashland/main.go
@@ -184,6 +184,17 @@ func land(req *landReq) {
 		return
 	}
 
+	err = commitRevIDs(landdir)
+	if err != nil {
+		sayf("<@%s|%s> failed to land %s: could not commit revision id: %s",
+			req.userID,
+			req.userName,
+			req.ref,
+			err,
+		)
+		return
+	}
+
 	cmd = dirCmd(landdir, "git", "push", "origin", req.ref, "-f")
 	cmd.Stderr = os.Stderr
 	err = cmd.Run()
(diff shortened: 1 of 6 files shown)


kr:
(not ptal; wip)

kr:
PTAL

Note the test failure is due to "tktk" comments. I'd appreciate some help with those.

jeffomatic:
Before I dive too deep here, how would the SDKs surface the revision? Or is it just some value that is programmatically available to the SDKs, if they want it?

kr:
> Before I dive too deep here, how would the SDKs surface the revision? Or is it just some value that is programmatically available to the SDKs, if they want it?

Good question, I haven't fully thought it through for the SDKs but I imagine we might want something similar to what I'm shooting for in cored: this string can be used as a fallback version string when the SDK is built directly from git instead of being a release build. So a release might have a version string of `1.1.3` but a build that's on some random commit in between releases would have version string `rev123` (perhaps with some explicit adornment like `dev-` or `git-` or whatever if we think that would be helpful). If that sounds good in principle, we can figure out some way to wire it up.

Or the SDKs could totally ignore these generated files and maybe we'd only use it in cored.

kr:
This is still not perfect, since the value would only change when something lands on main. When working on a branch, the `rev123` would stay the same. We might actually be able to do substantially better with build systems that let us call out to shell commands during the build, then we can query git dynamically during the build instead of relying completely on generated code. Like maybe the maven build tool can just shell out to `git rev-parse HEAD` etc.

kr:
LGTM. I'm also all ears for other ideas to address the underlying problem.

jbowens:
code lgtm

PR #562, opened 2017-02-22
protocol/patricia: remove func Copy

kr:
The Copy function merely made a copy of the Tree
structure. Many Go structures, such as http.Request and
bytes.Buffer, are safe to copy, and patricia.Tree is
too. (In fact, one can typically assume that a Go struct
is okay to copy unless its documentation says otherwise.
See sync.Mutex for an example.) We don't need a function
just to perform this basic operation.


kr:
PTAL

bobg:
LGTM

PR #563, opened 2017-02-22
docs/protocol/papers: add types to examples in blockchain programs paper

danrobinson:
This PR adds type annotations to the examples in the Ivy paper, which makes the paper more readable and brings it closer to our current design for the language.

### DIFF
--- a/docs/protocol/papers/blockchain-programs.md
+++ b/docs/protocol/papers/blockchain-programs.md
@@ -185,8 +185,8 @@ Ivy syntax brings additional structure to programs, making it easy to construct 
 
 Here is an example of a control program written in Ivy:
 
-	program SingleKeyProgram(publicKey) {
-		path spend(signature) {
+	program SingleKeyProgram(publicKey: PublicKey) {
+		path spend(signature: Signature) {
 			verify checksig(publicKey, tx.hash, signature)
 		}
 	}
@@ -194,6 +194,7 @@ Here is an example of a control program written in Ivy:
 Let’s break this program down piece by piece.
 
 * Programs can have **parameters**. This program has one parameter, `publicKey`. Values for a program’s parameters, called **arguments**, are specified at the time the program is *instantiated*, or created. In the case of a control program like this, that is the time that an unspent output is added to the blockchain state by a transaction.
+* Parameters have **types**—such as PublicKey and Signature. These types help the compiler catch bugs in your programs, and make the code more readable.
 * Programs define one or more **paths**. This program has only one path: `spend`. If this control program could be satisfied in different ways, it would have more than one path.
 * Each path can define its own parameters. Arguments for path parameters are provided in the input witness. These arguments are passed — and the path is chosen — at the time the program is executed. In the case of a control program like this, that is the time the unspent output is used as an input in a new transaction. This program takes one argument: a `signature`.
 * Paths contain one or more **conditions**. This path only uses a single condition, which uses the `CHECKSIG` instruction to check that the provided signature on the hash of the new transaction corresponds to the previously specified public key.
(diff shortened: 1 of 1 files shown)


danrobinson:
PTAL @kr @bobg @tonychain @oleganza

docs/protocol/papers/blockchain-programs.md
```diff
@@ -209,20 +210,20 @@ When the program is instantiated with a `publicKey` value — to be used in an o
...
+2. The program then executes, first pushing the public key and then the transaction hash to the stack.
+3. The public key and transaction hash are then swapped to put them in the correct order for the following `CHECKSIG` instruction.
+4. `CHECKSIG` pops all three items off the stack to check the signature, pushing `true` or `false` to the stack.
 5. `VERIFY` then pops the top value from the stack and causes the program to fail if the value is `false`. (In an actual control program, the `VERIFY` instruction of the last condition in a path is omitted, since it is performed by the VM itself.)
 
 Many control, issuance, and consensus programs use a multisignature check.
 
-	program MultiKeyProgram(n, m, publicKeys[n]) {
-		path spend(signatures[m]) {
+	program MultiKeyProgram(n: Number, m: Number, publicKeys: PublicKey[n]) {
```
tonychain: While this gets "m of n" across, it seems a bit awkward to me as actual syntax and unclear how it would map to an actual type system.

The KISS alternative I'd suggest is to just pass an array, which implicitly carries with it an `n`, e.g.:

`program MultiKeyProgram(threshold: Number, publicKeys: PublicKey[])`

Beyond that you're trying to parameterize the array type with additional type information, which feels like something that can't be done cleanly without something like a dependent type system (or at least a weak form of one... [this particular problem comes up frequently in Rust](https://github.com/rust-lang/rfcs/pull/1657/files))

docs/protocol/papers/blockchain-programs.md
```diff
@@ -453,16 +454,16 @@ While most state should be tracked locally in the program-level arguments for a
 
 First, one needs to create an asset for which only one unit can ever be issued. This requires some understanding of how the Chain Protocol handles issuances. Unique issuance — ensuring that issuances cannot be replayed — is a challenging problem that is outside the scope of this paper. The Chain Protocol’s solution is that each issuance input has a nonce that, when combined with the transaction’s `mintime`, `maxtime`, and asset ID, must be unique throughout the blockchain’s history. As a result, an issuance *program* can ensure that it is only used once by committing to a specific nonce, transaction mintime, and transaction maxtime:
 
-    program SinglyIssuableAssetSingletonToken(nonce, mintime, maxtime, amount, lockProgram) {
-    	path issue(outputIndex) {
+    program SinglyIssuableAssetSingletonToken(nonce: String, mintime: Time, maxtime: Time, amount: Number, lockProgram: Program) {
```
tonychain: Nit: I generally prefer the "not before" and "not after" terms (as popularized by e.g. X.509) for `mintime`/`maxtime` here

tonychain:
I think type information dramatically improves the clarity of these examples 👍

docs/protocol/papers/blockchain-programs.md
danrobinson: Mintime and maxtime are the words used throughout the protocol; maybe worth discussing (though out of scope of this PR)

docs/protocol/papers/blockchain-programs.md
danrobinson: You're right that this is a form of dependent typing. It's a hint that's unfortunately necessary given the way lists are currently represented (as multiple stack items rather than a single list). We can't just do an implicit carried `n` because sometimes the length of an array is specified in the program arguments but the array itself is provided in the path arguments (which is true for the list of signatures). But since this is something that we're planning to fix in the VM, maybe we should just do the right thing. Since the previous solution (`publicKeys[m]`) is approximately equally weird, could we land this and fix this problem in a separate PR if we come up with something?

kr:
LGTM, except that there is quite a bit of unrelated noise from whitespace changes, to the point that it gets in the way (I'm leaning on ?w=1 here). It would be nice IMO to do those cleanups in a separate "fix whitespace" commit.

danrobinson:
@kr Weird, I don't know how that happened or how to separate them out.

docs/protocol/papers/blockchain-programs.md
tonychain: > We can't just do an implicit carried n because sometimes the length of an array is specified in the program arguments but the array itself is provided in the path arguments

Won't `n` always be known in advance though, or failing that, not particularly relevant so long as the threshold is reached?

> But since this is something that we're planning to fix in the VM, maybe we should just do the right thing.

👍

kr:
@danrobinson I'm guessing a text editor function that automatically removes trailing whitespace on each line? I can help by making a whitespace-only PR, and then if you rebase this on top of that it should take care of it.

danrobinson:
Ah I think I can do that.

kr:
Once #564 lands, rebasing this on main should clear it up.

PR #564, opened 2017-02-22
docs/protocol/papers: remove trailing whitespace

kr:


kr:
PTAL @danrobinson

danrobinson:
LGTM

PR #565, opened 2017-02-22
protocol/patricia: restrict Contains to new protocol

kr:
The protocol now defines the Merkle Patricia Tree as a set,
rather than a dictionary[1]. Our usage of this package fits
the definition, but this package's own interface doesn't
match yet.

This patch updates Contains to take only a single "item"
argument, which is used as both the key and the value.
In the "set" definition, there is no distinction between
keys and values -- the same bit string fills both roles.
We also remove ContainsKey since it is no longer necessary.

A future patch will update Insert analogously.

[1]: https://github.com/chain/chain/blob/d60cd85416/docs/protocol/specifications/data.md#merkle-patricia-tree


kr:
PTAL

erykwalder:
:eyes:

protocol/patricia/patricia_test.go
```diff
@@ -192,32 +192,19 @@ func TestLookup(t *testing.T) {
...
 	}
 
-	contains = tr.Contains(bits("11111111"), vals[3])
-	if contains {
-		t.Errorf("expected tree to not contain %v, %x, but did", bits("11111111"), vals[3])
+	if tr.Contains(bits("00000000")) {
+		t.Errorf("expected tree to not contain %x, but did", bits("00000000"))
+	}
+	if tr.Contains(bits("00000010")) {
+		t.Errorf("expected tree to not contain %x, but did", bits("00000000"))
```
erykwalder: Should the %x val be `00000010`?

kr:
Maybe I should remove OutputTreeItem in a separate PR?

protocol/patricia/patricia.go
```diff
@@ -60,29 +60,20 @@ func walk(n *node, walkFn WalkFunc) error {
...
 
-	key := bitKey(bkey)
+	key := bitKey(item)
 	n := lookup(t.root, key)
 
 	var hash bc.Hash
 	h := sha3pool.Get256()
 	h.Write(leafPrefix)
-	h.Write(val[:])
+	h.Write(item)
```
erykwalder: Should we just call this ContainsKey and skip the value look up?

erykwalder:
That might be good

protocol/patricia/patricia.go
kr: I don't want to leave out the value comparison just yet because this package doesn't enforce that keys and values are equal. Once we've updated Insert, then I think that would be safe, right? Also, long run, it should just be "Contains" because it's semantically a set -- there are no keys or values, only items in the set.

kr:
Addressed feedback; PTAL.

erykwalder:
LGTM

PR #566, opened 2017-02-23
cmd/cored: print version string on startup

dominic:
Adds the following output on launch:

```
$ cored
Chain Core starting
* Version VERSION
* [Development|Production] mode

...
```

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -120,21 +120,17 @@ func main() {
 	v := flag.Bool("version", false, "print version information")
 	flag.Parse()
 
-	if *v {
-		fmt.Printf("cored %s\n", config.Version)
-		fmt.Printf("production=%t\n", config.Production)
-		fmt.Printf("build-commit=%v\n", config.BuildCommit)
-		fmt.Printf("build-date=%v\n", config.BuildDate)
-
-		return
+	if !*v {
+		fmt.Printf("Chain Core starting...\n\n")
 	}
 
-	fmt.Printf("Chain Core starting\n")
-	fmt.Printf("* Version %s\n", config.Version)
-	if config.Production {
-		fmt.Printf("* Production mode\n")
-	} else {
-		fmt.Printf("* Development mode\n")
+	fmt.Printf("cored %s\n", config.Version)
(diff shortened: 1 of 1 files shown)


dominic:
PTAL @jeffomatic

kr:
Perhaps we should consider using the same format as the `cored -version` output.

dominic:
@kr Is the `key=value` format important for a use case? Matching formats sounds good; this could be more readable as well:

```
$ cored --version
cored 1.1rc5+changes
production=false
build-commit=?
build-date=?
```

kr:
The -version format is meant to be readable by both humans and machines, and reasonably stable. It also (at least for the first line) follows the GNU coding standards for version output https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html.

I'm not partial to the details of the format, but however it looks, we should make sure that someone can write a program to recover, for example, the build commit without too much fuss. Right now it's easy:

```
$ cored -version | grep build-commit | cut -d= -f2
```

I'd like it to remain about that easy; don't wanna be too anti-social. :)

dominic:
Excellent - that's good info, thanks!

kr:
One other thing that is super minor -- I tend to shy away from the `*` character in CLI output, only because make uses the string `***` to signal an error. But that is *super* minor, and really it's about the three-star string, not so much single stars.

dominic:
@kr updated:

```
Chain Core starting...

cored 1.1rc5+changes
production: false
build-commit: ?
build-date: ?

...
```

Version output matches lines 3-7.

kr:
:eyes:

dominic:
(I'd also really like `cored -v` to work alongside `cored -version`, but the `flag` package seems somewhat limited.)

kr:
Looks pretty good to me. I think getting info out now would be roughly:

```
cored -version | grep build-commit | cut -d: -f2 | cut -c2-
```

which is not too bad. :)

If we change the format here, we should also update the other two commands (in chainprv) that print this format.

dominic:
@kr I wanted to avoid spaces to keep it a similar command length, so this works too:

```
cored -version | grep build-commit | cut -d' ' -f2
```

PR #568, opened 2017-02-23
docs/protocol/specifications: require MPTH input to be prefix-free

kr:
The Merkle Patricia Tree is only valid for a prefix-free
set. It is up to the environment applying this function
to ensure that property. The state commitment does so by
making all input elements of equal length (32 bytes).


kr:
PTAL

danrobinson:
LGTM

PR #569, opened 2017-02-23
docs/core/examples: use output ID instead of txid/position tuple

jeffomatic:
We want to encourage folks to think of outputs as things with scalar unique IDs.


erykwalder:
LGTM

PR #570, opened 2017-02-23
bin/jrun: display deprecation errors

jeffomatic:
This commit adds a compile flag so use of deprecated APIs
will trigger console warnings. Since jrun is used primarily to test
documentation examples, this is a really good thing!


jeffomatic:
PTAL a quickie

erykwalder:
LGTM

PR #571, opened 2017-02-23
core/query: change filter.Bool to filter.String

### LINKED ISSUE #567: querying for local assets results in CH602 error
To reproduce:

1. Open dashboard
2. Create a new asset
3. In the Assets list view, type in the following query: `is_local='yes'`.
4. A CH602 error will appear, even though the query is legal.

erykwalder:
Fixes #567


jbowens:
lgtm

PR #573, opened 2017-02-23
core/query: hide control program from annotated input

### LINKED ISSUE #543: Transaction input API response has superfluous control_program field
Currently, the API returns inputs in a transaction with a control_program field. This has never been a documented property of the API, although it appears in response objects returned by the Node JS SDK (which performs no filtering on the client side), and by the Ruby SDK (a historical mistake; see #542).

Internally, the control_program annotation appears to be important for adding account information from the spent output into the spending input's annotations (see issue 1754 in our internal archival repo), but it's also not something we need or necessarily want to reveal in an API response itself.

It would be simple to just leave this in the API, but I would prefer to either 1) document it, or 2) even better, figure out a way to ensure that it's not included in API responses.

erykwalder:
fixes #543


kr:
LGTM

PR #574, opened 2017-02-23
docs/core: introduce receivers

jeffomatic:
This commit updates the documentation so it uses receivers the first-class cross-core payment object; control programs are thusly deprecated. Both prose and code samples have been updated to educate users about receivers.

### DIFF
--- a/docs/core/examples/java/TransactionBasics.java
+++ b/docs/core/examples/java/TransactionBasics.java
@@ -26,26 +26,27 @@ class TransactionBasics {
     Transaction.submit(client, signedIssuance);
     // endsnippet
 
-    // snippet create-bob-issue-program
-    ControlProgram bobProgram = new ControlProgram.Builder()
-      .controlWithAccountByAlias("bob")
+    // snippet create-bob-issue-receiver
+    Receiver bobIssuanceReceiver = new Account.ReceiverBuilder()
+      .setAccountAlias("bob")
       .create(otherCoreClient);
+    String bobIssuanceReceiverSerialized = bobIssuanceReceiver.toJson();
     // endsnippet
 
-    // snippet issue-to-bob-program
-    Transaction.Template issuanceToProgram = new Transaction.Builder()
+    // snippet issue-to-bob-receiver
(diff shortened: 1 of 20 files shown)


docs/core/examples/java/TransactionBasics.java
```diff
@@ -115,14 +117,20 @@ public static void main(String[] args) throws Exception {
       // endsnippet
     }
 
-    // snippet create-bob-multiasset-program
-    bobProgram = new ControlProgram.Builder()
-      .controlWithAccountByAlias("bob")
+    // snippet create-bob-multiasset-receiver
+    Receiver bobGoldReceiver = new Account.ReceiverBuilder()
```
dominic: Would this be a good place to use `createReceiverBatch` in an example?

docs/core/examples/node/transactionBasics.js
```diff
@@ -183,16 +189,28 @@ Promise.all([
...
+  return (
 
-  // snippet create-bob-multiasset-program
-  otherClient.accounts.createControlProgram({
-    alias: 'bob'
-  }))
-  // endsnippet
+    // snippet create-bob-multiasset-receiver
+    Promise.all([
+      otherClient.accounts.createReceiver({
```
dominic: Same re: batch call

docs/core/examples/ruby/transaction_basics.rb
```diff
@@ -95,23 +105,37 @@
   # endsnippet
 end
 
-# snippet create-bob-multiasset-program
-bob_program = other_core.accounts.create_control_program(
-  alias: 'bob'
-).control_program
+# snippet create-bob-multiasset-receiver
+bob_gold_receiver = other_core.accounts.create_receiver(
```
dominic: Same re: batch call

docs/core/examples/java/TransactionBasics.java
jeffomatic: I thought about that too, but I think the batch syntax is a little too distracting/power-usery for this context. I actually wouldn't recommend using a batch call if you were creating only two receivers.

PR #578, opened 2017-02-23
protocol/state: remove OutputTreeItem

kr:
This function is simple enough to inline, and this also
helps prepare for a future change to the signature of
Insert in package patricia.


kr:
PTAL

bobg:
LGTM

PR #579, opened 2017-02-23
dashboard: proper error message creating object without key

chrisgarvin:


chrisgarvin:
Previous:
image

Updated:
image

PR #584, opened 2017-02-24
protocol/patricia: adopt set semantics (not dict)

kr:
With this change, Insert takes a single argument, the
data item to insert into the set; it no longer takes a
separate key and value. This brings the interface and
behavior up to date with the specification. We update
the package documentation and various parameter names
accordingly.

### DIFF
--- a/core/account/reserve_test.go
+++ b/core/account/reserve_test.go
@@ -49,7 +49,7 @@ func TestCancelReservation(t *testing.T) {
 
 	// Fake the output in the state tree.
 	_, s := c.State()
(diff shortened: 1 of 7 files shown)


kr:
The current MPTH spec is actually not published to chain.com yet, so the link in the package doc is misleading right now. The actual spec can be found at https://github.com/chain/chain/blob/e3d7b45b0f849abe520464ab1a3b04bc816a260d/docs/protocol/specifications/data.md#merkle-patricia-tree.

bobg:
:eyes:

bobg:
LGTM

PR #590, opened 2017-02-24
docs/protocol/specifications: require MPTH input to be prefix-free (redo)

kr:
Commit a7bc90d2cb47fdebed2afd734c45d4285e8406bd
mistakenly edited the Merkle Binary Tree section when it
should have edited the Merkle Patricia Tree section. The
prefix-free requirement applies to the patricia tree
construction; the traditional (binary) Merkle tree has
no such requirement.

The description from the previous commit:

The Merkle Patricia Tree is only valid for a prefix-free
set. It is up to the environment applying this function
to ensure that property. The state commitment does so by
making all input elements of equal length (32 bytes).


kr:
PTAL

PR #592, opened 2017-02-24
sdk/ruby: document input parameters and return types

dominic:


sdk/ruby/lib/chain/balance.rb
```diff
@@ -16,7 +16,7 @@ class Balance < ResponseObject
     attrib :sum_by
 
     class ClientModule < Chain::ClientModule
-      # @param [Hash] query
+      # @param [QueryOpts || Hash] query
```
jeffomatic: Balance also accepts a `sum_by` parameter.

sdk/ruby/lib/chain/batch_response.rb
```diff
@@ -1,9 +1,3 @@
-def ensure_key_sorting(h)
```
jeffomatic: NBD, but would help to put this type of thing in a separate PR. I know it's a bit more work, but in the long-run we'll get a tighter commit history.

sdk/ruby/lib/chain/transaction.rb
```diff
@@ -85,16 +88,17 @@ def submit(template)
...
+      # @return [BatchResponse<SubmitResponse>]
       def submit_batch(templates)
         client.conn.batch_request(
           'submit-transaction',
           {transactions: templates}
         ) { |item| SubmitResponse.new(item) }
       end
 
-      # @param [Hash] query
+      # @param [QueryOpts || Hash] query
```
jeffomatic: Transaction queries have a few additional options that should be exposed to the user:

- `timeout`
- `start_time`
- `end_time`

For details, see https://github.com/chain/chain/blob/main/docs/internal/api-spec.swagger.yml#L510

sdk/ruby/lib/chain/transaction_feed.rb
```diff
@@ -104,6 +109,7 @@ def delete(opts)
         nil
       end
 
+      # @param [QueryOpts || Hash] query
```
jeffomatic: We may want to make it parameterizable so it accepts `page_size`. According to this it does:

https://github.com/chain/chain/blob/main/docs/internal/api-spec.swagger.yml#L998

sdk/ruby/lib/chain/unspent_output.rb
```diff
@@ -73,8 +73,8 @@ class UnspentOutput < ResponseObject
     attrib :is_local
 
     class ClientModule < Chain::ClientModule
-      # @param [Hash] query
-      # @return Query
+      # @param [QueryOpts || Hash] query
```
jeffomatic: Unspents have an additional `timestamp` param: https://github.com/chain/chain/blob/main/docs/internal/api-spec.swagger.yml#L944

sdk/ruby/lib/chain/account.rb
```diff
@@ -79,15 +83,17 @@ def create_receiver(opts)
...
-      # @return [BatchResponse]
+      # @return [BatchResponse<Receiver>]
       def create_receiver_batch(opts_list)
         client.conn.batch_request('create-account-receiver', opts_list) { |item| Receiver.new(item) }
       end
 
-      # @param [Hash] query
+      # @param [Hash] opts Filtering information
+      # @option opts [String] filter Filter string, see {https://chain.com/docs/core/build-applications/queries}
+      # @option opts [Array<String>] filter_params Parameter values for filter string (if needed)
```
jeffomatic: You can do arrays of integers or strings. Dunno if you can do `Array<String|Integer>`, but `Array<Object>` is okay too.

PR #593, opened 2017-02-24
core: update info endpoint response with build info

jeffomatic:


dominic:
LGTM

kr:
Hey, I see this just landed, but we should prob try to stay consistent with the formatting of the existing places where we return this info. See also chain/chainprv#2490.

PR #595, opened 2017-02-24
all: update build info to new output format

kr:


kr:
PTAL

kr:
See https://github.com/chain/chainprv/pull/2490 and https://github.com/chain/chain/pull/593.

PR #611, opened 2017-02-27
protocol/bc: remove OutputID type

erykwalder:
OutputID was just a simple wrapper for bc.Hash. Given that each entry
type will have its own reference ID, there is no need for OutputID to
have its own distinguished type.


kr:
LGTM

PR #612, opened 2017-02-27
core/account: add expiration block processor

erykwalder:
This runs the expiration of account control programs as a separate
process. This enables the expiration to wait until the transaction
processor has completed before deleting any data that may be needed.


core/account/indexer.go
```diff
@@ -89,18 +94,16 @@ func (m *Manager) ProcessBlocks(ctx context.Context) {
...
+	m.pinStore.ProcessBlocks(ctx, m.chain, PinName, m.indexAccountUTXOs)
 }
 
-func (m *Manager) processBlock(ctx context.Context, b *bc.Block) error {
-	err := m.indexAccountUTXOs(ctx, b)
-	if err != nil {
-		return err
-	}
+func (m *Manager) expireControlPrograms(ctx context.Context, b *bc.Block) error {
+	<-m.pinStore.PinWaiter(query.TxPinName, b.Height)
```
kr: Should this also wait on the indexer pin in this package (`PinName`)?

erykwalder:
updated, PTAL

erykwalder:
Another option I'm realizing, is to just call a goroutine for each call to the block processor.

kr:
> Another option I'm realizing, is to just call a goroutine for each call to the block processor.

Oh interesting, what would that look like? I don't totally follow.

kr:
LGTM

PR #613, opened 2017-02-27
protocol/validation: pass only the pieces of state needed to ConfirmTx, not a whole block

bobg:
This keeps us from having to pass an incompletely constructed block to ConfirmTx from Chain.GenerateBlock.

### DIFF
--- a/protocol/block.go
+++ b/protocol/block.go
@@ -80,7 +80,7 @@ func (c *Chain) GenerateBlock(ctx context.Context, prev *bc.Block, snapshot *sta
 			continue
 		}
 
-		if validation.ConfirmTx(result, c.InitialBlockHash, b, tx) == nil {
+		if validation.ConfirmTx(result, c.InitialBlockHash, bc.NewBlockVersion, timestampMS, tx) == nil {
 			err = validation.ApplyTx(result, tx)
 			if err != nil {
 				return nil, nil, err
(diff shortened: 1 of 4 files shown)


bobg:
PTAL

protocol/block.go
```diff
@@ -80,7 +80,7 @@ func (c *Chain) GenerateBlock(ctx context.Context, prev *bc.Block, snapshot *sta
 			continue
 		}
 
-		if validation.ConfirmTx(result, c.InitialBlockHash, b, tx) == nil {
+		if validation.ConfirmTx(result, c.InitialBlockHash, bc.NewBlockVersion, timestampMS, tx) == nil {
```
erykwalder: maybe rather than doubling up values, use the ones on `b`?

erykwalder:
LGTM with a minor nit

protocol/block.go
bobg: When constructing an object like `b`, my strong preference is for it to be write-only until fully constructed. I'm sure this practice has some pithy software-engineering name that I don't know.

(An obvious exception here is `len(b.Transactions)`, which we could avoid reading by building a separate `transactions` slice and then assigning the complete thing to `b.Transactions` after the loop, but that's overkill here. [However, I did do that very thing [on the `entries`branch](https://github.com/chain/chain/commit/5e7f06e55b3911c2f696569e3c25cf613ddbd355#diff-b0c0b1bf19b48c8df1f778305584481eR62).])

PR #614, opened 2017-02-27
docker/ci: Add Rust tools to Dockerfile

tonychain:
Installs Rust nightly (locked to 2017-02-25 snapshot) via Rustup:

https://www.rustup.rs/

Also installs the following development tools:

- rustfmt: Lint syntax to ensure it follows standard style (ala gofmt)
- clippy: Lint for non-idiomatic Rust or other minor mistakes
- cargo-audit: Audit Cargo.lock for crates containing security vulnerabilities

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -20,6 +20,10 @@ RUN apk --no-cache add bash clang git g++ make nodejs openjdk8 python perl \
     && cd $(go env GOPATH)/src/github.com/GoASTScanner/gas \
     && git reset --hard d30c5cde3613e9ba0129febda849e4d4df1d57cd \
     && go install github.com/GoASTScanner/gas
+    && curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2017-02-25
+    && ~/.cargo/bin/cargo install rustfmt
+    && ~/.cargo/bin/cargo install clippy
+    && ~/.cargo/bin/cargo install cargo-audit
 
 ADD bin /usr/bin
 ENV CHAIN ${GOPATH:-$HOME/go}/src/chain


docker/ci/Dockerfile
```diff
@@ -20,6 +20,10 @@ RUN apk --no-cache add bash clang git g++ make nodejs openjdk8 python perl \
     && cd $(go env GOPATH)/src/github.com/GoASTScanner/gas \
     && git reset --hard d30c5cde3613e9ba0129febda849e4d4df1d57cd \
     && go install github.com/GoASTScanner/gas
+    && curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly
```
tonychain: This installs Rust into the current user's home directory. I'm guessing the current user here is `root`. Perhaps this needs to be changed to install it as the user wercker uses.

kr: Hmm, I don't know if wercker does anything special to run as a different user. I suspect it'll have the same home directory as the dockerfile commands. (But I don't know for sure.)

- - -

Should we specify a particular commit or release of Rust here instead of nightly? As it is, we're going to get an arbitrary nightly snapshot (from whenever we build the docker image) and be using that until we update it again. Since it'll be effectively fixed anyway, maybe we should put here explicitly what version it is?

tonychain: Yeah, we can use a specific nightly snapshot (e.g. `nightly-2017-02-25`). I could also try the beta channel instead of nightly, which might have all of the features we need.

Will start with using a specific nightly snapshot.

tonychain: Updated to use `nightly-2017-02-25`

PR #615, opened 2017-02-27
dashboard: print one-item objects on a single line

dominic:


dominic:
Derivation path:
screen shot 2017-02-27 at 1 13 16 pm

Object with one key:
screen shot 2017-02-27 at 1 13 22 pm

Object with two keys:
screen shot 2017-02-27 at 1 13 18 pm

jeffomatic:
1. Not sure if this will work generically if there are nested objects in asset defs and tags; you may need a more complex check for determining whether the value is truly printable on one line.
2. Would be nice (but not critical) to preserve space between key and value in the case of `{"foo": "bar"}`

PR #618, opened 2017-02-27
protocol/tx: allow entryID to panic rather than return an error

bobg:
Also, recover from the panic inside `mapTx` and turn it back into an error.


bobg:
PTAL

kr:
LGTM

PR #620, opened 2017-02-27
core/account: stop loading outpoints in reserve

erykwalder:
Reserving UTXOs used to fetch the outpoint along with the output ID. The
outpoint is no longer used for anything.


kr:
LGTM

PR #623, opened 2017-02-27
protocol/tx: add EntryRef, a hash-caching smart pointer for referring to entries

bobg:


bobg:
PTAL

bobg:
Note, this is a little different from the `EntryRef` on the `entries` branch: the hash is computed when the `EntryRef` is constructed, which avoids concurrency problems.

erykwalder:
LGTM

protocol/tx/entry.go
```diff
@@ -11,20 +11,44 @@ import (
...
+		// reports the position of that antecedent object within its
+		// transaction. Both inputs (spends and issuances) and outputs
+		// (including retirements) are numbered beginning at zero. Entries
+		// not originating in this way report -1.
+		Ordinal() int
+	}
+
+	// EntryRef holds an entry and its ID. The entry may be nil, but the
+	// ID of a valid entry must still be present.
+	EntryRef struct {
```
kr: I'd suggest not using the `type ( a; b )` form here. Rule of thumb for that is it's suitable for 3 or more one-liner type definitions in a row, or similar. IMO this is more readable as two separate statements.

protocol/tx/entry.go
```diff
@@ -11,20 +11,44 @@ import (
...
+// NewEntryRef produces a new EntryRef with a non-nil entry.
+func NewEntryRef(e entry) *EntryRef {
+	return &EntryRef{
+		entry: e,
+		Hash:  entryID(e),
+	}
+}
+
+// NewIDRef produces an EntryRef with a nil entry.
+func NewIDRef(hash bc.Hash) *EntryRef {
```
kr: Does this need to be exported?

protocol/tx/entry.go
```diff
@@ -11,20 +11,44 @@ import (
...
+	EntryRef struct {
+		entry
+		bc.Hash
+	}
+)
 
 var errInvalidValue = errors.New("invalid value")
 
+// NewEntryRef produces a new EntryRef with a non-nil entry.
+func NewEntryRef(e entry) *EntryRef {
```
kr: Does this need to be exported?

protocol/tx/entry.go
```diff
@@ -73,6 +97,10 @@ func writeForHash(w io.Writer, c interface{}) error {
 	case bc.AssetID:
 		_, err := w.Write(v[:])
 		return errors.Wrap(err, "writing bc.AssetID for hash")
+
+	case *EntryRef:
```
kr: It would be nice to avoid having extra cases in the hashing logic to accommodate this. How do you feel about storing the pointer fields in a separate parallel structure outside the `body` of each entry?

kr: (This isn't a concrete proposal, just an idea that I am wondering if we should consider.)

protocol/tx/entry.go
bobg: It will be. I had intended that the next PR after this one would export a bunch of stuff from this package (in preparation for one following that would move everything into `protocol/bc`, in further preparation for using these types in `protocol/vm`).

I could save the exporting for the last minute before it's needed, or I can anticipate the need, which will keep the later PRs smaller. I don't have too strong a preference about that, LMK if you do.

protocol/tx/entry.go
bobg: There are a couple of problems with this suggestion: first, if the hash lives in an `entry`, then you need special separate handling for the case where there's only a hash and no `entry`. Second, you may recall that we experimented with this while doing the initial implementation of `protocol/tx`: I briefly changed `entry` to a `struct` containing a hash and an `interface` rather than an `interface` itself, and you said you preferred `entry` to be the `interface`. We _could_ keep `entry` as an interface and just add a `Hash` method to it, implementing it identically for every entry type, but that seems strictly worse than having `EntryRef`.

protocol/tx/entry.go
kr: No strong preference here, sounds good to me.

protocol/tx/entry.go
kr: Ahh, I was trying to suggest something quite different. Should we move the pointers — not the hashes — out of the body in each concrete entry type? For example:

```go
type issuance struct {
	body struct {
		Anchor  bc.Hash
		Value   bc.AssetAmount
		Data    bc.Hash
		ExtHash bc.Hash
	}
	ordinal int
	anchor  entry // equiv to body.Anchor
}
```

or

```go
type spend struct {
	body struct {
		SpentOutput bc.Hash // must be an Output entry
		Data        bc.Hash
		ExtHash     bc.Hash
	}
	ordinal     int
	spentOutput *output // equiv to body.SpentOutput
}
```

etc

bobg: Ooh that's interesting. I'm going to give that a try separately before proceeding with this PR.

bobg:
Closing in favor of https://github.com/chain/chain/pull/641

PR #626, opened 2017-02-27
docs/protocol/specifications: Replace draft-irtf-cfrg-eddsa references with RFC 8032

tonychain:
RFC 8032 has finally been published by the IETF, so we can stop citing the draft.


kr:
LGTM

You might want to remove the meta-commentary "Tried to DRY out the Markdown a little bit here too. Let me know if you're not happy with the formatting and we can figure something out." from the PR description before landing this, so it doesn't end up in the commit message.

PR #629, opened 2017-02-28
dashboard: disable OS-autocomplete for all inputs

dominic:


jeffomatic:
Looks good, although some people get real crazy when they do this, e.g.:

http://stackoverflow.com/questions/35513968/disable-auto-correct-in-safari-text-input

Is this the only attribute we need to add?

jeffomatic:
Suspect most of these are mobile-related, but dunno if you had a chance to test out the solution with just `autocomplete`.

dominic:
Hm - I went with the MDN recommendation: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

Let me double-check Safari

dominic:
@jeffomatic Good call - Safari had some weirdness. Updated.

jeffomatic:
I do notice an auto-capitalize option in the OSX keyboard options, so maybe we could to `autoCapitalize='off'` as well? We might be getting to the point where a `{...disableAutoText}` export would be handy.

dominic:
@jeffomatic Updated again. I couldn't get any of my browsers to trigger autocapitalization (despite it being turned on in System Preferences) - might be a mobile-only option, but it also won't hurt anything.

PR #631, opened 2017-02-28
errors: immediately return on nil error

erykwalder:
Previously, nil errors were detected later in the internal wrap
function. However, most of the callers would also perform unnecessary
work before calling wrap. This changes the callers to return early if
the error is nil.


errors/errors.go
```diff
@@ -41,10 +41,6 @@ func Root(e error) error {
 // The argument stackSkip is the number of stack frames to ascend when
 // generating stack straces, where 0 is the caller of wrap.
 func wrap(err error, msg string, stackSkip int) error {
-	if err == nil {
```
kr: Maybe still keep this one to be defensive?

PR #637, opened 2017-02-28
protocol/bc: remove Outpoint

erykwalder:
This removes the Outpoint type, and removes all usages of it in the code
base. This also removes endpoints that make use of outpoints.


kr:
:eyes:

kr:
LGTM

PR #638, opened 2017-03-01
WIP docker: new docs image

dominic:
- [ ] Accept a version input (i.e. `1.1`) and pick the latest of each tag that matches
- [ ] Create `bin/` script that builds and runs this Docker image, uploads to S3 (some behavior in existing scripts)
- [ ] Speed up `git clone` - make sparse checkout + shallow clone work with tags
- [ ] Build Ruby, Go, Java, Nodejs from source instead of `codequest/ruby-golang-java-nodejs` image

more tbd

### DIFF
--- a/docker/docs/startup.sh
+++ b/docker/docs/startup.sh
@@ -1,0 +1,51 @@
+#!/bin/sh
+
+# Clone Chain Core
+git clone https://github.com/chain/chain.git $CHAIN
+cd $CHAIN
+
+# Install md2html
+go install chain/cmd/md2html
+
+# Build chain.com/docs
+git checkout cmd.cored-1.1.0
+cd $CHAIN/docs
+md2html $DOCS_DEST/docs
+
+# Build Java docs
+echo
+echo "Building Java SDK documentation..."
+
+cd $CHAIN/sdk/java
+git checkout sdk.java-1.1.0
+mvn javadoc:javadoc
+
+javadoc_dest_path=$DOCS_DEST/java
+mkdir -p $javadoc_dest_path
+cp -R target/site/apidocs/* $javadoc_dest_path
+
+# Build Ruby docs
+echo
+echo "Building Ruby SDK documentation..."
+
+cd $CHAIN/sdk/ruby
+git checkout sdk.ruby-1.1.0
+bundle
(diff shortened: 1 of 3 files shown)


docker/docs/startup.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
+
+# Clone Chain Core
+git clone https://github.com/chain/chain.git $CHAIN
```
ryandotsmith: You might consider not cloning the repo, but just copying the repo from the local fs.

docker/docs/startup.sh
```diff
@@ -0,0 +1,51 @@
...
+
+cd $CHAIN/sdk/java
+git checkout sdk.java-1.1.0
+mvn javadoc:javadoc
+
+javadoc_dest_path=$DOCS_DEST/java
+mkdir -p $javadoc_dest_path
+cp -R target/site/apidocs/* $javadoc_dest_path
+
+# Build Ruby docs
```
ryandotsmith: Can any of these tasks be parallelized?

ryandotsmith: If you want some suggesting on how to parallelize in bash, lemme know.

docker/docs/startup.sh
dominic: That would sure be faster. I wanted to avoid requiring a local checkout, but that's silly once you're running a script from the repo.

docker/docs/startup.sh
dominic: Advice appreciated :+1:

jeffomatic:
Rebased

jeffomatic:
Some thoughts here:

- mount the host's version of $CHAIN as a git-clone source, and fallback on Github (or no fallback).
- (maybe) assume all doc versions are on the `main` branch. iterate over available version directories and build language-specific documentation accordingly by checking out the respective stable branches

jeffomatic:
I took a brief look at this, and realized we can probably get away with re-using our existing CI image. Thus far, there is nothing that the doc build requires that the CI image can't perform. I suggest we close this PR for now and follow up with shell script that is run from the Docker host.

dominic:
SGTM

PR #640, opened 2017-03-01
docs/internal: use hyphens for all tags/branches

jeffomatic:
For the future, it's easiest to use hyphens for word separators
in tag names, as opposed to a mix of hyphens and periods.


dominic:
:+1:

PR #641, opened 2017-03-01
protocol/tx: add cross-entry pointers in parallel with their hashes

bobg:
The pointers are outside each entry's `body` and `witness`. This is an alternative to the approach in https://github.com/chain/chain/pull/623.

### DIFF
--- a/protocol/tx/header.go
+++ b/protocol/tx/header.go
@@ -13,7 +13,7 @@ type header struct {
 
 	// Results contains (pointers to) the manifested entries for the
 	// items in body.Results.
-	Results []entry
+	Results []entry // each entry is *output or *retirement
 }
 
 func (header) Type() string         { return "txheader" }
--- a/protocol/tx/header.go
+++ b/protocol/tx/header.go
@@ -10,6 +10,10 @@ type header struct {
 		MinTimeMS, MaxTimeMS uint64
 		ExtHash              bc.Hash
 	}
+
+	// Results contains (pointers to) the manifested entries for the
+	// items in body.Results.
+	Results []entry
 }
 
 func (header) Type() string         { return "txheader" }
@@ -17,12 +21,20 @@ func (h *header) Body() interface{} { return h.body }
 
 func (header) Ordinal() int { return -1 }
 
-func newHeader(version uint64, results []bc.Hash, data bc.Hash, minTimeMS, maxTimeMS uint64) *header {
+func newHeader(version uint64, results []entry, data bc.Hash, minTimeMS, maxTimeMS uint64) *header {
 	h := new(header)
 	h.body.Version = version
-	h.body.Results = results
 	h.body.Data = data
 	h.body.MinTimeMS = minTimeMS
 	h.body.MaxTimeMS = maxTimeMS
+
+	if len(results) > 0 {
+		h.Results = results
+		h.body.Results = make([]bc.Hash, 0, len(results))
+		for _, r := range results {
+			h.body.Results = append(h.body.Results, entryID(r))
+		}
+	}
+
 	return h
 }
--- a/protocol/tx/header.go
+++ b/protocol/tx/header.go
@@ -28,12 +28,9 @@ func newHeader(version uint64, results []entry, data bc.Hash, minTimeMS, maxTime
 	h.body.MinTimeMS = minTimeMS
 	h.body.MaxTimeMS = maxTimeMS
 
-	if len(results) > 0 {
-		h.Results = results
-		h.body.Results = make([]bc.Hash, 0, len(results))
-		for _, r := range results {
-			h.body.Results = append(h.body.Results, entryID(r))
-		}
+	h.Results = results
+	for _, r := range results {
+		h.body.Results = append(h.body.Results, entryID(r))
 	}
 
 	return h
--- a/protocol/tx/nonce.go
+++ b/protocol/tx/nonce.go
@@ -22,9 +22,7 @@ func (nonce) Ordinal() int { return -1 }
 func newNonce(p program, tr *timeRange) *nonce {
 	n := new(nonce)
 	n.body.Program = p
-	if tr != nil {
-		n.body.TimeRange = entryID(tr)
-		n.TimeRange = tr
-	}
+	n.body.TimeRange = entryID(tr)
+	n.TimeRange = tr
 	return n
 }
--- a/protocol/tx/nonce.go
+++ b/protocol/tx/nonce.go
@@ -8,6 +8,10 @@ type nonce struct {
 		TimeRange bc.Hash
 		ExtHash   bc.Hash
 	}
+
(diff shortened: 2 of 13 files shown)


bobg:
Unfortunately, transaction hashing has slowed down (almost certainly because of redundant hashing of component entries):

```
BenchmarkHashNonemptyTx-8     25921         46452         +79.21%
```

(The `EntryRef` change in https://github.com/chain/chain/pull/623 also slows things down, but by a lesser amount: 23.65%.)

I think this can be addressed in subsequent commits, but I don't plan to check in this one until I've proven out some optimization ideas. Stay tuned.

kr:
I would argue the extra hashing shouldn't block landing this PR until we've proven it makes a material difference to our overall performance.

protocol/tx/spend.go
```diff
@@ -9,17 +9,43 @@ type spend struct {
...
 	}
 	ordinal int
+
+	// SpentOutput contains (a pointer to) the manifested entry
+	// corresponding to body.SpentOutput. It may be nil, in which case
+	// spentInfo must be non-nil.
+	SpentOutput *output
+
+	// spentInfo contains validation-essential data in case the spent
+	// output object is not present (in SpentOutput).
```
kr: Is this an artifact of the mapping logic? I would sort of expect in an all-txgraph world this case "output object is not present" would not occur.

protocol/tx/header.go
```diff
@@ -10,19 +10,31 @@ type header struct {
 		MinTimeMS, MaxTimeMS uint64
 		ExtHash              bc.Hash
 	}
+
+	// Results contains (pointers to) the manifested entries for the
+	// items in body.Results.
+	Results []entry
```
kr: If there are restrictions on what possible types can be in this field (not necessarily all `entry` types), it could be nice to include an end-of-line comment about it. (Doesn't have to be verbose or elaborate.)

```
Results []entry // output or retirement
```

Same goes for the other places.

protocol/tx/spend.go
bobg: In fact "output object is not present" is the default case, and maybe we should consider doing away with the `SpentOutput` pointer. We can't construct a full-fledged `tx.output` from a `bc.TxInput` because the output commitment in the `TxInput` does not include the prevout's refdata. The same goes for reconstructing outputs when spending utxos from the db.

We could consider building a `tx.output` without populating its refdata, but then we'll have an object whose hash isn't what we're claiming in `tx.spend.body.SpentOutput`, which is a prospect I don't care for.

bobg:
Updated, PTAL

protocol/tx/spend.go
kr: > We can't construct a full-fledged `tx.output` from a `bc.TxInput`

Right ok, that's what I was thinking of when I said "artifact of the mapping logic". Once we've converted everything to txgraph, there will be no bc.TxInput, and the transaction will contain a full copy of the previous tx's output entry right? In that case, I would think we won't need this spentInfo type any more.

bobg: No, that's not right. In an all-txgraph world, we'll still be constructing spends from utxos in the database, and those don't contain refdata (nor should they - unless you really hate the `spentInfo` type, and [like me] you also hate the idea of an incompletely populated `output` object in the `spend`).

kr: I am pretty sure we need the previous output in the transaciton. Even aside from the party who's spending the money, any validating node needs a copy of the prevout in order to validate the tx. In txgraph-land, we've elminiated the "Output Commitment" and replaced with a full "output" entry. I don't see any way for a validator to do its job using just the synthesized spentInfo. How would it get the data to put in there?

@danrobinson?

kr: A validator needs to check that the output matches what's in the merkle tree full-state commitment, and to do that it needs to hash the full output object.

bobg: Oh it's not just refdata we'd have to add to the db. To fully reconstruct a correctly hashing prevout we also need the prevout's source refhash and "position" (i.e., the pointer into the previous transaction's mux).

This is certainly doable. An interim commit on the `entries` branch did exactly this. But is it desirable?

bobg: Resolution: yes it's desirable and is being addressed as part of https://github.com/chain/chain/issues/648

bobg:
rebased

bobg:
rebased

bobg:
Updated, PTAL

bobg:
PTAL

kr:
:eyes:

protocol/tx/nonce.go
```diff
@@ -8,16 +8,23 @@ type nonce struct {
...
 func (n *nonce) Body() interface{} { return n.body }
 
 func (nonce) Ordinal() int { return -1 }
 
-func newNonce(p program, tr bc.Hash) *nonce {
+func newNonce(p program, tr *timeRange) *nonce {
 	n := new(nonce)
 	n.body.Program = p
-	n.body.TimeRange = tr
+	if tr != nil {
```
kr: Hmmmm what do you think about adding a case to entryID to return a zero hash when the entry is nil? Then we might be able to get rid of most of these if statements in the constructors.

protocol/tx/transaction.go
```diff
@@ -25,25 +25,28 @@ func ComputeOutputID(sc *bc.SpendCommitment) (h bc.Hash, err error) {
...
+	o := newOutput(program{VMVersion: sc.VMVersion, Code: sc.ControlProgram}, sc.RefDataHash, 0)
+	o.setSourceID(sc.SourceID, sc.AssetAmount, sc.SourcePosition)
 
 	h = entryID(o)
 	return h, nil
 }
 
 // TxHashes returns all hashes needed for validation and state updates.
 func TxHashes(oldTx *bc.TxData) (hashes *bc.TxHashes, err error) {
+	defer func() {
```
kr: This is a little surprising… I didn't think we would be changing how error handling works in this PR. Is it debugging code?

bobg: Actually no - it's newly needed to make `TestTxRangeErrs` pass in `protocol/validation`.

protocol/tx/nonce.go
bobg: Nice suggestion, done.

protocol/tx/header.go
```diff
@@ -10,19 +10,31 @@ type header struct {
...
 	h := new(header)
 	h.body.Version = version
-	h.body.Results = results
 	h.body.Data = data
 	h.body.MinTimeMS = minTimeMS
 	h.body.MaxTimeMS = maxTimeMS
+
+	if len(results) > 0 {
+		h.Results = results
+		h.body.Results = make([]bc.Hash, 0, len(results))
```
bobg: Done.

bobg:
Updated, PTAL

PR #642, opened 2017-03-01
protocol/tx: add pseudoentry idWrapper for caching hashes

bobg:
This is a proposed follow-on change for https://github.com/chain/chain/pull/641 that brings its performance back in line with that of https://github.com/chain/chain/pull/623 (i.e., with this change plus #641, the tx hashing benchmark slows down by only ~33%).

C.f. https://github.com/chain/chain/pull/641#issuecomment-283419770

### DIFF
--- a/core/txbuilder/finalize_test.go
+++ b/core/txbuilder/finalize_test.go
@@ -255,21 +255,23 @@ func benchGenBlock(b *testing.B) {
 		"00" + // common witness extensible string length
 		"01" + // inputs count
 		"01" + // input 0, asset version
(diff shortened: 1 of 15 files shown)


kr:
This seems like a reasonable optimization strategy (and there might be other strategies to consider), but I am –1 on merging this unless a profile shows that it makes a meaningful difference to the performance of Chain Core. (Last time we measured, cored spends 1–2% of its time under heavy load in hashing; if that's still the case, I would expect this change to produce a negligible change in performance and so not be worth merging.)

PR #644, opened 2017-03-01
core/fetch: bootstrap before creating pins

### LINKED ISSUE #643: core/pin:  processors fail to make progress after booting from snapshot
If we boot from a state snapshot (and don't have all of the blocks from 1 to _n_), the block processors get stuck [here](https://github.com/chain/chain/blob/main/core/pin/pin.go#L209) trying to process a block that we never downloaded.

Seems like the correct solution is to altering the bootstrapping process to download the snapshot and corresponding block, initialize all of the pins to the snapshot's height and then start the block processors. That might require quite a bit of refactoring.

jbowens:
Download and store the bootstrapping snapshot before creating block
processor pins. This ensures block processors will start processing at
the right height.

Maybe at some point the package-level `core/fetch` state and
goroutines should be encapsulated in a new type.

Fixes #643.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -368,6 +368,15 @@ func launchConfiguredCore(ctx context.Context, db *sql.DB, conf *config.Config, 
 	)
 
 	go leader.Run(db, *listenAddr, func(ctx context.Context) {
+		if !conf.IsGenerator {
+			fetch.Init(ctx, remoteGenerator)
+			// If don't have any blocks, bootstrap from the generator's
+			// latest snapshot.
+			if c.Height() == 0 {
+				fetch.BootstrapSnapshot(ctx, c, remoteGenerator, fetchhealth)
+			}
+		}
+
 		// This process just became leader, so it's responsible
 		// for recovering after the previous leader's exit.
 		recoveredBlock, recoveredSnapshot, err := c.Recover(ctx)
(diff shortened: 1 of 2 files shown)


jbowens:
PTAL

bobg:
Looks GTM but is there a regression test you can include?

jbowens:
> Looks GTM but is there a regression test you can include?

I don't think it's feasible right now. In the future it could be an integration test running against multiple Cores, but I don't think we have the ability to do that yet.

bobg:
LGTM

jbowens:
Is this too invasive for a 1.1.1 fix or should I open another PR against the `chain-core-server-1.1.x` branch with this same fix?

kr:
> Is this too invasive for a 1.1.1 fix or should I open another PR against the chain-core-server-1.1.x branch with this same fix?

I think the main thing to consider is impact. I think of three things a bug needs to be critical:

- no workaround
- high likelihood of hitting it
- causes damage (e.g. data loss, downtime)

A bug that meets all of those criteria would be a good candidate to fix in a point release. Once we've decided a bug needs to be fixed, then we can focus on finding a minimally-invasive way to patch it.

(This is pretty recent thinking, feel free to debate or whatever. It could probably use some refinement.)

How does that sound to you? How would it apply to this case?

kr:
Oh and by "high" likelihood I really meant any significant likelihood. For example I'd consider a 5% chance "high". A one-in-a-million chance would be low. (I don't know exactly where to draw the line. But then our estimates are fuzzy anyway.)

jbowens:
That all sounds good. I definitely think #643 should be considered critical. It makes it impossible to boot add a new Core to a network (testnet or otherwise).

I'm not sure whether this is the minimally-invasive patch we want or not.

PR #645, opened 2017-03-01
dashboard: keep image margins consistent

chrisgarvin:


chrisgarvin:
**Empty list image for assets/accounts/keys/transactions/feeds/tokens:**
image

**Original balances view:**
image

**Modified:**
image

PR #647, opened 2017-03-01
installer/mac: improve startup port detection, 1.1.1

dominic:
Use `lsof` to check for a process listening on a required port instead of 
creating and binding a new socket. The previous implementation was prone
to false negatives when checking if a port was available.

### DIFF
--- a/installer/mac/ChainCore/Config.swift
+++ b/installer/mac/ChainCore/Config.swift
@@ -7,30 +7,27 @@ class Config: NSObject {
 
     /// Checks if the port is in use by another process.
     static func portInUse(_ port: UInt) -> Bool {
-        let sock = socket(PF_INET, SOCK_STREAM, IPPROTO_TCP)
-        if sock <= 0 {
-            return false
+        var output : [String] = []
+
+        let task = Process()
+        task.launchPath = "/usr/sbin/lsof"
+        task.arguments = ["-n", "-i:\(port)"]
+
+        let outpipe = Pipe()
+        task.standardOutput = outpipe
+        task.launch()
+
+        let outdata = outpipe.fileHandleForReading.readDataToEndOfFile()
+        if var string = String(data: outdata, encoding: .utf8) {
+            string = string.trimmingCharacters(in: .newlines)
+            output = string.components(separatedBy: "\n")
         }
 
-        var listenAddress = sockaddr_in()
-        listenAddress.sin_family = UInt8(AF_INET)
-        listenAddress.sin_port = in_port_t(port).bigEndian
-        listenAddress.sin_len = UInt8(MemoryLayout<sockaddr_in>.size)
-        listenAddress.sin_addr.s_addr = inet_addr("127.0.0.1")
+        task.waitUntilExit()
 
-        let bindRes = withUnsafePointer(to: &listenAddress) { (sockaddrPointer: UnsafePointer<sockaddr_in>) in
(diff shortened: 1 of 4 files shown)


jeffomatic:
Can you add a description of the original bug/repro steps to the commit message?

installer/mac/updates/updates.xml
```diff
@@ -6,22 +6,23 @@
...
     <language>en</language>
       <item>
-        <title>Chain Core Developer Edition 1.1.0</title>
+        <title>Chain Core Developer Edition 1.1.1</title>
         <description>
           <![CDATA[
-            <p><b>NOTE:</b> This update will create a fresh blockchain; please backup
-              any blockchain data you wish to preserve.
+            <p><b>NOTE:</b> This update will create a fresh blockchain if you were previously
+              using versions 1.0.x; please backup any blockchain data you wish to preserve.
```
jeffomatic: Nice improvisation here. I wonder what to do about this in general for point releases. Maybe we need a way to stack changelogs if you're jumping up more than one build version.

installer/mac/ChainCore/Config.swift
```diff
@@ -7,30 +7,27 @@ class Config: NSObject {
...
+            string = string.trimmingCharacters(in: .newlines)
+            output = string.components(separatedBy: "\n")
         }
 
-        let bindErr = Darwin.errno
-        close(sock)
+        task.waitUntilExit()
 
-        if bindRes == -1 && bindErr == EADDRINUSE {
+        if (output.count > 0 && output[0] != "") {
```
jeffomatic: Nit: maybe we can change this to:

```
return output.count > 0 && output[0] != ""
```

installer/mac/updates/updates.xml
dominic: I have typically stacked app update release notes with the following format and it works really well:

```
Hey check out these changes:

- Some bug fix

In the previous 1.1 release:

- Keep these release notes going forward until 1.2
- And these too
```

jeffomatic: That makes sense to me.

installer/mac/ChainCore/Config.swift
```diff
@@ -7,30 +7,27 @@ class Config: NSObject {
...
     /// Checks if the port is in use by another process.
     static func portInUse(_ port: UInt) -> Bool {
-        let sock = socket(PF_INET, SOCK_STREAM, IPPROTO_TCP)
-        if sock <= 0 {
-            return false
-        }
+        var output : [String] = []
+
+        let task = Process()
+        task.launchPath = "/usr/sbin/lsof"
```
kr: Would it be better to have cored exit with a particular status code if the port is in use? And then have the app just try to start it and handle the error when it doesn't work?

Just throwing it out as an option.

jeffomatic: Yeah, might make sense in the long run.  But I'm also down to punt on a cored change (and the corresponding point release) in favor of a short-term fix that only concerns the Mac app.

dominic: It would make sense, but we need a consistent solution for Postgres as well. I haven't investigated how `pg_isready` would respond to this.

Another thing to understand in the future is why this code (previously used [in the Postgres.app launcher](https://github.com/PostgresApp/PostgresApp/blob/cd8016c834a4fec26b6e4a449b6428eed9e00bc6/Postgres/Server.swift#L311-L337)) doesn't function appropriately here.

Issue #519, opened 2017-02-13
core/leader: lead func data race


jbowens:
There's a race condition when a cored is leader, gets demoted and becomes leader again. `leader.Run` will spawn several goroutines when cored becomes leader. These goroutines are expected to exit whenever the context is cancelled. If for whatever reason they don't exit immediately (maybe they're stuck in a slow DB call), it's possible for the old leadership and the new leadership goroutines to be running concurrently.

For most of the leadership goroutines, that's okay, but some rely on the fact that only one leadership goroutine should be running when they access data structures. For ex, `core/generator.Generator.Generate` will read/write the `latestBlock` and `latestSnapshot` fields without using a mutex. This is okay as long as there's only one Generator goroutine running at a time.

We can either:

1. fix each leadership goroutine to use mutexes instead of relying on it being the only goroutine of its kind
2. when demoted, use a `sync.WaitGroup` and wait for all of the leadership goroutines to recognize that they've been cancelled and exit before moving on and trying to become leader again

bobg:
3. All cores, leader or not, spawn all singleton goroutines that might be needed, and those activate/deactivate as required.

kr:
This might be a good time to apply the advice, “write code that’s easy to delete, not easy to extend”. Since we have a pretty good feeling that in the longer term the best answer is to avoid “leader”-ish processes entirely, which of these options would be easiest to cut out in the future? (Or doesn't it matter? Maybe this isn't helpful.) (By, ‘avoid “leader”-ish processes’, I mean design the shared data structures in postgres so that it's ok for all or several coreds to make concurrent updates.)

My first impulse is that option 1 would be the least invasive, since maybe there are only one or two goroutines (e.g. generator) that need that kind of synchronization. We can just add comments to the other ones like "there's a race here but it's ok for more than one of these to be running".

But in general I would vote for the one with less code churn (especially this month).

jbowens:
Oof. I started to go down the option 1 route, but it still leaves the race sequence:

1. cored is leader.
2. cored is demoted and cancels its lead contexts.
3. cored re-elected leader
4. api.lead calls Recover, finding the latestBlock and latestSnapshot
5. Generate still hasn't recognized that it's been cancelled and finishes committing a new block, then exits.
6. api.lead's new call to Generate starts but starting from the old latest block and gets stuck trying to commit a different block to latestBlock.height + 1.

I think option 1 combined with #1073 is our best bet.

jbowens:
This is now fixed.

PR #651, opened 2017-03-01
cmd/cored: always run expire reservations loop

jbowens:
In the [spirit of reducing responsibility of leader processes](https://github.com/chain/chain/issues/519#issuecomment-280741917), always
run  the goroutine to expire reservations regardless of whether
the process is leader. It's harmless to run on non-leader
processes. There won't be any reservations to expire.


kr:
LGTM

PR #636, opened 2017-02-28
docker/ci: fix rust installation

kr:

PR #652, opened 2017-03-02
docker/ci: Revert "Add Rust tools to Dockerfile"

kr:
This reverts commit 58c1cd0ee36aa0ee70d3bcfadff9a74ec9dccd71.

Since Alpine uses musl and Rust doesn't easily work with
musl out of the box, it'll take some more work to get
this up and running. See also #636 for some discussion.

So let's back this out for now and we can try again.


kr:
PTAL

boymanjor:
LGTM

PR #656, opened 2017-03-02
core: use expvar.Handler

jbowens:
Use expvar.Handler instead of duplicating the code in expvar. The
expvar.Handler function is new in Go 1.8.


kr:
:eyes:

kr:
LGTM

kr:
N.B. this will force everyone to upgrade their Go installation.

PR #661, opened 2017-03-02
core/fetch: bootstrap before creating pins

### LINKED ISSUE #643: core/pin:  processors fail to make progress after booting from snapshot

jbowens:
Download and store the bootstrapping snapshot before creating block
processor pins. This ensures block processors will start processing at
the right height.

This is the 1.1.x equivalent of #644. If you have any suggestions to
whittle its surface area down farther, let me know.

Fix #643.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -364,6 +364,15 @@ func launchConfiguredCore(ctx context.Context, db *sql.DB, conf *config.Config, 
 	)
 
 	go leader.Run(db, *listenAddr, func(ctx context.Context) {
+		if !conf.IsGenerator {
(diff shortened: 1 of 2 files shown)


kr:
LGTM

PR #665, opened 2017-03-02
protocol/bc: add SpendCommitment type

### LINKED ISSUE #648: Validation trusts the value in a spend without confirming it matches the spent outputid
We used to compute an unspent ID from the values in a spend's prevout "output commitment" and then check it existed in in the state tree. We lost that when we converted to entry-based hashing:

https://github.com/chain/chain/commit/b817512a72b085f4c9440fafd27bbc220713df44#diff-fb46f84aaf51fdc47bacd0fb932c9f26

Now there's no separate UnspentID; the value in the state tree is simply the outputID in the spend. But that outputID needs to be recomputed from the other values in the spend (according to the txgraph mapping algorithm) and the recomputed value must match the spentOutputID in the spend for the spend to be valid. As it is, we're just reading and trusting the spentOutputID.

erykwalder:
Spend TxInputs need to contain all of the data necessary to reconstruct
output ids in order to spend them. Performing the hashing during
validation gives us cryptographic integrity that the spent output data
is correct.

In order to preserve this data, the mapping process was updated to
return additional information which is stored in the account_utxos
table. This data is then populated in a SpendCommitment type on the
TxInput structs.

Fixes #648

### DIFF
--- a/core/migrate/data.go
+++ b/core/migrate/data.go
@@ -42,4 +42,10 @@ var migrations = []migration{
 			DROP COLUMN index;
 		ALTER TABLE account_utxos ADD PRIMARY KEY (output_id);
 	`},
+	{Name: `2017-03-02.0.core.add-output-source-info.sql`, SQL: `
+		ALTER TABLE account_utxos
+			ADD COLUMN source_id bytea NOT NULL,
+			ADD COLUMN source_pos bigint NOT NULL,
+			ADD COLUMN ref_data_hash bytea NOT NULL;
+	`},
 }
--- a/protocol/bc/spend.go
+++ b/protocol/bc/spend.go
@@ -1,13 +1,21 @@
 package bc
 
+import (
+	"fmt"
+	"io"
+
+	"chain/crypto/sha3pool"
+	"chain/encoding/blockchain"
+	"chain/errors"
+)
+
 // SpendInput satisfies the TypedInput interface and represents a spend transaction.
 type SpendInput struct {
 	// Commitment
-	SpentOutputID Hash
-	OutputCommitment
+	SpendCommitment
 
 	// The unconsumed suffix of the output commitment
-	OutputCommitmentSuffix []byte
+	SpendCommitmentSuffix []byte
 
 	// Witness
 	Arguments [][]byte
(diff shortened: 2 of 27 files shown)


bobg:
:eyes:

protocol/bc/txhashes.go
```diff
@@ -4,14 +4,25 @@ type (
...
-		// each OutputID is also the corresponding UnspentID
-		ResultHashes []Hash
-		Issuances    []struct {
+		Results   []ResultInfo
+		Issuances []struct {
 			ID           Hash
 			ExpirationMS uint64
 		}
-		VMContexts []*VMContext // one per old-style Input
+		SpentOutputIDs []Hash
```
bobg: Do the entries in this list correspond 1:1 with the inputs of the oldtx? (And for inputs that aren't spends, are those entries the zero hash?) Please add a comment saying so.

protocol/bc/txhashes.go
```diff
@@ -40,3 +51,5 @@ func (t TxHashes) SigHash(n uint32) Hash {
 // that can compute the hash of a blockheader. It is a variable here
 // to avoid a circular dependency between the bc and tx packages.
 var BlockHeaderHashFunc func(*BlockHeader) Hash
+
+var OutputHash func(*SpendCommitment) (Hash, error)
```
bobg: Please godoc this var.

protocol/bc/spend.go
```diff
@@ -1,40 +1,148 @@
...
+			if err != nil {
+				return errors.Wrap(err, "reading reference data hash")
+			}
+			return nil
+		}
+		return nil
+	})
+}
+
+func (sc *SpendCommitment) Hash(suffix []byte, assetVersion uint64) (spendhash Hash) {
```
bobg: Am I correct in thinking that we don't expect this to get used? That it's only for the !SerPrevout case?

protocol/tx/map.go
```diff
@@ -33,8 +33,13 @@ func mapTx(tx *bc.TxData) (headerID bc.Hash, hdr *header, entryMap map[bc.Hash]e
 
 	for i, inp := range tx.Inputs {
 		if oldSp, ok := inp.TypedInput.(*bc.SpendInput); ok {
+			var oldSpID bc.Hash
+			oldSpID, err = ComputeOutputID(&oldSp.SpendCommitment)
```
bobg: In the absence of distinguishing types for outputIDs and IDs of other kinds of entry, please observe a convention of putting "out" somewhere in the varname when it holds an outputID. As it is, `oldSpID` looks like it could be the ID of a spend entry.

protocol/tx/transaction.go
```diff
@@ -14,6 +14,23 @@ func init() {
...
+		}
+	}()
+	o := newOutput(valueSource{
+		Ref:      sc.SourceID,
+		Value:    sc.AssetAmount,
+		Position: sc.SourcePosition,
+	}, program{VMVersion: sc.VMVersion, Code: sc.ControlProgram}, sc.RefDataHash, 0)
+
+	h = entryID(o)
+	return h, nil
```
bobg: Specifying `nil` here will override the setting of `err` that happens in the `defer` above. A bare `return` should do the trick.

protocol/bc/spend.go
erykwalder: correct

protocol/tx/transaction.go
erykwalder: I think the defer actually takes precedence:
https://play.golang.org/p/yHPkJUHgkn

kr: The defer executes after the return statement, so that you can modify the returned values even when they've been supplied directly to return.

bobg: Oh of course - if `recover()` returns a value then we're in a panic and normal control flow from the `entryID` call to the `return` call didn't happen. Duh, sorry.

kr: That's also true, but note that defer runs after return even when we're not in a panic. Along the lines of @erykwalder's example: https://play.golang.org/p/_qBWbRSA2x

protocol/tx/map.go
```diff
@@ -33,13 +33,13 @@ func mapTx(tx *bc.TxData) (headerID bc.Hash, hdr *header, entryMap map[bc.Hash]e
 
 	for i, inp := range tx.Inputs {
 		if oldSp, ok := inp.TypedInput.(*bc.SpendInput); ok {
-			var oldSpID bc.Hash
-			oldSpID, err = ComputeOutputID(&oldSp.SpendCommitment)
+			var spendOutputID bc.Hash
```
bobg: nit: `spentOutputID` would be better

bobg:
LGTM (with one tiny nit) but please wait for other :eyes: to take a look too

kr:
:eyes:

core/migrate/data.go
```diff
@@ -42,4 +42,10 @@ var migrations = []migration{
 			DROP COLUMN index;
 		ALTER TABLE account_utxos ADD PRIMARY KEY (output_id);
 	`},
+	{Name: `2017-03-02.0.core.add-output-source-info.sql`, SQL: `
+		ALTER TABLE account_utxos
+			ADD COLUMN source_id bytea NOT NULL,
+			ADD COLUMN source_pos bigint NOT NULL,
+			ADD COLUMN ref_data_hash bytea NOT NULL;
```
jbowens: We expect customers to completely drop the database and recreate it before attempting to run a 1.1.1 binary right? I think that's a fine assumption when dealing with a blockchain reset.

Once landed, we should double check that the migration has the same name and hash on the 1.1.1 branch and `main` so that Cores updating from 1.1.1 to 1.2.0 do not try to run this migration again.

kr:
It looks like the base branch for this PR is `main`. It's more important right now to be preparing a fix for the 1.1.x branch. Check out docs/internal/package-tags-branches.md and #670. Maybe also wait until the fallout from #670 is cleaned up before landing this? (And we might have to open a new PR anyway. Is there a way to change the base branch of an open PR?)

bobg:
> Is there a way to change the base branch of an open PR?

Yep, there's a base-branch pull-down menu when you click the Edit button next to the PR title.

PR #671, opened 2017-03-02
protocol: remove generator tx top sorting

erykwalder:
Previously, GenerateBlock would explicitly check if transactions were
top sorted, and if not would sort them. The pool of transactions already
maintains a good-enough top sorting of transactions, so this extra work
can be removed.

Additionally, the removed functions were incorrectly comparing output
ids and transaction ids.

This is 1.1-stable version of #662


bobg:
LGTM

Can we set up this branch to require code review, like `main` does?

PR #672, opened 2017-03-02
sdk/ruby: backport Ruby SDK 1.0.1 to 1.0-stable

jeffomatic:
This should contain all relevant changes when running the following command:

```
git diff cmd.cored-1.0.2...sdk.ruby-1.0.1 sdk/ruby
```


jeffomatic:
I'm going to declare bankruptcy on tracking pre-existing release tags on the 1.0-stable branch. Will replace this PR with one that updates 1.0-stable with all Ruby-related changes leading up to Ruby 1.0.2.

PR #677, opened 2017-03-02
protocol/bc: add SpendCommitment type

erykwalder:
Spend TxInputs need to contain all of the data necessary to reconstruct
output ids in order to spend them. Performing the hashing during
validation gives us cryptographic integrity that the spent output data
is correct.

In order to preserve this data, the mapping process was updated to
return additional information which is stored in the account_utxos
table. This data is then populated in a SpendCommitment type on the
TxInput structs.

1.1-stable version of #665

### DIFF
--- a/core/account/indexer.go
+++ b/core/account/indexer.go
@@ -235,6 +235,7 @@ func (m *Manager) upsertConfirmedAccountOutputs(ctx context.Context, outs []*acc
 		SELECT unnest($1::bytea[]), unnest($2::bigint[]), unnest($3::bytea[]), unnest($4::bytea[]),  unnest($5::bigint[]),
 			   unnest($6::text[]), unnest($7::bigint[]), unnest($8::bytea[]), $9,
 				 unnest($10::bytea[]), unnest($11::bigint[]), unnest($12::bytea[])
+		ON CONFLICT (tx_hash, index) DO NOTHING;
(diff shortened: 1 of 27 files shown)


core/account/indexer.go
```diff
@@ -216,14 +224,17 @@ func (m *Manager) upsertConfirmedAccountOutputs(ctx context.Context, outs []*acc
...
+		refData = append(refData, out.refData[:])
 	}
 
 	const q = `
 		INSERT INTO account_utxos (tx_hash, index, output_id, asset_id, amount, account_id, control_program_index,
-			control_program, confirmed_in)
+			control_program, confirmed_in, source_id, source_pos, ref_data_hash)
 		SELECT unnest($1::bytea[]), unnest($2::bigint[]), unnest($3::bytea[]), unnest($4::bytea[]),  unnest($5::bigint[]),
-			   unnest($6::text[]), unnest($7::bigint[]), unnest($8::bytea[]), $9
-		ON CONFLICT (tx_hash, index) DO NOTHING
```
bobg: Why did we lose the `ON CONFLICT` clause? It's still present in https://github.com/chain/chain/pull/665

erykwalder: Added below

bobg:
LGTM

PR #679, opened 2017-03-02
core/account: add expiration block processor

erykwalder:
This runs the expiration of account control programs as a separate
process. This fixes the expiration to wait until the transaction
processor has completed before deleting any data that may be needed.

1.1-stable version of #612


kr:
LGTM

PR #680, opened 2017-03-02
docs/internal: don't commit to maintaining 1.0-stable

jeffomatic:
It will take a fair amount of legwork to reconstruct 1.0-stable
from our various package release branches, and it may be error-
prone. For now, let's not commit to maintaining that branch, and
hope that updates to 1.0 will be rare.


jeffomatic:
PTAL

erykwalder:
LGTM

PR #681, opened 2017-03-02
all: make non-sdk java code java 7 compliant

boymanjor:
This commit is in preparation for migrating the
existing CI docker images to CentOS. The image
will have Java 7 installed.


boymanjor:
PTAL @jeffomatic

PR #682, opened 2017-03-02
cmd/cored: initialize TLS flag after parsing env

kr:
Func init runs too early for this var to be initialized
(tlsCrt gets written during env.Parse), so config.TLS
was always getting set false.


kr:
PTAL

bobg:
LGTM

PR #683, opened 2017-03-02
database/raft: use new raft package to store core config

tessr:


tessr:
I believe this is ready for review 😁

database/raft/Makefile
```diff
@@ -0,0 +1,26 @@
+
```
bobg: This file looks like it's for testing and debugging (and isn't required for normal building)? Please add a comment at the top that makes that clear.

database/raft/file.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	err = fsync(f)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+	err = f.Close()
+	if err != nil {
+		return errors.Wrap(err)
+	}
+	return errors.Wrap(os.Rename(temp, name))
```
bobg: Nit: the important thing here is the `os.Rename` call; don't bury it inside a call to `errors.Wrap`.

database/raft/fsync_darwin.go
```diff
@@ -0,0 +1,16 @@
...
+
+package raft
+
+import (
+	"os"
+	"syscall"
+)
+
+func fsync(f *os.File) error {
+	_, _, errno := syscall.Syscall(syscall.SYS_FCNTL, f.Fd(), syscall.F_FULLFSYNC, 0)
```
bobg: Huh, can you really not rely on `f.Sync()` on all platforms?

database/raft/state/peer_test.go
```diff
@@ -0,0 +1,23 @@
...
+	expected_s := State{peers: map[uint64]string{}}
+
+	s.RemovePeerAddr(1)
+	if !reflect.DeepEqual(s, expected_s) {
+		t.Errorf("RemovePeerAddr(%d) => %v want %v", 1, s, expected_s)
+	}
+
+}
+
+func TestSetPeerAddr(*testing.T) {
```
bobg: This needs filling in or removal.

database/raft/state/peer_test.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func TestGetPeerAddr(*testing.T) {
```
bobg: This needs filling in or removal.

database/raft/state/state.go
```diff
@@ -0,0 +1,227 @@
...
+	"chain/log"
+)
+
+const nextNodeID = "raft/nextNodeID"
+
+var ErrAlreadyApplied = errors.New("entry already applied")
+
+// State is a general-purpose data store designed to accumulate
+// and apply replicated updates from a raft log.
+// The zero value is an empty State ready to use.
```
bobg: I don't think this is true, because the `map`s will be uninitialized (by either `make` or with a literal).

database/raft/state/state.go
```diff
@@ -0,0 +1,227 @@
...
+// Provisional read operation.
+func (s *State) Get(key string) (value []byte) {
+	return s.state[key]
+}
+
+// Set encodes a set operation setting key to value.
+// The encoded op should be committed to the raft log,
+// then it can be applied with Apply.
+func Set(key string, value []byte) (instruction []byte) {
+	// TODO(kr): make a way to delete things
```
bobg: Is that not `Delete`, below?

cmd/cored/main.go
```diff
@@ -143,6 +147,55 @@ func runServer() {
 	ctx := context.Background()
 	env.Parse()
 
+	raftDir := filepath.Join(*dir, "raft") // TODO(kr): better name for this
```
bobg: Yes, this does need a better name than `DIR`.

database/raft/fsync_darwin.go
tessr: I didn't write this line, so I'll defer to @ameets and @kr :)

cmd/cored/main.go
tessr: Isn't this `DIR/raft` now? But maybe we should add the raft id or something. (Not sure what makes a good directory name...)

database/raft/Makefile
tessr: (done)

database/raft/fsync_darwin.go
kr: I'll defer to the explanation in [Apple's fsync man page](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man2/fsync.2.html):

     Note that while fsync() will flush all data from the host to the drive (i.e. the "permanent storage
     device"), the drive itself may not physically write the data to the platters for quite some time and it
     may be written in an out-of-order sequence.

     Specifically, if the drive loses power or the OS crashes, the application may find that only some or
     none of their data was written.  The disk drive may also re-order the data so that later writes may be
     present, while earlier writes are not.

     This is not a theoretical edge case.  This scenario is easily reproduced with real world workloads and
     drive power failures.

     For applications that require tighter guarantees about the integrity of their data, Mac OS X provides
     the F_FULLFSYNC fcntl.  The F_FULLFSYNC fcntl asks the drive to flush all buffered data to permanent
     storage.  Applications, such as databases, that require a strict ordering of writes should use F_FULLF-
     SYNC to ensure that their data is written in the order they expect.  Please see fcntl(2) for more
     detail.

(Personally I think it would be better if the default were the safe way perhaps with a flag to turn it off, and even then the flag should arguably be a system-level option rather than an API call parameter. But it is what it is.)

Linux [until about 2008 it seems](https://www.mail-archive.com/linux-fsdevel@vger.kernel.org/msg12025.html) didn't provide the safer option at all, and now I think it only provides the safer option? It's hard to get a straight answer out of the Internet.

Basically I tried to use the safest available option that I'm aware of on each platform. Incidentally, etcd itself does the same thing https://github.com/coreos/etcd/blob/2724c3946eb2f3def5ed38a127be982b62c81779/pkg/fileutil/sync_darwin.go.

kr: This might be worth adding a note in the comments to explain the situation.

tessr: Will do.

tessr: (done)

tessr:
This is ready for another round.

bobg:
:eyes:

Note, the PR title should change to say `database/raft` instead of `core/raft`.

tessr:
Good catch, thanks!

core/api.go
```diff
@@ -97,6 +98,7 @@ func (a *API) needConfig() func(f interface{}) http.Handler {
 	return jsonHandler
 }
 
+// Adds the Core API routes to a preexisting http handler.
```
jbowens: Should be 
```
// Handler adds the ...
```
https://golang.org/doc/effective_go.html#commentary

core/config/config.proto
```diff
@@ -0,0 +1,28 @@
...
+	bc.ProtoHash blockchain_id = 4;
+	string generator_url = 5;
+	string generator_access_token = 6;
+	string block_hsm_url = 7;
+	string block_hsm_access_token = 8;
+	uint64 configured_at = 9;
+	string block_pub = 10;
+	repeated BlockSigner signers = 11;
+	uint32 quorum = 12;
+	uint64 max_issuance_window = 13	;
```
jbowens: nit: extra space before the semicolon

database/raft/state/state.go
```diff
@@ -0,0 +1,225 @@
...
+	for _, op := range instr.Operations {
+		switch op.Type {
+		case statepb.Op_SET:
+			s.state[op.Key] = op.Value
+			s.version[op.Key] = index
+		case statepb.Op_DELETE:
+			delete(s.state, op.Key)
+			delete(s.version, op.Key)
+			//TODO (ameets):increment version or delete entry?
+			//s.version[op.Key] = index
```
jbowens: Does this TODO need addressing?

ameets: It should be deleted

database/raft/raft.go
```diff
@@ -0,0 +1,920 @@
...
+	}
+	if crc32.Checksum(snapPB.Data, crcTable) != snapPB.Crc {
+		return errors.Wrap(errors.New("bad snapshot crc"))
+	}
+	err = snapshot.Unmarshal(snapPB.Data)
+	return errors.Wrap(err)
+}
+
+// TODO(kr): under what circumstances should we do this?
+func (sv *Service) evict(nodeID uint64) {
```
bobg: This stub should be removed until/unless it's needed.

database/raft/raft.go
```diff
@@ -0,0 +1,920 @@
...
+		return nil, errors.Wrap(err)
+	}
+
+	_, st, ents, err := wal.ReadAll()
+	if err != nil {
+		return nil, errors.Wrap(err)
+	}
+
+	sv.raftStorage.ApplySnapshot(raftSnap)
+	if raft.IsEmptySnap(raftSnap) {
```
bobg: `if !raft.IsEmptySnap(...)` (and remove empty if branch).

database/raft/rand.go
```diff
@@ -0,0 +1,15 @@
+package raft
+
+import (
+	"crypto/rand"
+)
+
+func randID() []byte {
+	c := 16 // prevent collisions
```
bobg: Why not `var b [16]byte`?

cmd/corectl/main.go
```diff
@@ -130,36 +142,34 @@ func configGenerator(db *sql.DB, args []string) {
 				fatalln("error:", "bad ed25519 public key length")
 			}
 			url := args[i+1]
-			signers = append(signers, config.BlockSigner{
+			signers = append(signers, &config.BlockSigner{
 				Pubkey: pubkey,
-				URL:    url,
+				Url:    url,
```
tessr: I 100% agree, but not much we can do about it :(

cmd/corectl/main.go
```diff
@@ -8,33 +8,38 @@ import (
...
 )
 
 const version = "1.1.0"
 
 // config vars
 var (
-	dbURL = env.String("DATABASE_URL", "postgres:///core?sslmode=disable")
+	dbURL      = env.String("DATABASE_URL", "postgres:///core?sslmode=disable")
+	listenAddr = env.String("LISTEN", ":1999")
+	dir        = env.String("DIR", defaultDir())
```
tessr: What do you think of DATA_DIR?

cmd/cored/main.go
```diff
@@ -143,6 +147,55 @@ func runServer() {
...
+		Handler:      secureheader.DefaultConfig,
+		ReadTimeout:  httpReadTimeout,
+		WriteTimeout: httpWriteTimeout,
+		// Disable HTTP/2 for now until the Go implementation is more stable.
+		// https://github.com/golang/go/issues/16450
+		// https://github.com/golang/go/issues/17071
+		TLSNextProto: map[string]func(*http.Server, *tls.Conn, http.Handler){},
+	}
+
+	go func() {
```
tessr: 1. It moved up because we need to start listening on the raft routes (`/raft/`) before we can set up the rest of the server (e.g. check if there's a config). We can't even read from raft storage without making this new node available to raft, because raft requires a round of consensus for reads as well as writes. 

2. It sprouted a goroutine (🌱 ) because the `ListenAndServe` calls are blocking and we needed to proceed through the rest of the set up. 

I'll add comments explaining all of this.

cmd/cored/main.go
```diff
@@ -528,3 +544,8 @@ func (w *errlog) Write(p []byte) (int, error) {
 	}
 	return len(p), nil // report success for the MultiWriter
 }
+
+func defaultDir() string {
+	// TODO(kr): something in ~/Library on darwin?
+	return filepath.Join(os.Getenv("HOME"), ".cored")
```
tessr: Now that you mention it, we set up directories through the installers for both windows and mac. We should probably just stick this in the same directory?

database/raft/raft_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+var idCases = []struct {
+	id   uint64
+	data []byte
+}{
+	{1, []byte{0, 0, 0, 0, 0, 0, 0, 1, 0x7e, 0x43, 0x31, 0x89}},
+	{2, []byte{0, 0, 0, 0, 0, 0, 0, 2, 0x6d, 0x13, 0xc2, 0x7d}},
+}
+
+var idErrorCases = []struct {
```
kr: This could be [][]byte.

database/raft/raft.go
```diff
@@ -0,0 +1,920 @@
...
+		}
+		sendmsg(addr, data)
+	}
+}
+
+// best effort. if it fails, oh well -- that's why we're using raft.
+func sendmsg(addr string, data []byte) {
+	resp, err := http.Post("http://"+addr+"/raft/msg", contentType, bytes.NewReader(data))
+	if err != nil {
+		log.Error(context.Background(), err)
```
kr: Maybe add a todo here to change this to print "warning" instead of the word "error" to the log.

database/raft/raft.go
```diff
@@ -0,0 +1,920 @@
...
+	}
+	sv.id = x.ID
+	var raftSnap raftpb.Snapshot
+	err = decodeSnapshot(x.Snap, &raftSnap)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	ctx := context.Background()
+	sv.raftStorage.ApplySnapshot(raftSnap)
```
kr: We should check this error.

database/raft/raft.go
```diff
@@ -0,0 +1,920 @@
...
+func (sv *Service) join(addr, baseURL string) error {
+	reqURL := strings.TrimRight(baseURL, "/") + "/raft/join"
+	b, _ := json.Marshal(struct{ Addr string }{addr})
+	resp, err := http.Post(reqURL, contentType, bytes.NewReader(b))
+	if err != nil {
+		return errors.Wrap(err)
+	}
+	defer resp.Body.Close()
+
+	var x struct {
```
kr: Might want to factor this out into a named type since we also use this same type in serveJoin.

database/raft/raft.go
```diff
@@ -0,0 +1,920 @@
...
+	// The actual replicated data set.
+	stateMu   sync.Mutex
+	stateCond sync.Cond
+	state     *state.State
+	done      bool
+
+	// Current log position.
+	// access only from runUpdates goroutine
+	snapIndex uint64
+	confState raftpb.ConfState
```
kr: Probably need to move confState (and possibly also snapIndex) under stateMu.

cmd/corectl/main.go
kr: How about `CORED_DATA`?

cmd/corectl/main.go
kr: Not that it's any consolation, but "Go style" makes an exception for this case:

> Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

database/raft/raft.go
kr: In general, it would be good to run it once under -race and see if it warns about anything.

cmd/corectl/main.go
tessr: I'm going with `CORED_DATA_DIR` because we seem to like to stick nouns on the end of our env vars

database/raft/raft.go
tessr: I've run this as-is with `-race` and didn't get any warnings (same with when I ran the tests, too). I'll look at this, though

database/raft/rand.go
kr: Normal people almost never use arrays in Go. (We do quite a bit because we have 256-bit hash floating around all over the place so we got used to them.) So prob the most idiomatic way to write this is `b := make([]byte, 16)`.

tessr: I am going to idiomaticize this and also clarify that comment.

database/raft/raft.go
tessr: I just went ahead and did it /shruggie

tessr:
It's a miracle! We have a Chain Core running on raft on Windows!!

screen shot 2017-03-16 at 5 20 26 pm

This is ready for another pass.

tessr:
Note also that the raft change now requires that config changes through `corectl` happen while `cored` isn't running (since each process locks the raft wal).

kr:
:eyes:!

tessr:
Rebased and shrunk! PTAL

core/config/config.go
```diff
@@ -6,22 +6,24 @@ import (
...
-	"chain/database/sql"
-	chainjson "chain/encoding/json"
+	"chain/database/raft"
 	"chain/errors"
 	"chain/protocol"
 	"chain/protocol/bc"
 	"chain/protocol/state"
 )
 
+//go:generate protoc -I. -I$CHAIN/.. --go_out=. config.proto
```
bobg: Would it be better to put this line at the top of the file? And/or to add a comment saying that this produces the code for the `Config` and `BlockSigner` types?

core/config/config.proto
```diff
@@ -0,0 +1,28 @@
...
+	bc.ProtoHash blockchain_id = 4;
+	string generator_url = 5;
+	string generator_access_token = 6;
+	string block_hsm_url = 7;
+	string block_hsm_access_token = 8;
+	uint64 configured_at = 9;
+	string block_pub = 10;
+	repeated BlockSigner signers = 11;
+	uint32 quorum = 12;
+	uint64 max_issuance_window = 13;
```
bobg: Let's stick with the convention of adding a unit designator to identifiers like this (so, `max_issuance_window_ms`).

cmd/cored/main.go
```diff
@@ -196,9 +207,9 @@ func runServer() {
 	if err != nil {
 		chainlog.Fatalkv(ctx, chainlog.KeyError, err)
 	}
-	resetInDevIfRequested(db)
+	resetInDevIfRequested(db, raftDB)
 
-	conf, err := config.Load(ctx, db)
+	conf, err := config.Load(ctx, db, raftDB)
```
jbowens: Do we know how we're going to migrate existing configurations to the new raft/protobuf? Maybe if we don't have a config, we should check the `config` table in PostgreSQL? If it's there, try to save it via raft?

core/api.go
```diff
@@ -113,8 +114,8 @@ func (a *API) needConfig() func(f interface{}) http.Handler {
 	return jsonHandler
 }
 
-func (a *API) buildHandler() {
-	// Setup the muxer.
+// buildCoreHandler adds the Core API routes to a preexisting http handler.
+func (a *API) buildCoreHandler() {
```
kr: Maybe we can get away without renaming this method?

core/api.go
```diff
@@ -183,18 +184,17 @@ func (a *API) buildHandler() {
...
-	handler = maxBytes(handler)
+	handler = maxBytes(handler) // TODO(tessr): consider moving this to non-core specific mux
 	handler = webAssetsHandler(handler)
 	handler = healthHandler(handler)
 	for _, l := range a.requestLimits {
 		handler = limit.Handler(handler, alwaysError(errRateLimited), l.perSecond, l.burst, l.key)
 	}
 	handler = gzip.Handler{Handler: handler}
 	handler = coreCounter(handler)
-	handler = reqid.Handler(handler)
```
kr: Haha, so we currently have two reqid handlers in the chain? Sure enough.

core/config/config.proto
kr: This reminds me, it would be good to double check that the generated type is actually compatible. Protoc generates json tags, but I am not 100% sure that all the conventions match (e.g. decoding hex vs base64). The BlockSigner pubkey field in particular might be problematic.

log/log.go
```diff
@@ -47,7 +47,8 @@ const (
 
 	KeyMessage = "message" // produced by Message
 	KeyError   = "error"   // produced by Error
-	KeyStack   = "stack"   // used by Printkv to print stack on subsequent lines
+	KeyWarning = "warning"
```
kr: Is KeyWarning unused?

cmd/cored/main.go
tessr: Oof, yeah, this is a good point. I've been trying to punt on "polishing" work but this will break systems with the config in postgres, so my inclination is to try to squeak that into this PR

log/log.go
tessr: Yes, good catch

core/api.go
tessr: yeah, fair

tessr:
Hey guess what!!!!!!!

This is ready for another pass! (13 days later)

kr:
:eyes:

cmd/corectl/main.go
```diff
@@ -8,24 +8,30 @@ import (
...
 	"chain/generated/rev"
 	"chain/log"
+	"chain/protocol/bc"
 )
 
 // config vars
 var (
-	dbURL = env.String("DATABASE_URL", "postgres:///core?sslmode=disable")
+	dbURL      = env.String("DATABASE_URL", "postgres:///core?sslmode=disable")
+	listenAddr = env.String("LISTEN", ":1999")
```
kr: Maybe I'm not totally up to date on the thinking here. Somehow I thought the plan was to have corectl act as a client to do config and reset operations. I'm pretty sure we already expose the necessary RPCs.

Making corectl actually do raft raises questions like, it'll have to join the cluster to do its operation, but then does it remove itself at the end? Or will the user have to do that as a separate step?

core/config/config.go
```diff
@@ -2,6 +2,10 @@
 // Chain Core.
 package config
 
+// The following go:generate command is included in order to generate code for
```
kr: mild suggestion:

s/The following go:generate command is included in order to//

core/config/config.proto
```diff
@@ -0,0 +1,28 @@
...
+	bool is_generator = 3;
+	bc.ProtoHash blockchain_id = 4;
+	string generator_url = 5;
+	string generator_access_token = 6;
+	string block_hsm_url = 7;
+	string block_hsm_access_token = 8;
+	uint64 configured_at = 9;
+	string block_pub = 10;
+	repeated BlockSigner signers = 11;
+	uint32 quorum = 12;
```
kr: Should this be `int` to avoid the type conversion on config.go line 197?

core/config/config.go
```diff
@@ -38,31 +44,47 @@ var (
...
+		if config != nil {
+			val, err := proto.Marshal(config)
+			if err != nil {
+				return nil, errors.Wrap(err)
+			}
+			err = rDB.Insert(ctx, "/core/config", val)
+			if err != nil {
+				return nil, errors.Wrap(err)
+			}
+			return config, nil
```
kr: Should this migration logic also delete the old config to avoid confusion? Maybe it should even drop the table???

core/config/config.proto
tessr: Do protobufs have an `int` type? (I'm pretty sure they don't.)

cmd/corectl/main.go
tessr: My original plan had been to turn `corectl` into a client as a polish step, since I'm reluctant to pile more into this PR, and running this seems to work okay. BUT given that I was also thinking that the ability to remove a node from a cluster is also a polish step, I guess I should fix this (that is, turn corectl into a client) in this PR.

core/config/config.proto
kr: Oh I guess not, my bad. 😅

tessr:
This is ready for another pass 🏁

kr:
:eyes:

kr:
ok, actually :eyes: now

wercker.yml
```diff
@@ -1,4 +1,4 @@
-box: chaindev/ci:20170306
+box: chaindev/ci:20170405
```
kr: Is this a real diff?

kr:
LGTM, but this should prob wait for #850 (so no more ProtoHash).

wercker.yml
tessr: yes.

tessr: although, looks like I'll have to build a new docker image and update this again, since there's a merge conflict.

PR #684, opened 2017-03-02
cmd/cored: configure root CA pool

kr:
By default, package crypto/tls uses the operating
system's set of global root CAs. This is unsuitable for
a typical production deployment, where the opeator will
run its own CA and provision its own certs.

This change provides a new environment variable:

    ROOT_CA_CERTS=certs.pem

This variable holds a filename of a file containing a list
of PEM-encoded root X.509 certificates.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -4,11 +4,13 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/hex"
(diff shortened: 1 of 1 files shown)


kr:
PTAL

kr:
Note the base branch is 1.1-stable.

bobg:
Looks straightforward enough, though I have minimal TLS expertise.

Issue #668, opened 2017-03-02
Transaction input account annotations not always set properly


jeffomatic:
UPDATE: This affects the following input annotations:

- account_id
- account_alias
- account_tags
- is_local

~@chrisgarvin originally reported an issue where the API was erroneously returning `"is_local": "no"` for transaction inputs that are local to the core. #612 was an attempt to address this issue, but it appears to be ongoing. We're now trying to create a reliable repro case.~

jbowens:
The problem is that the tx annotating process uses the `account_control_programs` table for indexing spends. If the ACP expired, it's not guaranteed to still exist despite the account UTXO still being unspent. We can update the tx annotating to use the `account_utxos` table, but we'll need to ferry along the `change` flag.

jbowens:
Actually, maybe we don't even need the change flag.

jeffomatic:
Ah okay. I'm guessing this actually does more than just cause an incorrect `is_local` flag...it probably omits the entire account annotation on the input, right? If so I can update the original comment in this issue.

jbowens:
yeah, the entire account annotation is omitted

jbowens:
We are targeting a fix for 1.1.x, right? But not super high priority?

jeffomatic:
@jbowens I think we can deprioritize relative to our immediate bugfix push, but since this is an API bug it would be good to retire it as soon as we get the headroom.

jbowens:
Fixed on `1.1-stable` but will need a more invasive fix for `main`

PR #686, opened 2017-03-03
core/account: do not delete expired, in-use acps

jbowens:
Do not delete account control programs that are expired but still hold
assets.

This is a minimal, short-term fix for #668 within the 1.1.x release branch.
In the future, we should use `account_utxos` to annotate inputs &
outputs instead.


bobg:
LGTM but would love to see a test.

PR #687, opened 2017-03-03
sdk/java: self signed certs

ryandotsmith:
WIP

PR #688, opened 2017-03-03
sdk/java: allow customizable CA certs

jeffomatic:
This is an interface revision of #687. It should be backported to the `1.1-stable` branch.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -9,7 +9,10 @@ import java.io.InputStreamReader;
 import java.lang.reflect.Type;
 import java.net.*;
 import okio.Buffer;
-import java.security.*
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.security.cert.Certificate;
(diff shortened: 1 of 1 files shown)


jeffomatic:
PTAL. Usage example:

```
    String pem = new String(Files.readAllBytes(Paths.get("./certs.pem")));
    Client client = new Client.Builder()
      .setURL("https://localhost:8443")
      .setTrustedCerts(pem)
      .build();
```

kr:
Should we read the file for them? Take a filepath as the parameter?

jeffomatic:
I generally prefer being agnostic to the source of the certs, e.g. they could be sitting in an envvar too. Might be convenient to add a file-based option though? `setTrustedCertsFromFile`?

ryandotsmith:
@jeffomatic +1 for reading from a file. I think we are moving this direction for cored config.

jeffomatic:
ok

jeffomatic:
@kr @ryandotsmith revised to read from file, PTAL

cf: https://github.com/chain/chain/pull/688/commits/e7ef31f109bfc821017d3e2ba04865b998a26b45

PR #693, opened 2017-03-03
core/account: delete spents in separate processor

jbowens:
Delete spent account outputs in a separate block processor from the
indexing block processor. This is a prerequisite to removing the `query`
processor's dependency on the `account_control_programs` table.

For #668.

The new ordering of block processor dependencies:
(account, asset) <- query <- (expire ACPs, delete spent outputs)

### DIFF
--- a/core/query/index.go
+++ b/core/query/index.go
@@ -59,6 +59,7 @@ func (ind *Indexer) ProcessBlocks(ctx context.Context) {
 // saves all annotated transactions to the database.
 func (ind *Indexer) IndexTransactions(ctx context.Context, b *bc.Block) error {
 	<-ind.pinStore.PinWaiter("asset", b.Height)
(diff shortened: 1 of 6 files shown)


core/query/index.go
```diff
@@ -39,6 +39,7 @@ func (ind *Indexer) ProcessBlocks(ctx context.Context) {
 // saves all annotated transactions to the database.
 func (ind *Indexer) IndexTransactions(ctx context.Context, b *bc.Block) error {
 	<-ind.pinStore.PinWaiter("asset", b.Height)
+	<-ind.pinStore.PinWaiter("account", b.Height)
```
kr: Just double checking, this can't use account.PinName because it would be a dependency cycle?

kr:
LGTM

PR #694, opened 2017-03-03
core/query: consolidate index.go and indexer.go

jbowens:
`indexer.go` contained the type definition for `query.Indexer` but all
of the methods and logic were defined in `index.go`. It's hard to
remember which is which.


kr:
LGTM

PR #695, opened 2017-03-03
core/txbuilder: remove unneeded code, unexport some identifiers

bobg:
The `WitnessComponent` interface used to be instantiated by a few different types relating to smart contracts. Now there's only `SignatureWitness` so there's no need for an interface.

`KeyIDs` and `AddWitnessKeys` were only ever being used together, so they're now coalesced.

Some previously exported identifiers didn't need to be, so no longer are.


bobg:
PTAL

jbowens:
lgtm

PR #696, opened 2017-03-03
core/account: annotate spend inputs correctly

### LINKED ISSUE #668: Transaction input account annotations not always set properly

jbowens:
Correctly annotate spend inputs with account data. When annotating
inputs, determine the source account by looking up the spent output in
`account_utxos` instead of looking up the control program in
`account_control_programs`. The control program isn't guaranteed to
still exist once it's expired.

In a follow up, we should change the output annotating to use
`account_utxos` as well. We'll need to add the `change` flag to
the `account_utxos` table.

Fix #668.

### DIFF
--- a/core/account/annotate.go
+++ b/core/account/annotate.go
@@ -9,26 +9,30 @@ import (
 
 	"chain/core/query"
 	"chain/database/pg"
+	"chain/errors"
+	"chain/protocol/bc"
 )
+
+var empty = json.RawMessage(`{}`)
 
 // AnnotateTxs adds account data to transactions
 func (m *Manager) AnnotateTxs(ctx context.Context, txs []*query.AnnotatedTx) error {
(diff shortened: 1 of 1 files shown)


jbowens:
PTAL. Note this was fixed on the `1.1-stable` branch in #686 with a smaller, short-term fix.

jbowens:
PTAL

erykwalder:
:eyes:

erykwalder:
LGTM

PR #698, opened 2017-03-04
docker/ci: update OS to CentOS 6

boymanjor:
The majority of our clients will be running a version of RHEL
and our build/test environments should be closely aligned
with our intended, deploy targets. 

This commit also includes necessary updates to the java tests.

The image includes:
- Go 1.8
- Java 1.7
- Ruby 2.0.0
- Node v4.8
- various build/test dependencies

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -3,85 +3,25 @@ FROM centos:6
 ENV CHAIN="/go/src/chain" \
     BUNDLE_APP_CONFIG="/usr/local/bundle" \
     GEM_HOME="/usr/local/bundle" \
-    GOLANG_VERSION="1.8" \
-    GOLANG_URL="https://golang.org/dl/go1.8.linux-amd64.tar.gz" \
-    GOLANG_SHA256="53ab94104ee3923e228a2cb2116e5e462ad3ebaeea06ff04463479d7f12d27ca" \
     GOPATH="/go" \
     GOROOT="/usr/local/go" \
     JAVA_HOME="/opt/jdk1.7.0_79" \
     JRE_HOME="/opt/jdk1.7.0_79/jre" \
     MAVEN_HOME="/usr/share/maven" \
-    MAVEN_VERSION="3.3.9" \
-    PATH="$PATH:/usr/local/bundle/bin:/go/bin:/usr/local/go/bin:/opt/jdk1.7.0_79/bin:/opt/jdk1.7.0_79/jre/bin" \
-    RUBY_MAJOR="2.0" \
-    RUBY_VERSION="2.0.0-p594"
-
-RUN yum -y update \
-    && yum install -y autoconf automake bash curl gcc gcc-c++ git libtool make ruby wget unzip && yum clean all \
-# Install Go 1.8
-    && curl -fsSL -o golang.tar.gz "$GOLANG_URL" \
-    && echo "$GOLANG_SHA256  golang.tar.gz" | sha256sum -c - \
-    && tar -C /usr/local -xzf golang.tar.gz \
-    && rm golang.tar.gz \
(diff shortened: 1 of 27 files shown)


boymanjor:
PTAL

docker/ci/Dockerfile
```diff
@@ -1,27 +1,87 @@
...
+    GOLANG_SHA256="53ab94104ee3923e228a2cb2116e5e462ad3ebaeea06ff04463479d7f12d27ca" \
+    GOPATH="/go" \
+    GOROOT="/usr/local/go" \
+    JAVA_HOME="/opt/jdk1.7.0_79" \
+    JRE_HOME="/opt/jdk1.7.0_79/jre" \
+    MAVEN_HOME="/usr/share/maven" \
+    MAVEN_VERSION="3.3.9" \
+    PATH="$PATH:/usr/local/bundle/bin:/go/bin:/usr/local/go/bin:/opt/jdk1.7.0_79/bin:/opt/jdk1.7.0_79/jre/bin" \
+    RUBY_MAJOR="2.0" \
+    RUBY_VERSION="2.0.0-p594"
```
kr: Minor but I'd slightly prefer leaving a trailing backslash on the last line here. That way it's less likely that additional lines will forget it, and it makes the diff less noisy.

docker/ci/Dockerfile
```diff
@@ -1,27 +1,87 @@
...
+
+RUN env && yum -y update \
```
kr: What is the purpose of the `env` command here?

docker/ci/Dockerfile
```diff
@@ -1,27 +1,87 @@
...
+    && gem install bundler \
+    && bundle config --global path "$GEM_HOME" \
+    && bundle config --global bin "$GEM_HOME/bin" \
+# Install gas
     && git clone https://github.com/GoASTScanner/gas $(go env GOPATH)/src/github.com/GoASTScanner/gas \
     && cd $(go env GOPATH)/src/github.com/GoASTScanner/gas \
     && git reset --hard d30c5cde3613e9ba0129febda849e4d4df1d57cd \
-    && go install github.com/GoASTScanner/gas
+    && go install github.com/GoASTScanner/gas \
+    && rm -rf $(go env GOPATH)/src/github.com/GoASTScanner/gas && cd
```
kr: What is the `cd` command doing at the end?

Also, same minor suggestion about trailing backslash.

docker/ci/Dockerfile
boymanjor: Whoops. This was left over from testing.

docker/ci/Dockerfile
boymanjor: Because I'm deleting the working directory, I call `cd` to return to `$HOME` as a courtesy to the person adding the next line to this file.

docker/ci/Dockerfile
boymanjor: I totally agree with you, but the Dockerfile doesn't recognize the following `RUN` as a proper docker command. It parses it as a continuation of the `ENV` command and returns an error.

boymanjor:
@kr what do you think of the following:
- separate each install step into its own script file
- include those files in the `bin` directory
- move the `ADD bin /usr/bin` command to before the `RUN`
- call each script within the `RUN` command and delete it after a successful run

This would also get rid of a bunch of the versioning related env vars specified at the beginning of the file.

docker/ci/Dockerfile
kr: Hah, okay then.

kr:
That sounds great. I'd also be fine with one big script file. Whatever feels best.

boymanjor:
@kr PTAL

kr:
:eyes:

kr:
LGTM

PR #700, opened 2017-03-04
cmd/corectl: add 'migrate' subcommand

jbowens:
Remove the cmd/migratedb command and add the same functionality into a
corectl subcommand.

Also, no longer automatically run all migrations before any corectl
command. Instead, if a subcommand requires an initialized database,
run migrations only if the schema has never been migrated. This allows
new Cores to still be configured with a single command. It also prevents
accidental migration.

`cored` will continue to automatically apply all pending migrations on
startup.


jbowens:
PTAL

I think automatically applying all migrations on every corectl command can be pretty unexpected. I can imagine a scenario where you're upgrading to a new version of Chain Core, but before you deploy the new versions of cored you need to tweak configuration with corectl. If you're using the new version of corectl, you'd accidentally run the migrations before the new cored is deployed. :hurtrealbad:

I was going to remove the automatic migration from corectl altogether, but it's pretty convenient to just run `corectl config-generator` on a blank Postgres DB and have it just work. I figured a happy medium was to automatically apply the migrations only if it's a fresh database.

erykwalder:
LGTM

PR #703, opened 2017-03-06
core: tick network RPC version to 3

jeffomatic:
This accounts for updates affecting validation-critical hashing
in #665 (#677 in 1.1-stable).

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -23,7 +23,7 @@ var (
 )
 
 const (
-	networkRPCVersion = 2
+	networkRPCVersion = 3
 )
 
 func (a *API) reset(ctx context.Context, req struct {


jeffomatic:
PTAL

jbowens:
Should this be for the 1.1-stable branch? If `main` is the 1.2 dev branch, will it need to be ticked to 4 given our roadmap for 1.2?

jeffomatic:
@jbowens Good question. I was thinking about recommending that we increment the network version for *every* change that forces a blockchain reset and/or affects the semantics or format of the network RPC. But it doesn't remove the task of having to synchronize network versions across main and the release branches.

The process I'm going for right now is to make any changes reflecting the state-of-the-art of *any* file in `main` (hence this PR). That way, if any future change on `main` affects the network version, it will be clear that it should be updated to `4` and not `3`. If this PR lands, I will follow up with a PR that pushes this change to `1.1-stable`.

jbowens:
Hrm, how are we planning on handling updates to the network protocol, given that you can't atomically update all cored instances and clients without taking downtime? If 1.2 adds a gRPC API, I assumed we would need to run the existing JSON API and the new gRPC API in tandem in 1.2. Then in 1.3 we could remove the JSON API.

Are two network RPC versions only compatible if they're exactly equal?

jeffomatic:
The existing version scheme requires strict equality of the network RPC version for compatibility.

For upgrades, I think we'd need an expanded conversation on whether the network version is the right concept for framing compatibility. Right now it's got a fairly broad definition and captures some updates that might be trivial (e.g., wire format changes that are simple to make backward-compatible) and hard breaking changes (e.g., changes to the transaction structure or hashing schemes).

jeffomatic:
@jbowens I think upgrading is a good topic for exploration, but do you think it's a blocking question for a 1.1.2 release?

kr:
I tend to agree with all of this. We should have a broader discussion about network-level protocol compatibility for 1.2, and that should not block 1.1.2. So this PR should probably land on 1.1-stable and then we can figure out what to do over the next couple of months for 1.2.

PR #711, opened 2017-03-07
all: include chain-core-server-1.1.2 in all builds

jeffomatic:
This updates all installers to compile cored version 1.1.2. Additionally,
it updates contents of the root changelog on `main` to point to each
release's on-branch version, to avoid redundant commits to both main and
the release.

Complete list of updates:

- cored `latestVersion` constant
- corectl `version` constant
- Docker image build script
- Mac installer and changelog
- Windows installer and changelog


jeffomatic:
PTAL, should be the final backport commit for 1.1.2

PR #712, opened 2017-03-07
log: adopt -kv naming convention

kr:
Standard library package log uses the suffix "f" to
indicate that a function takes a format string. We adopt
the same convention here, plus we use the suffix "kv" to
incicate the arguments are key-value pairs.

### DIFF
--- a/core/account/accounts.go
+++ b/core/account/accounts.go
@@ -70,7 +70,7 @@ func (m *Manager) ExpireReservations(ctx context.Context, period time.Duration) 
 	for {
 		select {
 		case <-ctx.Done():
-			log.Messagef(ctx, "Deposed, ExpireReservations exiting")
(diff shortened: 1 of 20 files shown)


core/account/accounts.go
```diff
@@ -70,7 +70,7 @@ func (m *Manager) ExpireReservations(ctx context.Context, period time.Duration)
 	for {
 		select {
 		case <-ctx.Done():
-			log.Messagef(ctx, "Deposed, ExpireReservations exiting")
+			log.Printf(ctx, "Deposed, ExpireReservations exiting")
```
bobg: Should we also have a `Print` for this case?

bobg:
LGTM with one small comment.

core/account/accounts.go
kr: We sure could but I'll save behavior changes for another PR.

PR #713, opened 2017-03-07
protocol/bc: move protocol/tx code here

bobg:
...In anticipation of rewriting validation in terms of txentries objects.

Among other benefits, this allows us to get rid of a lot of `init_test.go` files, yay!

(This is a repeat of https://github.com/chain/chain/pull/705.)


bobg:
PTAL

bobg:
rebased

bobg:
PTAL

bobg:
PTAL

kr:
EYES

protocol/bc/map.go
```diff
@@ -76,10 +74,36 @@ func mapTx(tx *bc.TxData) (headerID bc.Hash, hdr *header, entryMap map[bc.Hash]e
 				}
 
 				assetID := oldIss.AssetID()
-				b := vmutil.NewBuilder()
-				b = b.AddData(oldIss.Nonce).AddOp(vm.OP_DROP).AddOp(vm.OP_ASSET).AddData(assetID[:]).AddOp(vm.OP_EQUAL)
 
-				nonce = newNonce(program{1, b.Program}, tr)
+				// This is the program
+				//   [PUSHDATA(oldIss.Nonce) DROP ASSET PUSHDATA(assetID) EQUAL]
+				// minus a circular dependency on protocol/vm.
```
kr: This reminds me I would like to get rid of the vm -> bc dependency. Package vm should instead define one or two small interfaces that bc satisfies. But this seems fine to me for now.

protocol/bc/txtransaction.go
```diff
@@ -1,25 +1,15 @@
...
 
 import (
 	"fmt"
 
 	"chain/crypto/sha3pool"
 	"chain/errors"
-	"chain/protocol/bc"
 )
 
-func init() {
```
kr: \o/

bobg:
rebased

PR #715, opened 2017-03-07
docs: use release branches as source of production docs

jeffomatic:
Now that we have the 1.1-stable branch, we can simplify our
documentation workflow by simply generating and uploading
documentation from the tip of that branch. For the time being, it
doesn't seem like a separate docs tag is necessary.

This commit updates various pieces of documentation to account for
this simplified workflow.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -41,11 +41,13 @@ Log into `aws` with the command:
 $ aws configure
 ```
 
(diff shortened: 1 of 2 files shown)


jeffomatic:
PTAL

jeffomatic:
PTAL

PR #716, opened 2017-03-07
core/signers: correctly record xpubs for a signer with multiple keys

bobg:
Avoid the Go pitfall involving reuse of loop variables.


bobg:
PTAL

kr:
LGTM

jeffomatic:
Would it be possible to extend our Java integration tests to catch this case?

dominic:
re @jeffomatic's comment, I'd call https://github.com/chain/chain/blob/main/sdk/java/src/test/java/com/chain/integration/TransactionTest.java#L177 an incomplete test of this use case. We should have a test that fails with less than the minimum number of signatures too.

PR #717, opened 2017-03-07
core/asset: avoid Go loop-variable pitfall

bobg:
This is the result of a code audit following the discovery of the bug fixed in https://github.com/chain/chain/pull/716. I found only one other place that needed the same change.


bobg:
PTAL

kr:
LGTM

PR #718, opened 2017-03-07
core/asset, core/signers: correctly record xpubs for a signer with multiple keys; avoid Go loop pitfall

bobg:
This is https://github.com/chain/chain/pull/716 + https://github.com/chain/chain/pull/717.


bobg:
PTAL

jbowens:
lgtm

PR #719, opened 2017-03-07
CHANGELOG.md: update release notes for Chain Core 1.1.3

kr:

PR #720, opened 2017-03-07
core: tick version to 1.1.3

kr:


kr:
PTAL @jeffomatic @ryandotsmith 

I guess I should have done this in the same PR as #719 but oh well.

PR #722, opened 2017-03-08
protocol/tx: catch more nils in entryID

bobg:
Bart at chalkboard: A nil interface is not the same as a nil pointer.

### DIFF
--- a/protocol/tx/entry.go
+++ b/protocol/tx/entry.go
@@ -26,14 +26,13 @@ type entry interface {
 var errInvalidValue = errors.New("invalid value")
 
 func entryID(e entry) (hash bc.Hash) {
-	// This nil test only handles the case where e is the zero value of
-	// the entry interface.
 	if e == nil {
 		return hash
 	}
 
-	// This nil test handles the case where e is a nil pointer.
-	if reflect.ValueOf(e).IsNil() {
+	// Nil pointer; not the same as nil interface above. (See
+	// https://golang.org/doc/faq#nil_error.)
+	if v := reflect.ValueOf(e); v.Kind() == reflect.Ptr && v.IsNil() {
 		return hash
 	}
 
--- a/protocol/tx/entry.go
+++ b/protocol/tx/entry.go
@@ -26,7 +26,15 @@ type entry interface {
 var errInvalidValue = errors.New("invalid value")
 
 func entryID(e entry) (hash bc.Hash) {
+	// This nil test only handles the case where e is the zero value of
+	// the entry interface.
 	if e == nil {
+		return hash
+	}
+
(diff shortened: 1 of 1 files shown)


bobg:
PTAL

protocol/tx/entry.go
```diff
@@ -26,10 +26,17 @@ type entry interface {
...
 
 func entryID(e entry) (hash bc.Hash) {
+	// This nil test only handles the case where e is the zero value of
+	// the entry interface.
 	if e == nil {
 		return hash
 	}
 
+	// This nil test handles the case where e is a nil pointer.
+	if reflect.ValueOf(e).IsNil() {
```
kr: May I suggest combining these into a single test and still including the commentary. Plus maybe a link to the FAQ might obviate some of the verbiage in this file.

```go
// Check for both entryID(nil) and entryID(x)
// where x is a concrete nil pointer.
// See https://golang.org/doc/faq#nil_error.
if e == nil || reflect.ValueOf(e).IsNil() {
	return bc.Hash{}
}
```

bobg: Is that idiomatic? I wonder whether it wouldn't be better to defend against the panic in `IsNil`:

```
if e == nil {
  return bc.Hash{}
}

v := reflect.ValueOf(e)
switch v.Kind() {
  case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Interface, reflect.Slice:
    if v.IsNil() {
      return bc.Hash{}
    }
}

...
```

(That could be factored out into a `SafeIsNil` or something.)

kr: Ah I just reread the docs for IsNil; good catch.

I would say in general it's ok (idiomatic and reasonable) to do things that panic as long as the contract is clear to the caller. For example if we have a function that divides, the caller knows they shouldn't provide 0 as the divisor; in that case we don't even need to document that it would panic.

In this case, we would definitely need to document that entryID panics when called with a non-pointer-typed value, or else do the check. And it's not *that* hard to do the check, so why not. We only care about pointer types here, so it would suffice to do:

```go
// Check for both entryID(nil) and entryID(x)
// where x is a concrete nil pointer.
// See https://golang.org/doc/faq#nil_error.
if v := reflect.ValueOf(e); e == nil || v.Kind() == reflect.Ptr && v.IsNil() {
	return bc.Hash{}
}
```

bobg: I preferred to leave it as two separate `if`s to avoid constructing a `reflect.Value` in the `e == nil` case, but otherwise took your advice.

bobg:
PTAL

kr:
LGTM

PR #724, opened 2017-03-08
sdk/java: retry IOExceptions

jbowens:
Do not rely on OkHttp to retry IOExceptions and perform our own
retry logic in these cases.

1.1-stable version of #723


boymanjor:
lgtm

PR #726, opened 2017-03-08
sdk/java: ensure the client retry timespan exceeds leader election

boymanjor:
Leader election could potentially take 15secs. We need
to ensure the client retry logic takes this into account.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -408,15 +408,17 @@ public class Client {
   private static final Random randomGenerator = new Random();
   private static final int MAX_RETRIES = 10;
   private static final int RETRY_BASE_DELAY_MILLIS = 40;
-  private static final int RETRY_MAX_DELAY_MILLIS = 4000;
+
+  // the max amount of time cored leader election could take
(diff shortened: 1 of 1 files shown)


sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -408,15 +408,17 @@ private OkHttpClient buildHttpClient(Builder builder) {
   private static final Random randomGenerator = new Random();
   private static final int MAX_RETRIES = 10;
   private static final int RETRY_BASE_DELAY_MILLIS = 40;
-  private static final int RETRY_MAX_DELAY_MILLIS = 4000;
+
+  // the max amount of time cored leader election could take
+  private static final int RETRY_MAX_DELAY_MILLIS = 15000;
```
jbowens: could this be too much? I think this could cause it to wait 15 seconds between individual retries.

jbowens: If we want a snappy recovery, I think we want to retry often with short delays between retries.

jbowens: jk, talked about irl

jbowens:
lgtm

PR #727, opened 2017-03-08
sdk/java: increase index to core url monotically

boymanjor:
The client keeps a list of core URLs which it uses
to load balance requests to the server. An index
is kept corresponding to a position in this list. When
an error occurs the index is increased. 

Previously, it was possible for separate requests to
read the same index and increase it separately. We have
updated the logic to mitigate this by finding the position 
in the list (performing the modulo operation) just before
a request is made.


kr:
I'm not 100% sure about the description. This one is a bit subtle so it could prob use a little more explanation.

kr:
LGTM

PR #728, opened 2017-03-08
docs/protocol/specifications: update spec with new data structures and validation rules

danrobinson:
This updates the protocol specification to match the new data structures and hashing format now used in protocol/bc. The new specification also describes the validation rules that should be applied to those data structures.


danrobinson:
PTAL @bobg @kr

bobg:
:eyes:

docs/protocol/specifications/vm1.md
```diff
@@ -41,7 +41,7 @@ Virtual machines for control and issuance programs inside transactions are versi
 
 Nodes ignore programs with unknown versions, treating them like “anyone can issue/spend.” To discourage use of unassigned versions, block signers refuse to include transactions that use unassigned VM versions.
 
-Blocks do not specify VM version explicitly. [Consensus programs](data.md#consensus-program) use VM version 1 with additional [block-context restrictions](#block-context) applied to some instructions. Upgrades to block authentication can be made via additional fields in the block commitment string.
+Blocks do not specify VM version explicitly. [Consensus programs](blockchain.md#output-1) use VM version 1 with additional [block-context restrictions](#block-context) applied to some instructions. Upgrades to block authentication can be made via additional fields in the block commitment string.
```
bobg: Why don't blocks specify the VM version explicitly?

I don't know whether the answer belongs here, as it's not "specification," strictly speaking. But we've talked in the past about having a companion "motivations" document explaining various design decisions in the spec. I'd love to see that prioritized.

docs/protocol/specifications/vm1.md
```diff
@@ -57,9 +57,9 @@ All other instructions are encoded simply by a single-byte opcode. The protocol
 
 A program executes in a context, either a *block* or a *transaction*. Some instructions have different meaning based on the context.
 
-Transactions use [control programs](data.md#control-program) to define predicates governing spending of an asset in the next transaction, *issuance programs* for predicates authenticating issuance of an asset, and *program arguments* to provide input data for the predicates in output and issuance programs.
+Transactions use [control programs](blockchain.md#output-1) to define predicates governing spending of an asset in the next transaction, *issuance programs* for predicates authenticating issuance of an asset, and *program arguments* to provide input data for the predicates in output and issuance programs.
```
bobg: s/the next/a later/

docs/protocol/specifications/vm1.md
```diff
@@ -75,8 +75,8 @@ Execution of any of the following instructions results in immediate failure:
 * [AMOUNT](#amount)
 * [MINTIME](#mintime)
 * [MAXTIME](#maxtime)
-* [TXREFDATAHASH](#txrefdatahash)
-* [REFDATAHASH](#refdatahash)
+* [TXDATAHASH](#txdatahash)
```
bobg: As has been discussed elsewhere, whether or not this is a hash is an app-level decision. We should leave hash out of the name and just call this `TXDATA`. (`DATAHASH` can be `INPUTDATA` if just plain `DATA` is too generic.)

docs/protocol/specifications/vm1.md
```diff
@@ -116,6 +117,8 @@ Execution of any of the following instructions results in immediate failure:
 
 **Run Limit** is a built-in 64-bit integer specifying remaining total cost of execution. Run limit is decreased by the cost of each instruction and also affected by data added to and removed from the data stack and alt stack. Every byte added to either stack costs 1 unit, and every byte removed from either stack refunds 1 unit. (This includes explicit additions and removals by stack-manipulating instructions such as PUSHDATA and DROP, and also implicit additions and removals as when other instructions consume arguments and produce results.)
 
+**Expansion Flag** indicates whether the [expansion opcodes](#expansion-opcodes) are allowed in the program or not. If the flag is off, these opcodes immediately fail the program execution.
```
bobg: > these opcodes immediately fail the program execution

When executed? When encountered (even in a non-executing branch)? At some other time?

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1104,59 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
-    2. VM version equals `version`,
-    3. asset ID equals `assetid`,
-    4. amount equals `amount`,
-    5. `refdatahash` is an empty string or it matches the [SHA3-256](data.md#sha3) hash of the reference data.
+5. If the current entry is a [Mux](blockchain.md#mux-1):
+    1. Finds a [destination entry](blockchain.md#value-destination-1) at the given `index`.
+    2. If the entry satisfies all of the following conditions pushes [true](#vm-boolean) on the data stack; otherwise pushes [false](#vm-boolean):
+        1. the destination entry is an [output](blockchain.md#output-1) or a [retirement](blockchain.md#retirement-1),
+        2. if the destination is an output: control program equals `prog` and VM version equals `version`,
+        3. if the destination is a retirement: `prog` is an empty string and `version` is zero,
```
bobg: Do you want to special-case (`OP_FAIL`, vmversion 1) as a synonym here, for backward compatibility?

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1104,59 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
+        4. asset ID equals `assetid`,
+        5. amount equals `amount`,
+        6. `datahash` is an empty string or it matches the [SHA3-256](blockchain.md#sha3) hash of the data.
```
bobg: "...or it matches the data in the destination entry."

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1104,59 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
+5. If the entry is an [issuance](blockchain.md#issuance-1) or a [spend](blockchain.md#spend-1):
+    1. If the [destination entry](blockchain.md#value-destination-1) is a [Mux](blockchain.md#mux-1), performs checks as described in the step 5.
```
bobg: s/the step/step/

docs/protocol/specifications/vm1.md
```diff
@@ -57,9 +57,9 @@ All other instructions are encoded simply by a single-byte opcode. The protocol
...
 
-Blocks use [consensus programs](data.md#consensus-program) to define predicates for signing the next block and *program arguments* to provide input data for the predicate in the previous block. Consensus programs have restricted functionality and do not use version tags. Some instructions (such as [ASSET](#asset) or [CHECKOUTPUT](#checkoutput)) that do not make sense within a context of signing a block are disabled and cause an immediate validation failure.
+Blocks use [consensus programs](blockchain.md#block-header) to define predicates for signing the next block and *program arguments* to provide input data for the predicate in the previous block. Consensus programs have restricted functionality and do not use version tags. Some instructions (such as [ASSET](#asset) or [CHECKOUTPUT](#checkoutput)) that do not make sense within a context of signing a block are disabled and cause an immediate validation failure.
```
bobg: s/within a context/within the context/

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1104,59 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
+        2. if the destination is an output: control program equals `prog` and VM version equals `version`,
+        3. if the destination is a retirement: `prog` is an empty string and `version` is zero,
+        4. asset ID equals `assetid`,
+        5. amount equals `amount`,
+        6. `datahash` is an empty string or it matches the [SHA3-256](blockchain.md#sha3) hash of the data.
+5. If the entry is an [issuance](blockchain.md#issuance-1) or a [spend](blockchain.md#spend-1):
+    1. If the [destination entry](blockchain.md#value-destination-1) is a [Mux](blockchain.md#mux-1), performs checks as described in the step 5.
+    2. If the [destination entry](blockchain.md#value-destination-1) is an [output](blockchain.md#output-1) or a [retirement](blockchain.md#retirement-1):
+        1. If `index` is not zero, pushes [false](#vm-boolean) on the data stack.
+        2. Otherwise, performs checks as described in the step 5.2.
```
bobg: s/the step/step/

docs/protocol/specifications/vm1.md
```diff
@@ -1169,40 +1190,55 @@ Code  | Stack Diagram   | Cost
...
 
-#### TXREFDATAHASH
+#### TXDATAHASH
 
 Code  | Stack Diagram   | Cost
 ------|-----------------|-----------------------------------------------------
 0xc7  | (∅ → hash)      | 1; [standard memory cost](#standard-memory-cost)
 
-Pushes the SHA3-256 hash of the [transaction](data.md#transaction)'s reference data.
+Pushes the SHA3-256 hash of the transaction's reference data as specified in the Data field of the [transaction header](blockchain.md#transaction-header).
```
bobg: s/SHA3-256 hash of the// (also in `DATAHASH` below)

docs/protocol/specifications/vm1.md
```diff
@@ -1260,9 +1296,13 @@ Fails if executed in the [transaction context](#transaction-context).
 
 Code  | Stack Diagram   | Cost
 ------|-----------------|-----------------------------------------------------
-0x50, 0x61, 0x62, 0x65, 0x66, 0x67, 0x68, 0x8a, 0x8d, 0x8e, 0xa6, 0xa7, 0xa9, 0xab, 0xb0..0xbf, 0xca, 0xcd..0xcf, 0xd0..0xff  | (∅ → ∅)     | 1
+0x50, 0x61, 0x62, 0x65, 0x66, 0x67, 0x68, 0x8a, 0x8d, 0x8e, 0xa6, 0xa7, 0xa9, 0xab, 0xb0..0xbf, 0xcd..0xcf, 0xd0..0xff  | (∅ → ∅)     | 1
+
+The unassigned codes are reserved for future expansion.
+
+If the [expansion flag](#vm-state) is set, these opcodes have no effect on the state of the VM apart from reducing run limit by 1.
```
bobg: > these opcodes have no effect on the state of the VM apart from reducing run limit by 1

"...when executed."

Also, it's technically wrong to say they have no other effect on the state of the VM: they also increase the PC by one. Better to say, "these opcodes act exactly like `NOP`." (Except that I see there no longer is a `NOP` in the spec. Since when? The implementation still includes `NOP` at 0x61.)

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+
+### Hash
+
+A fixed-length 32-byte string.
+
+### List
+
+A `List` is encoded as a [Varstring31](#string) containing the serialized items, one by one, as defined by the schema. 
+
+Note: since the `List` is encoded as a variable-length string, its length prefix indicates not the number of _items_,
```
bobg: Alas, what's implemented (and what we're therefore stuck with for now) is [the opposite](https://github.com/chain/chain/blob/1c5b3b087b9cd6a068e65cf83e3bef171f753bd3/protocol/bc/entry.go#L101): the length prefix is the number of items.

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+1. BlockHeader entry,
+2. BlockHeader entry from the previous block, `PrevBlockHeader`.
+3. List of transactions included in block.
+
+**Algorithm:**
+
+1. Verify that the block’s version is greater or equal the block version in the previous block header.
+2. Verify that `Height` is equal to `PrevBlockHeader.Height + 1`.
+4. Verify that `PreviousBlockID` is equal to the entry ID of `PrevBlockHeader`.
+5. Verify that `Timestamp` is greater than `PrevBlockHeader.Timestamp`.
```
bobg: Strictly greater than, or greater-than-or-equal-to?

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+2. timestamp,
+3. block version.
+
+**Algorithm:**
+
+1. If the block version is 1, verify that version is equal to 1.
+2. If the `Maxtime` is greater than zero, verify that it is greater than or equal to the `Mintime`.
+3. If the `Mintime` is greater than zero:
+    1. Verify that the input timestamp is greater than or equal to the `Mintime`.
+4. If the transaction maxtime is greater than zero:
```
bobg: Why "the `Maxtime`" in (2) and "the transaction maxtime" here? Can this be combined with (2) so that Maxtime>=Mintime and blocktime<=Maxtime are sub-bullets?

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+but the number of _bytes_ of all the items in their serialized form.
+
+### Struct
+
+A `Struct` is encoded as a concatenation of all its serialized fields.
+
+### Public Key
+
+In this document, a *public key* is the 32-byte binary encoding
+of an Ed25519 (EdDSA) public key, as defined in [CFRG1](https://tools.ietf.org/html/draft-irtf-cfrg-eddsa-05).
```
tarcieri: All references to this draft should be switched to RFC 8032.

docs/protocol/specifications/blockchain.md
kr: Nice catch, but why "alas"? Unless there's some problem I'm not seeing with what we've implemented, it seems better. Finding the length in bytes would take extra work.

It might help if we avoid thinking of this whole thing as a serialization format; that's not what it is. (In hindsight, it would have made more sense to compute the txid as the hash of the entry hashes rather than the entries themselves. That'd be even less serial than what we have now.)

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+--------------------|----------------------|----------------
+Destination         | ValueDestination1    | The Destination ("forward pointer") for the value contained in this spend. This can point directly to an Output entry, or to a Mux, which points to Output entries via its own Destinations.
+Arguments           | List<String>         | Arguments for the control program contained in the SpentOutput.
+AnchoredEntry       | Pointer                      | Optional pointer to a single entry of any type, which uniquely identifies that entry as one that can use this one as an `Anchor`.
+
+#### Spend Validation
+
+1. Verify that `SpentOutput` is present in the transaction, but do not validate it.
+2. [Validate program](#program-validation) `SpentOutput.ControlProgram` with the given `Arguments` and the transaction version.
+3. Verify that `SpentOutput.Value` is equal to `Destination.Value`.
```
bobg: `SpentOutput.Source.Value`

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+4. [Validate](#value-destination-1-validation) `Destination`.
```
bobg: Uh oh - recursion in validation can't happen in both the forward and the backward directions. It _should_ happen in the backward direction, starting from a tx header's `Results`. But if one of those results is an `Output`, and it recursively validates its `Spend` source, and that `Spend` recursively validates its `Output` destination... well, then validation may take a while.

danrobinson: Good catch—it should just check if it is present in the transaction. I thought it said that (maybe that's in a different one.) Will fix

bobg: Hm, I take it back - validation of `ValueDestination`s does not recur into the referenced entry, so this is safe.

However, it may be worth adding some language emphasizing this so casual readers don't think there's an infinite regress here.

danrobinson: Oh right, that's why. I made this same mistake last week and Oleg set me straight

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+AssetDefinition     | [Asset Definition](#asset-definition)     | Asset definition for the asset being issued.
+Arguments           | List<String>                              | Arguments for the control program contained in the SpentOutput.
+AnchoredEntry       | Pointer                                   | Optional pointer to a single entry of any type, which uniquely identifies that entry as one that can use this one as an `Anchor`.
+
+#### Issuance Validation
+
+**Inputs:**
+
+1. Issuance entry,
+2. initial block ID.
```
bobg: Hm: if the initial block ID is required as input here, I wonder whether it should be specified as input to (at least) tx header validation.

Generally speaking, anything required to validate an entry X is part of the validation state required for _all_ entries that might recursively reach X. Might be worth defining a "validation state" data structure.

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+
+Transaction ID is defined as an [Entry ID](#entry-id) of the [transaction header](#transaction-header) structure.
+
+
+#### Transaction Header Validation
+
+**Inputs:**
+
+1. TxHeader entry,
+2. timestamp,
```
bobg: The timestamp and block version come from a block header. Previously we made a distinction between context-free validation of a transaction ("well-formedness") and validation in the context of a block. Is that distinction gone? (It's OK with me if it is.)

bobg: ...but it would mean that our tx-submission flow, which validates the proposed tx, would have to guess at the next block's timestamp and version, and will occasionally get it wrong, which means it will accept some txs that will eventually be rejected. Which I guess isn't too different from the situation we now have.

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+Witness field       | Type                         | Description
+--------------------|------------------------------|----------------
+Arguments           | List<String>                 | Arguments for the program contained in the Nonce.
+AnchoredEntry       | Pointer                      | Optional pointer to a single entry of any type, which uniquely identifies that entry as one that can use this one as an `Anchor`.
+
+
+#### Nonce Validation
+
+1. [Validate](#program-validation) `Program` with the given `Arguments`.
+2. Verify that `Issuance` points to an issuance that is present in the transaction (meaning visitable by traversing `Results` and `Sources` from the transaction header) and whose `Anchor` is equal to this nonce's ID.
```
bobg: s/Issuance/AnchoredEntry/ (right?)

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+--------------------|----------------------|----------------
+SpentOutput         | Pointer<Output1>      | The Output entry consumed by this spend.
+Data                | Hash                 | Hash of the reference data for this entry, or a string of 32 zero-bytes (representing no reference data).
+ExtHash             | [ExtStruct](#extension-struct) | If the transaction version is known, this must be 32 zero-bytes.
+
+Witness field       | Type                 | Description
+--------------------|----------------------|----------------
+Destination         | ValueDestination1    | The Destination ("forward pointer") for the value contained in this spend. This can point directly to an Output entry, or to a Mux, which points to Output entries via its own Destinations.
+Arguments           | List<String>         | Arguments for the control program contained in the SpentOutput.
+AnchoredEntry       | Pointer                      | Optional pointer to a single entry of any type, which uniquely identifies that entry as one that can use this one as an `Anchor`.
```
bobg: There probably needs to be a validation rule for `AnchoredEntry`

docs/protocol/specifications/blockchain.md
danrobinson: Yes, good catch—this rule should actually be eliminated. The `AnchoredEntry` should only be validated by the entry pointing to this one.

docs/protocol/specifications/blockchain.md
```diff
@@ -0,0 +1,613 @@
...
+--------------------|------------------------------|----------------
+Arguments           | List<String>                 | Arguments for the program contained in the Nonce.
+AnchoredEntry       | Pointer                      | Optional pointer to a single entry of any type, which uniquely identifies that entry as one that can use this one as an `Anchor`.
+
+
+#### Nonce Validation
+
+1. [Validate](#program-validation) `Program` with the given `Arguments`.
+2. Verify that `Issuance` points to an issuance that is present in the transaction (meaning visitable by traversing `Results` and `Sources` from the transaction header) and whose `Anchor` is equal to this nonce's ID.
+3. Verify that both mintime and maxtime in the `TimeRange` are not zero.
```
bobg: Also recursively validate the TimeRange?

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1104,59 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
 
 Code  | Stack Diagram  | Cost
 ------|----------------|-----------------------------------------------------
-0xc2  | (∅ → assetid)   | 1; [standard memory cost](#standard-memory-cost)
+0xc2  | (∅ → assetid)  | 1; [standard memory cost](#standard-memory-cost)
+
+If the current entry is an [issuance](blockchain.md#issuance-1) or a [spend](blockchain.md#spend-1) entry, pushes the `Value.AssetID` of that entry.
 
-Pushes the asset ID assigned to the current input on the data stack.
+If the current entry is a [nonce](blockchain.md#nonce) entry, verifies that the `AnchoredEntry` field is an [issuance](blockchain.md#issuance-1) entry, and pushes the `Value.AssetID` of that issuance entry.
```
bobg: ...and fails if `AnchoredEntry` is not an issuance?

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1104,59 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
+Fails if the entry is not a [mux](blockchain.md#mux-1), an [issuance](blockchain.md#issuance-1) or a [spend](blockchain.md#spend-1).
 
 #### ASSET
 
 Code  | Stack Diagram  | Cost
 ------|----------------|-----------------------------------------------------
-0xc2  | (∅ → assetid)   | 1; [standard memory cost](#standard-memory-cost)
+0xc2  | (∅ → assetid)  | 1; [standard memory cost](#standard-memory-cost)
+
+If the current entry is an [issuance](blockchain.md#issuance-1) or a [spend](blockchain.md#spend-1) entry, pushes the `Value.AssetID` of that entry.
```
bobg: A spend doesn't have a `Value.AssetID`, but it does have a `SpentOutput.Source.Value.AssetID`.

bobg: (Same for `Value.Amount` in the description of `AMOUNT`, below.)

docs/protocol/specifications/validation.md
```diff
@@ -294,55 +131,47 @@ Note: requirement for the input and output sums to be below 2<sup>63</sup> impli
...
 
-1. Let S be the input blockchain state.
-2. For each transaction in the block:
-    1. [Apply the transaction](#apply-transaction) to S, yielding a new state S′.
-    2. Replace S with S′.
-3. Replace the block header in S with the input block header, yielding a new state S′.
-4. Replace S with S′.
-5. Remove elements of the issuance memory in S where the expiration timestamp is less than the block’s timestamp, yielding a new state S′.
-6. Return S′.
+1. Evaluate the [consensus program](blockchain.md#block-header):
```
bobg: This is redundant with step 2 (validate the block - which should really be "validate the block header"), which includes evaluation of the consensus program.

docs/protocol/specifications/validation.md
```diff
@@ -294,55 +131,47 @@ Note: requirement for the input and output sums to be below 2<sup>63</sup> impli
...
+    1. [Create a VM 1](vm1.md#vm-state) with initial state and expansion flag set to `false`.
+    2. [Prepare VM](vm1.md#prepare-vm) with program arguments from the block witness.
+    4. Set the VM’s program to the consensus program as specified by the blockchain state’s block header.
+    5. Execute [Verify Predicate](vm1.md#verify-predicate) operation. If it fails, halt and return false.
+2. [Validate the block](blockchain.md#block-header-validation) with “previous block header” set to the block header in the current blockchain state; if invalid, halt and return blockchain state unchanged.
+3. Let `S` be the input blockchain state.
+4. For each transaction in the block, in order:
+    1. [Apply the transaction](#apply-transaction) using the input block’s header to blockchain state `S`, yielding a new state `S′`.
+    2. If transaction failed to be applied (did not change blockchain state), halt and return the input blockchain state unchanged.
+    3. Test that [assets merkle root](blockchain.md#assets-merkle-root) of `S′` is equal to the assets merkle root declared in the block commitment; if not, halt and return blockchain state unchanged.
```
bobg: The assets merkle root of `S′` will evolve as each transaction is applied, but the one in the block commitment (should ready "body") will not. Sounds like this step is not a sub-bullet of (4), but should be its own next top-level bullet. Is that right?

(That's how [it's implemented now](https://github.com/chain/chain/blob/374d08480f2f78562593fa5b5c1f544088fc4a41/protocol/validation/block.go#L86), even though [the present spec](https://github.com/chain/chain/blob/main/docs/protocol/specifications/validation.md#validate-block) gets this wrong too.)

docs/protocol/specifications/vm1.md
oleganza: VM version is not as upgradable in blocks as in txs. You can't just bump the VM version and old clients will go "ok, i don't know how this is authenticated, but i'll take that block". There is no "higher authority" to defer to above the block signers. See detailed discussion here: https://chain.com/docs/protocol/papers/blockchain-extensibility#consensus-upgrades

docs/protocol/specifications/vm1.md
oleganza: >  whether or not this is a hash is an app-level decision

The spec says "Data is a Hash field type". I understand that it could not be a hash, but just some 32-byte string, but then we should probably redefine that field. However, I'd rather not define it as varstring to highlight that any actual data longer than 32 bytes should go somewhere else in the transport layer.

docs/protocol/specifications/vm1.md
oleganza: We do not distinguish "executed" vs "encountered" as of Oct2016 release, AFAIR. What wording would you suggest?

docs/protocol/specifications/vm1.md
oleganza: Ouch, that's redundant. We can just ignore `prog` and `version` for now, and restrict them to be empty string and `1` if expansion flag is `false`. Will fix.

docs/protocol/specifications/vm1.md
oleganza: fixed.

docs/protocol/specifications/vm1.md
oleganza: fixed

docs/protocol/specifications/vm1.md
oleganza: Added clarification here: https://github.com/chain/chain/pull/728/commits/9158fdcbead679db64b478a54d731904a091eb97#diff-a108b3a30ca8a7fff9c31042faeb0761R250

We _will_ have unexecuted instructions due to random jumps. And that's fine.

docs/protocol/specifications/vm1.md
oleganza: Ah, now i understand. We don't have "entry type" field here, so we need to carefully specify how one can enforce either retirement or a normal output.

oleganza: Done: https://github.com/chain/chain/pull/728/commits/cb52405e02f4c67959008bb1f429d8bc6f0f5457

docs/protocol/specifications/vm1.md
oleganza: fixed

docs/protocol/specifications/vm1.md
oleganza: WDYT https://github.com/chain/chain/pull/728/commits/afe36cc051af8f99f8e3dbfaf595c037006d5138

docs/protocol/specifications/vm1.md
oleganza: All NOPs are simply "expansion opcodes" together will all unassigned opcodes. This has been since Oct 2016.

WDYT: https://github.com/chain/chain/pull/728/commits/c0e2379468c8dfccd7408be0425020402789d16f

docs/protocol/specifications/blockchain.md
oleganza: Fixed spec: https://github.com/chain/chain/pull/728/commits/adc5bbf0b4c3c7ccc2657f7fe61881929a434a65

docs/protocol/specifications/blockchain.md
oleganza: Was always strictly greater. Clarified: https://github.com/chain/chain/pull/728/commits/18b9c0f002efc6db9b652075a2539b8a60cabe07

docs/protocol/specifications/blockchain.md
oleganza: Cleaned this up: https://github.com/chain/chain/pull/728/commits/582f85721ca5b24a6a7b6946d42a568e019fad72

docs/protocol/specifications/blockchain.md
oleganza: Fixed: https://github.com/chain/chain/pull/728/commits/b3bce610bce961224d7f07eca281cd095c719b70

docs/protocol/specifications/blockchain.md
oleganza: Fixed: ccc3536

docs/protocol/specifications/blockchain.md
oleganza: Clarified here for the future us: 7fbed53

docs/protocol/specifications/blockchain.md
oleganza: I agree with the principle, but not sure if that would lead to an overkill. Could we see organizing that data as an implementation detail, irrelevant to the logic of the spec?

docs/protocol/specifications/blockchain.md
oleganza: Good point. I will move context-specific stuff to `Apply Transaction` in `validation.md`.

oleganza: Moved here: ea9593d

@danrobinson Bob has a good point re:unconfirmed txs.

docs/protocol/specifications/blockchain.md
oleganza: Removed in c19f125

docs/protocol/specifications/blockchain.md
oleganza: That's validated by `Issuance`: 

> Verify that `Anchor.AnchoredEntry` points to this entry.

docs/protocol/specifications/blockchain.md
oleganza: Good point! Fixed: ede0531

docs/protocol/specifications/vm1.md
oleganza: Yup! Added: b926760

docs/protocol/specifications/vm1.md
oleganza: Fixed. 84e793a

docs/protocol/specifications/validation.md
oleganza: Yup. Cleaned up and clarified: ca14b22

docs/protocol/specifications/validation.md
oleganza: OMGZZZ. You are 100% correct, how did this exist for 5 months without being noticed? 

Fixed: dea297d

docs/protocol/specifications/vm1.md
bobg: I don't mean make it a varstring; I mean keep it as a 32-byte string but call it something other than a hash. A `string32` for instance (which happens to be suitable for storing 256-bit hashes).

docs/protocol/specifications/vm1.md
bobg: Rather than a clarification elsewhere, I suggest:

"If the flag is off, these opcodes immediately fail the program when encountered during execution."

Then mention in the descriptions of `JUMP` and `JUMPIF` that this may cause some instructions to not be executed, and therefore to not trigger this rule.

docs/protocol/specifications/vm1.md
```diff
@@ -1066,9 +1074,11 @@ Code  | Stack Diagram                  | Cost
 ------|--------------------------------|-----------------------------------------------------
 0xae  | (∅ → hash)                     | 256 + [standard memory cost](#standard-memory-cost)
 
-Computes the [transaction signature hash](data.md#transaction-signature-hash) corresponding to the current input.
+Computes the transaction signature hash corresponding to the current entry. Equals [SHA3-256](blockchain.md#sha3) of the concatenation of the current [entry ID](blockchain.md#entry-id) and [transaction ID](blockchain.md#transaction-id):
```
bobg: Need to define "current entry." The execution context described above mentions the index of one of the transaction's inputs, but that's outdated; the current entry could also be an interior node, such as  a mux.

docs/protocol/specifications/vm1.md
oleganza: Good point. Added.

docs/protocol/specifications/vm1.md
oleganza: Fixed, thanks.

docs/protocol/specifications/vm1.md
oleganza: Done. TXDATA and ENTRYDATA.

bobg:
LGTM

Would anyone else like to add an approval before this PR lands?

bobg:
Sorry, !LGTM

I overlooked these unaddressed review comments: https://github.com/chain/chain/pull/728#discussion_r110284417 and https://github.com/chain/chain/pull/728#discussion_r110284821

docs/protocol/specifications/validation.md
```diff
@@ -294,55 +131,46 @@ Note: requirement for the input and output sums to be below 2<sup>63</sup> impli
...
-5. Return true.
+1. Validate transaction using the checks below. If any check fails, halt and return the input blockchain state unchanged.
+    1. If the block header version is 1, verify that transaction version is equal to 1.
+    2. If the `Mintime` is greater than zero: verify that it is less than or equal to the block header timestamp.
+    3. If the `Maxtime` is greater than zero: verify that it is greater than or equal to the block header timestamp.
+    4. [Validate transaction header](blockchain.md#transaction-header-validation).
+2. Let `S` be the input blockchain state.
+3. For each visited [nonce entry](blockchain.md#nonce) in the transaction:
+    1. If [nonce ID](blockchain.md#entry-id) is already stored in the nonce set of the blockchain state, halt and return the input blockchain state unchanged.
+    2. Add ([nonce ID](blockchain.md#entry-id), transaction maxtime) to the nonce set in `S`, yielding a new state `S′`.
```
bobg: Is the transaction maxtime the right expiration time for the nonce, or should it be the nonce's own timerange.maxtime? (Or the max of those two numbers? Or something else?)

bobg:
Ping - just a couple of outstanding issues to address, then we can land this big spec change.

docs/protocol/specifications/validation.md
bobg: See also https://github.com/chain/chain/issues/1128

oleganza: yeah, good catch this should be the nonce's maxtime.

oleganza: fixed: 1be3d56023b043067c91c17263e9c750ba2bdbca

docs/protocol/specifications/vm1.md
```diff
@@ -1121,7 +1121,9 @@ Code  | Stack Diagram                                        | Cost
     2. If the entry satisfies all of the following conditions pushes [true](#vm-boolean) on the data stack; otherwise pushes [false](#vm-boolean):
         1. the destination entry is an [output](blockchain.md#output-1) or a [retirement](blockchain.md#retirement-1),
         2. if the destination is an output: control program equals `prog` and VM version equals `version`,
-        3. if the destination is a retirement: `prog` is an empty string and `version` is zero,
+        3. if the destination is a retirement:
+            * `version` must be zero,
+            * if [expansion flag](#vm-sate) is `false`, `prog` must be an empty string.
```
oleganza: So the risk here is:

Core was updated with new txhash computation via entries, but for retirement-using transactions SDK/API was still creating signature programs `CHECKOUTPUT(vm=1,prog="FAIL*")`. If we update validation logic to require `CHECKOUTPUT(vm=0,prog="")` we'd make these checks fail which means we'd need a blockchain reset.

But if we allow `CHECKOUTPUT(vm=1,prog="FAIL*")` for compatibility, we'll introduce a malleability — someone can satisfy that requirement with both normal retirement and ugly retirement (that clobbers UTXO set). Which may break some application-level accounting logic. To mitigate that we can soft-fork a rule that forbids FAIL-prefixed outputs outright so people have to use proper retirements.

oleganza: Addressed here: df23139

Muxes and Outputs forbid programs `vm0;*` and `vm1;FAIL*` so that you cannot trick `CHECKOUTPUT(vm1;FAIL*)` to accept an output instead of a retirement.

/cc @danrobinson

bobg:
OK, this time for real: LGTM, would anyone else like to add their review before this lands?

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1114,61 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
-    2. VM version equals `version`,
-    3. asset ID equals `assetid`,
-    4. amount equals `amount`,
-    5. `refdatahash` is an empty string or it matches the [SHA3-256](data.md#sha3) hash of the reference data.
+5. If the current entry is a [Mux](blockchain.md#mux-1):
+    1. Finds a [destination entry](blockchain.md#value-destination-1) at the given `index`.
+    2. If the entry satisfies all of the following conditions pushes [true](#vm-boolean) on the data stack; otherwise pushes [false](#vm-boolean):
+        1. the destination entry is an [output](blockchain.md#output-1) or a [retirement](blockchain.md#retirement-1),
+        2. if the destination is an output: control program equals `prog` and VM version equals `version`,
+        3. if the destination is a retirement:
```
oleganza: Done!

docs/protocol/specifications/vm1.md
```diff
@@ -1100,44 +1114,61 @@ Note: [standard memory cost](#standard-memory-cost) is applied *after* the instr
...
+            * `version` must be 1,
```
oleganza: I guess none because the current implementation of constraints (that use CHECKOUTPUT) copies output data to the conditions (so it's `VM1;FAIL` for retirements).

oleganza:
@bobg replied to your last 2 questions.

docs/protocol/specifications/vm1.md
oleganza: Relevant snippet: https://github.com/chain/chain/blob/47206b41b689ec03396f6d9223218a16cc500297/core/txbuilder/witness.go#L170-L174

jeffomatic:
It sounds like we want to land this branch before the official 1.2.0 release. It's got an LGTM from @bobg already, does anyone else want to take a look?

kr:
LGTM

PR #729, opened 2017-03-08
sdk/java: only advance URL on temporary errors

jbowens:
If a call throws an APIException, we should only advance the client
URL index if the error is a temporary, retriable error. Permanent
errors shouldn't cause us to move on to the next Core URL.


boymanjor:
lgtm

PR #732, opened 2017-03-09
installer/mac: update to version 1.1.1

jeffomatic:
This replaces a previous commit that linked macapp 1.1.1 to server
1.1.2. We ended up using server 1.1.3.

This update should also be made to main, to keep main up-to-date.


jeffomatic:
PTAL @dominic

PR #739, opened 2017-03-09
core/account: carry `change` onto `account_utxos`

jbowens:
When indexing account UTXOs, carry the `change` flag onto the
`account_utxos` SQL table from `account_control_programs`. This
will allow the query block processor to use exclusively `account_utxos`
for annotating transactions.

Note that this migration creates a new, non-nullable column. The
migration atomically backfills the column. However, in a rolling deploy,
an older cored binary may try to insert nulls into the `change` column.
This is okay because new rows are only ever inserted by the account
block processor. If the block processor is running on an older 1.1.x
cored, the block processor will error out and be reattempted until it
succeeds. It'll succeed when a 1.2.x process becomes leader and
reattempts the block processor.

Practically, this means the processing of blocks may be delayed for
the duration of the rolling deploy during an upgrade to 1.2.

### DIFF
--- a/core/account/indexer.go
+++ b/core/account/indexer.go
@@ -94,6 +94,7 @@ type accountOutput struct {
 	rawOutput
 	AccountID string
 	keyIndex  uint64
+	change    bool
 }
 
 func (m *Manager) ProcessBlocks(ctx context.Context) {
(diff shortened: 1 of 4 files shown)


kr:
LGTM

PR #740, opened 2017-03-10
testutil: add cycle detection to DeepEqual

bobg:
This is a carved-off piece of the `entries-validation` branch, landing in anticipation of that because txentries data structures are not acyclic.


bobg:
PTAL

bobg:
PTAL

kr:
LGTM

PR #741, opened 2017-03-10
core/txdb: test reading/writing snapshot issuances

jbowens:
Add a unit test ensuring that writing a snapshot with issuance memory and
reading that snapshot back out preserves the issuance memory. There are
already analagous tests for the state tree.


bobg:
LGTM

PR #742, opened 2017-03-10
core/coretest: add missing pins to CreatePins

jbowens:
These two new pins were added in a97b23 and 872e19 but weren't
added to the `coretest.CreatePins` function. The account package's
ProcessBlocks function will start all three of its block processors,
so all three of the pins should be initialized.


bobg:
PTAL

PR #743, opened 2017-03-10
protocol/bc: export various identifiers

bobg:
This is in anticipation of landing the `entries-validation` branch, which uses some of the new data structures in `protocol/bc` in other packages.

### DIFF
--- a/protocol/bc/output.go
+++ b/protocol/bc/output.go
@@ -19,7 +19,7 @@ func (o *Output) Body() interface{} { return o.body }
 
 func (o Output) Ordinal() int { return o.ordinal }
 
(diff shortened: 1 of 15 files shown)


bobg:
PTAL

kr:
:eyes:

protocol/bc/output.go
```diff
@@ -1,26 +1,26 @@
 package bc
 
-type output struct {
+type Output struct {
```
kr: The exported identifiers really ought to have comments. Especially this one to explain how it's different from TxOutput.

bobg:
Updated, PTAL

jbowens:
lgtm

PR #744, opened 2017-03-10
core/account: annotate outputs using `account_utxos`

jbowens:
Use the `account_utxos` table instead of the `account_control_programs`
table for annotating outputs with account data. We already used it for
annotating inputs but were waiting on the addition of the `change`
column to use it for annotating outputs too.


kr:
LGTM

PR #745, opened 2017-03-10
protocol/state: remove PriorIssuances type

jbowens:
Remove the PriorIssuances type. We only ever use it as a
`map[bc.Hash]uint64`.


kr:
:+1:

PR #746, opened 2017-03-10
log: add function AddPrefixkv

kr:
Other packages can use AddPrefixkv to add arbitrary
items to the log prefix. A future change will use this
to make chain/net/http/reqid depend on chain/log,
instead of the other way around. This will remove
special-case code in package log, and enable reqid to
use the logging functions defined here.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -21,6 +21,9 @@ import (
 
 const rfc3339NanoFixed = "2006-01-02T15:04:05.000000000Z07:00"
 
+// context key type
+type key int
+
 var (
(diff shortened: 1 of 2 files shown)


kr:
PTAL

I kind of want to bikeshed the function name. AddPrefixkv, AppendPrefixkv, WithPrefixkv? Not such a fan of With- because this function doesn't fully replace the existing prefix, it only appends to it.

jbowens:
AddPrefixkv sgtm

PR #748, opened 2017-03-11
net/http/reqid: add log prefix

kr:
With this change, package chain/net/http/reqid uses
log.AddPrefixkv to add the request ID to the log prefix
in the context.

It also removes special-case code from chain/log to add
the request ID there.

This also reverses the dependency between these two
packages, and allows reqid to use other logging
functions from chain/log instead of needing to use
standard library package log.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -137,7 +137,7 @@ func Printkv(ctx context.Context, keyvals ...interface{}) {
 
 	// Prepend the log entry with auto-generated fields.
 	out := fmt.Sprintf(
-		"%s=%s %s=%s %s=%s",
+		"%s=%s %s=%s",
 		KeyCaller, vcaller,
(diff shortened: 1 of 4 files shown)


log/log.go
```diff
@@ -143,17 +138,9 @@ func Printkv(ctx context.Context, keyvals ...interface{}) {
 	// Prepend the log entry with auto-generated fields.
 	out := fmt.Sprintf(
 		"%s=%s %s=%s %s=%s",
-		KeyReqID, formatValue(reqid.FromContext(ctx)),
```
jbowens: I think you need to update the format string above too

kr:
Addressed feedback; PTAL.

kr:
Oh I guess maybe I didn't need to write that test, go vet caught it. Thanks, go vet!

jbowens:
🥇

PR #749, opened 2017-03-11
core/query: remove unused AnnotatedAsset fields

jbowens:
The VMVersion and RawDefinition fields on the query.AnnotatedAsset
struct were not used or populated except in the response of a
/create-asset request. The SDKs also do not expose these fields.


kr:
:cool:

PR #750, opened 2017-03-11
log: omit functions from 'at' field by name

kr:
Each log line has an entry of the form at=main.go:123
added to it by Printkv. It's not helpful for this
file-line information to point to a logging helper
function (such as log.Printf or core.logHTTPError), so
there were mechanisms to supply an alternative:

- Outside of package log, helper functions were expected
to include an at=foo.go:456 key-value pair to override
the item added by Printkv. In practice, none ever did.
The code to obtain this information is fiddly and gross.

- Within package log, function 'caller' took a 'skip'
argument specifying how many frames to skip before
returning file-line information. This was used
inconsistently. Helper functions Printf and Error used
it to add override entries; Fatalkv and
RecoverAndLogError should have but didn't.

Hopefully, the new mechanism in this patch is easy
enough to use that it'll be used appropriately.

The new approach keeps a lookup table of names of
functions to avoid. Package log then walks up the call
stack, skipping any filtered functions, and prints to
the log output the first unfiltered location it finds.

### DIFF
--- a/core/http.go
+++ b/core/http.go
@@ -12,8 +12,8 @@ import (
 )
 
 func init() {
-	log.FilterFunc("chain/core.logHTTPError")
-	log.FilterFunc("chain/core.WriteHTTPError")
+	log.SkipFunc("chain/core.logHTTPError")
+	log.SkipFunc("chain/core.WriteHTTPError")
 }
 
 // errBadReqHeader indicates the user supplied a malformed request header,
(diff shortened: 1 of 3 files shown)


kr:
PTAL

kr:
On second thought I think we should avoid "filter" as a verb. It is ambiguous whether things are being included or excluded. Better to say "avoid" or "omit" or something. Revision coming up.

kr:
Updated, PTAL.

jbowens:
> Better to say "avoid" or "omit" or something.

Maybe for the commit title too?

jbowens:
lgtm

PR #751, opened 2017-03-11
protocol/bc, protocol/vm, protocol/validation: rewrite validation in terms of new objects

bobg:


bobg:
rebased

bobg:
We won't need this intermediate step (legacy validation written in terms of txentries). The real action is in _new_ validation (in terms of txentries), to which this PR was a prelude.

PR #754, opened 2017-03-13
protocol/patricia: use set semantics in tests

jbowens:
The external interface for the patricia tree already exposes set
semantics instead of key-value pairs. The package unit tests still use
the unexported `insert` that took a key value pair. This commit rewrites
those tests to use the exported `Insert` and deletes the now redundant
`insert` method.


kr:
:eyes:

kr:
LGTM

PR #755, opened 2017-03-14
protocol: remove Store getter method

jbowens:
The `protocol.Chain.Store` method for getting a `Chain`'s underlying Store
is only used in one place, `core/fetch.BootstrapSnapshot`. We have
access to the store at its callsite, so pass the store in instead. This
keeps the `protocol.Chain` type cleaner. Generally, the store should
only be used by the `Chain` type itself.


kr:
LGTM

PR #756, opened 2017-03-14
net/http/reqid: remove unused exported constant

kr:
Nothing ever uses this "default value" for a missing
request ID. It would be just as well to use the empty
string to represent that case. This is especially true
now that we no longer print a reqid=xxx field at all in
the log when the context doesn't contain a reqid.


kr:
PTAL

bobg:
LGTM

PR #757, opened 2017-03-14
core: fix TestRecovery

jbowens:
TestRecovery has been broken for a while. It only runs when the `LONG=1`
environment variable is set, so it's easy to accidentally break.

### DIFF
--- a/core/recovery_test.go
+++ b/core/recovery_test.go
@@ -153,6 +153,7 @@ func TestRecovery(t *testing.T) {
 			if err != nil {
 				t.Errorf("generateBlock returned err %s", err)
 			}
+			cancel()
 			close(completed)
 		}()
 
--- a/core/recovery_test.go
+++ b/core/recovery_test.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/hex"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"runtime"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


jbowens:
PTAL

bobg:
:eyes:

core/recovery_test.go
```diff
@@ -86,53 +120,65 @@ func TestRecovery(t *testing.T) {
...
+		select {
+		case <-crashed:
+			t.Log("goroutine crashed on query")
+		case <-completed:
+			t.Log("goroutine completed successfully")
 		}
-		if ok {
+
+		if atomic.LoadInt64(&calls) < n {
 			// The driver never crashed the goroutine, so n is now greater than
```
bobg: This old comment should be updated the reflect the new code: "`calls` never reached `n` so the goroutine completed without simulating a crash"

core/recovery_test.go
```diff
@@ -86,53 +120,65 @@ func TestRecovery(t *testing.T) {
...
 			// the total number of queries performed during `generateBlock`.
-			databaseDumps = append(databaseDumps, pgtest.Dump(t, cloneURL, false, "*_id_seq"))
+			databaseDumps = append(databaseDumps, pgtest.Dump(t, cloneURL, false, "*_id_seq", "block_processors"))
 			break
 		}
 
 		// We crashed at some point during block generation. Do it again,
 		// without crashing.
-		err = generateBlock(ctx, t, wrappedDB, timestamp, poolTxs)
+		err = generateBlock(context.Background(), t, wrappedDB, timestamp, poolTxs)
```
bobg: Could reuse `ctx` here.

bobg:
LGTM with a couple of small comments.

core/recovery_test.go
jbowens: ah, I forgot about `runCtx`. I guess `runCtx` should be cancelled but `ctx` should be okay.

PR #758, opened 2017-03-14
database/pg/pgtest: invoke WrapDB func in Prepare

jbowens:
Alter `pgtest.WrapDB` to call the provided closure during
`driver.Conn.Prepare` instead of in the `driver.Stmt` functions.
We can do this now that Go 1.8 properly handles panics in
`driver.Conn.Prepare`:

https://go-review.googlesource.com/#/c/23576/


kr:
LGTM

PR #760, opened 2017-03-15
cmd/cored: use revid constant for dev builds

kr:
This gives us more precise information on what code
contributed to a fresh-from-git build, and slightly
simplifies the logic for initializing the version string
on startup.


kr:
I'll follow up with something along these lines for corectl in another PR.

jbowens:
lgtm

PR #761, opened 2017-03-15
all: use pg.DB

jbowens:
Most places in the code base take a `pg.DB` in their external
interfaces. Some take a `*sql.DB`. Update the remaining instances of
`*sql.DB` to `pg.DB`.

This was motivated by trying to call `accounts.NewManager` with a
`pg.DB` and haing to do a `db.(*sql.DB)` cast.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -226,7 +226,7 @@ func runServer() {
 	}
 }
 
-func launchConfiguredCore(ctx context.Context, db *sql.DB, conf *config.Config, processID string) http.Handler {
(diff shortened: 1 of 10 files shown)


kr:
Wow, looks like we don't open SQL transactions anywhere any more?! (Previously, we would have to use *sql.DB in places that needed to open a transaction, or *sql.Tx in places that needed to run inside a dbtx (but couldn't open it themselves for any reason), or pg.DB for places that didn't care whether it ran in a dbtx or not.)

kr:
LGTM

PR #762, opened 2017-03-15
cmd/slashland: fix revid computation

kr:
Two fixes here:

1. The revid is now measured on the base branch,
which is not always main. If a local main has never been
checked out, the rev-list command fails, so we must use
the remote ref (e.g. "origin/main" or "origin/1.1-stable").

2. Include the branch name in the revid string. This
lets us distinguish between rev123 on main vs rev123 on
1.1-stable (which can be a different commit).

### DIFF
--- a/cmd/slashland/main.go
+++ b/cmd/slashland/main.go
@@ -184,7 +184,7 @@ func land(req *landReq) {
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


kr:
PTAL

jbowens:
lgtm

PR #763, opened 2017-03-15
core/leader: try more frequently; failover faster

jbowens:
Update the leader failover to update every 500 milliseconds. Allow a
new cored to takeover if the existing leader hasn't updated within 1
second.

Note, in practice, this means that when upgrading to 1.2 in a rolling
deploy, the first deployed 1.2 cored will quickly become leader.

Tested this locally by running a java program that creates
transactions sequentially. Tried killing cored and failing over
between the two.


kr:
LGTM

PR #765, opened 2017-03-15
cmd/slashland: avoid making empty commits

kr:
Before this change, we would blindly add a commit on
every rebase or attempt to land. Now, if the commit
would be empty, we skip it. This lets us land faster
when the HEAD commit has already been through CI.


kr:
Tested on https://github.com/chain/chainprv/pull/2517.

jbowens:
🙌

PR #766, opened 2017-03-15
cmd/corectl: print additional version info

kr:
This output mirrors what cored prints with its -version
flag. We also take this opportunity to switch from a
hard-coded "latest release version" string to using the
automatically-generated revid.


kr:
PTAL

jbowens:
lgtm

PR #767, opened 2017-03-15
core: add Run, RunUnconfigured

jbowens:
Add new Run and RunUnconfigured functions. Unexport all of
`core.API`'s fields. This moves some of the cmd/cored functionality into
the `core` package. More can probably be moved, but I wanted to avoid
doing too much at once.

Tested locally with a couple transactions in dashboard. Both running as a
generator and connecting to testnet.


kr:
:eyes:

cmd/cored/main.go
```diff
@@ -236,205 +229,103 @@ func launchConfiguredCore(ctx context.Context, db pg.DB, conf *config.Config, pr
...
+
+	// Allow loopback/localhost requests in Developer Edition.
+	opts = append(opts, core.AlternateAuth(authLoopbackInDev))
+	opts = append(opts, core.IndexTransactions(*indexTxs))
+	// Add any configured API request rate limits.
+	if *rpsToken > 0 {
+		opts = append(opts, core.RateLimit(limit.AuthUserID, 2*(*rpsToken), *rpsToken))
+	}
+	if *rpsRemoteAddr > 0 {
+		opts = append(opts, core.RateLimit(limit.AuthUserID, 2*(*rpsRemoteAddr), *rpsRemoteAddr))
```
kr: limit.RemoteAddrID?

core/api_test.go
```diff
@@ -195,7 +195,8 @@ func TestMux(t *testing.T) {
 		}
 	}()
 	m := &MockHSMHandler{}
```
kr: Maybe MockHSMHandler should become a LaunchOption? Maybe in a future PR?

cmd/cored/main.go
```diff
@@ -236,205 +229,103 @@ func launchConfiguredCore(ctx context.Context, db pg.DB, conf *config.Config, pr
...
+	}
+	// If the Core is configured as a block signer, add the sign-block RPC handler.
+	// TODO(jackson): Should this be handled using the callback passed to
+	// API.Handler like how we add the mockHSM routes?
```
kr: I am :-1: on this alternative. Using LaunchOption makes perfect sense for this IMO. In fact, it might also make sense to do the same for mockHSM routes too (noted there as well).

core/launch.go
```diff
@@ -0,0 +1,225 @@
...
+	expireReservationsPeriod = time.Second
+)
+
+// LaunchOption describes a runtime configuration option.
+type LaunchOption func(*API)
+
+// AlternateAuth configures the Core to use authFn to authenticate
+// incoming requests in addition to the default access token authentication.
+func AlternateAuth(authFn func(*http.Request) bool) LaunchOption {
+	return func(a *API) { a.altAuth = authFn }
```
kr: Based on my reading of API.altAuth and apiAuthn.alt, currently this LaunchOption is actually required yeah? If the caller didn't provide it, I think the core would crash.

I think if we're using this pattern it should be a rule that anything that's an "option" needs to be optional, as in, you can leave it out and things still work.

cmd/cored/main.go
```diff
@@ -236,205 +229,103 @@ func launchConfiguredCore(ctx context.Context, db pg.DB, conf *config.Config, pr
...
 			Username:     processID,
 			CoreID:       conf.ID,
 			BuildTag:     buildTag,
 			BlockchainID: conf.BlockchainID.String(),
 		}
-		submitter = &txbuilder.RemoteGenerator{Peer: remoteGenerator}
-	} else {
-		gen = generator.New(c, generatorSigners, db)
-		submitter = gen
+		opts = append(opts, core.RemoteGenerator(remoteGen))
```
kr: nit: it would actually read pretty nicely imo to inline remoteGen here.

cmd/cored/main.go
```diff
@@ -236,205 +229,103 @@ func launchConfiguredCore(ctx context.Context, db pg.DB, conf *config.Config, pr
...
 	}
 
-	// Set up the pin store for block processing
-	pinStore := pin.NewStore(db)
-	err = pinStore.LoadAll(ctx)
+	// Launch the Core. This will startup the various Core subsystems,
```
kr: nit:

startup = noun
start up = verb

core/launch.go
```diff
@@ -0,0 +1,225 @@
...
+	"chain/protocol/bc"
+)
+
+const (
+	blockPeriod              = time.Second
+	expireReservationsPeriod = time.Second
+)
+
+// LaunchOption describes a runtime configuration option.
+type LaunchOption func(*API)
```
kr: I don't want to bikeshed these names too much, but just a little…

```
func Run
func RunUnconfigured
type Option
```

???

or perhaps RunOption instead of Option to keep them more strongly associated like you did here.

I'm also down with these names as-is.

core/launch.go
```diff
@@ -0,0 +1,225 @@
...
+		a.submitter = gen
+	}
+}
+
+// RemoteGenerator configures the launched Core to fetch blocks from
+// the provided remote generator.
+func RemoteGenerator(client *rpc.Client) LaunchOption {
+	return func(a *API) {
+		a.remoteGenerator = client
+		a.submitter = &txbuilder.RemoteGenerator{Peer: client}
```
kr: Maybe LocalGenerator and RemoteGenerator should each clear out the config from the other? So generator = nil here and remoteGenerator = nil there. That would defensively prevent getting into a weird inconsistent state in case we somehow applied both LaunchOptions. Orrrr, maybe we should check for that case here and panic immediately.

Maybe a generalization of the principle is that options should prevent the *API from ever getting into an inconsistent state (including starting out in an inconsistent state, as with AlternateAuth above). Whether by clobbering each other or by detecting and panicking.

core/transact.go
```diff
@@ -174,13 +174,10 @@ func recordSubmittedTx(ctx context.Context, db pg.DB, txHash bc.Hash, currentHei
...
 
-// CleanupSubmittedTxs will periodically delete records of submitted txs
+// cleanupSubmittedTxs will periodically delete records of submitted txs
 // older than a day. This function blocks and only exits when its context
 // is cancelled.
-//
-// TODO(jackson): unexport this and start it in a goroutine in a core.New()
-// function?
-func CleanupSubmittedTxs(ctx context.Context, db pg.DB) {
+func cleanupSubmittedTxs(ctx context.Context, db pg.DB) {
```
kr: same nit (as long as this function name is changing anyway):

cleanup = noun
clean up = verb

core/launch.go
jbowens: Oops, good catch. How about I make it truly optional in apiAuthn?

jbowens:
Updated, PTAL

core/run.go
```diff
@@ -0,0 +1,222 @@
...
+			panic("core configured with local and remote generator")
+		}
+		a.generator = gen
+		a.submitter = gen
+	}
+}
+
+// RemoteGenerator configures the launched Core to fetch blocks from
+// the provided remote generator.
+func RemoteGenerator(client *rpc.Client) RunOption {
```
kr: What do you think about GeneratorLocal and GeneratorRemote so they sort together in the godoc output? Nbd tho.

kr:
LGTM with one small idea

PR #768, opened 2017-03-15
core/account: use pgtest.NewTx where possible

jbowens:
Some of the `core/account` tests can use `pgtest.NewTx` now that
`NewManager` takes a `pg.DB`. This shaves a few seconds off the
package's test time.

```
PASS
ok  	chain/core/account	5.309s
```

vs

```
PASS
ok  	chain/core/account	9.114s
```


kr:
LGTM

PR #771, opened 2017-03-16
core: implement http.Handler on *API

jbowens:
Rework the construction of the API's http.Handler to avoid a second
function call (to API.Handler). Instead, construct the http.Handler
during Run. Also:
  * move the MockHSMHandler into a RunOption.
  * move the blockchain ID middleware into the core package

Tested locally through dashboard, including transacting on testnet, resetting, reconfiguring and creating mock HSM keys.


kr:
LGTM

PR #772, opened 2017-03-16
core: find executable using standard library function

kr:


kr:
PTAL

I ran it once, joined testnet, and reset my core.

jbowens:
lgtm

PR #773, opened 2017-03-16
vendor: update coreos/etcd

tessr:
Update our copy of coreos/etcd to `5015480e0c385e7fd9aa8825c26f18c6eba17395`. This picks up changes that are required for compatibility with Windows.


tessr:
PTAL 

I'd be surprised if this broke anything but I ran Chain Core and ran one transaction, just to be sure.

kr:
LGTM

PR #774, opened 2017-03-16
sdk/java: expose Client.Builder method to set SslSocketFactory

boymanjor:
This will allow client applications to build their own
SslSocketFactory and provide it to the SDK's http client.


boymanjor:
PTAL

kr:
LGTM

PR #776, opened 2017-03-17
core/leader: add State enum

jbowens:
Replace IsLeader() with an enum that can be FOLLOWING, RECOVERING or
LEADING. RECOVERING is used when a process has been elected leader but
has not yet finished recovering the blockchain and becoming leader.

Use this more fine-grained status to enable the /info endpoint even when
RECOVERING. This fixes snapshot downloads so that the /info endpoint can
be used to track the progress of a downloading snapshot.

It's OK for a Recovering process to call forwardToLeader. It'll detect that
it's trying to forward to itself and return a 503 status code.

I tested this locally by connecting to testnet through dashboard. I was able to
watch the snapshot download progress in the lefthand bar.

### DIFF
--- a/core/leader/leader.go
+++ b/core/leader/leader.go
@@ -4,6 +4,7 @@ package leader
 
 import (
 	"context"
+	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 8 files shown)


core/leader/leader.go
```diff
@@ -12,16 +13,40 @@ import (
...
+	Recovering
+	Leading
+)
+
+func (ps ProcessState) String() string {
+	switch ps {
+	case Following:
+		return "FOLLOWING"
+	case Recovering:
+		return "RECOVERING"
```
kr: My personal preference is to avoid ALL-CAPS but nbd.

kr:
LGTM

PR #777, opened 2017-03-17
core/fetch: refactor snapshot bootstrap

jbowens:
Refactor bootstrapping by snapshot to not rely on global
variables in core/fetch for reporting download progress.
Previously, as a proxy for detecting successful recovery
from the snapshot, there was an undocumented relationship
between fetch.BootstrapSnapshot and fetch.Fetch where
Fetch would mark the snapshot as completed:

```
// If we downloaded a snapshot, now that we've recovered and successfully		
// booted from the snapshot, mark it as done.		
if sp := SnapshotProgress(); sp != nil {		
	sp.done()		
}
```

Instead, return a progress type from BootstrapSnapshot
that can be stored on the core.API and used in /info calls.
Remove this value from the core.API after Chain.Recover
completes. This commit also reorganizes all bootstrapping
-specific code in core/fetch to a bootstrap.go file.

I tested this locally by connecting to testnet, observing
dashboard & curling http://localhost:1999/info.

### DIFF
--- a/core/fetch/bootstrap.go
+++ b/core/fetch/bootstrap.go
@@ -32,12 +32,16 @@ type SnapshotProgress struct {
 // or encounters any kind of validation error, it'll re-attempt the
 // snapshot download a few times.
 func (s *SnapshotProgress) Attempt() int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	return s.attempt
 }
 
 // Height returns the blockchain height of the snapshot being
 // downloaded.
 func (s *SnapshotProgress) Height() uint64 {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	return s.height
 }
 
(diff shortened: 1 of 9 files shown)


ryandotsmith:
🎰

jbowens:
PTAL

jbowens:
PTAL

bobg:
:eyes:

core/fetch/bootstrap.go
```diff
@@ -0,0 +1,206 @@
...
+	downloadProgress *progressReader
+
+	stopped chan struct{}
+}
+
+// Attempt returns the how many times Core has attempted to
+// download a bootstrap snapshot. If a download request times out
+// or encounters any kind of validation error, it'll re-attempt the
+// snapshot download a few times.
+func (s *SnapshotProgress) Attempt() int {
```
bobg: These accessors should acquire the mutex before reading `SnapshotProgress` fields.

jbowens: Good catch. Made em accessors but forgot to actually lock the mu

bobg:
LGTM

PR #778, opened 2017-03-17
core/leader: encapsulate state

jbowens:
Encapsulate all of the core/leader state into a struct instead of
using package-level global variables. Also, add a test for leader
failover since it's now possible to unit test.

### DIFF
--- a/core/api_test.go
+++ b/core/api_test.go
@@ -366,7 +366,7 @@ func toTxTemplate(ctx context.Context, inp map[string]interface{}) (*txbuilder.T
 
 type alwaysLeader struct{}
 
-func (al alwaysLeader) Address(_ context.Context) (string, error) {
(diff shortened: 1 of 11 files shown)


core/api_test.go
```diff
@@ -371,3 +363,13 @@ func toTxTemplate(ctx context.Context, inp map[string]interface{}) (*txbuilder.T
 	err = json.Unmarshal(jsonInp, tpl)
 	return tpl, err
 }
+
+type alwaysLeader struct{}
+
+func (al alwaysLeader) Address(_ context.Context) (string, error) {
```
kr: It's okay to write 

```go
func (alwaysLeader) Address(context.Context) (string, error)
```

(leaving out the variable name if none of the parameters in that set of parentheses are used)

kr:
Yay for unit tests \o/

PR #779, opened 2017-03-17
core/fileutil: new package

tessr:
package `fileutil` contains OS-compatible utilities for writing Chain Core's application data

### DIFF
--- a/core/fileutil/dir.go
+++ b/core/fileutil/dir.go
@@ -1,0 +1,12 @@
+//+build !windows,!darwin
+
+package fileutil
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func DefaultDir() string {
+	return filepath.Join(os.Getenv("HOME"), ".cored")
+}
--- a/core/fileutil/dir.go
+++ b/core/fileutil/dir.go
@@ -7,6 +7,6 @@ import (
 	"path/filepath"
 )
 
-func DefaultDir() string {
-	return filepath.Join(os.Getenv("HOME"), ".cored")
+func defaultDir() string {
+	return filepath.Join(os.Getenv("HOME"), ".chaincore")
 }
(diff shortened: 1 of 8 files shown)


core/fileutil/dir_darwin.go
```diff
@@ -0,0 +1,12 @@
+//+build darwin
```
boymanjor: TIL that naming the file `*_darwin.go` renders the `//+build darwin` tag redundant.

boymanjor:
LGTM, with a minor comment.

core/fileutil/dir.go
```diff
@@ -0,0 +1,12 @@
+//+build !windows,!darwin
+
+package fileutil
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func DefaultDir() string {
```
kr: I would suggest making one more file though, doc.go with the exported function that just calls the "real" one which is unexported.

kr: That way you can write the godoc just once, on the exported version.

core/fileutil/dir_windows.go
```diff
@@ -0,0 +1,10 @@
+package fileutil
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func DefaultDir() string {
+	return filepath.Join(os.Getenv("LOCALAPPDATA"), "Chain")
```
kr: Do you think it would be a good idea to put the word "core" somewhere in this path? I realize we are perhaps somewhat constrained because the installer maybe puts certain things here already. But we could add a subdirectory called `Chain Core` or `Core`. WDYT?

tessr: We can totally just call this directory Chain Core--nothing else is here at the moment. (Postgres should probably put its data files here too, but it doesn't, yet.)

PR #780, opened 2017-03-17
all: clean up

jbowens:
Clean up some small items found manually and by linters:

* blockPeriod, expireReservationsPeriod are no longer used in cmd/cored/main.go
* move remoteSigner struct definition to be beside its method declarations
* don't include newline in error message in md2html
* simplify boolean check in slashland
* tweak a few function comments to be in "[FunctionName] ..." format
* fix misspelling of "Receiver" in a comment
* check unchecked error value in TestRecovery
* avoid a couple of unnecessary conversions to *bc.Hash in ComputeTxHashes

### DIFF
--- a/cmd/md2html/main.go
+++ b/cmd/md2html/main.go
@@ -296,7 +296,7 @@ func writeCode(w io.Writer, path, snippet string) {
 func readSnippet(path, snippet string) (string, error) {
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
-		return "", fmt.Errorf("Unable to read source file: %s", path)
+		return "", fmt.Errorf("unable to read source file: %s: %s", path, err)
 	}
 
 	src := string(b)
--- a/cmd/md2html/main.go
+++ b/cmd/md2html/main.go
@@ -296,7 +296,7 @@ func writeCode(w io.Writer, path, snippet string) {
 func readSnippet(path, snippet string) (string, error) {
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/md2html/main.go
```diff
@@ -296,7 +296,7 @@ func writeCode(w io.Writer, path, snippet string) {
 func readSnippet(path, snippet string) (string, error) {
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
-		return "", fmt.Errorf("Unable to read source file: %s\n", path)
+		return "", fmt.Errorf("Unable to read source file: %s", path)
```
kr: While you're at it, how about s/Unable/unable/?

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

kr: Also, do we know if err itself was deliberately excluded from the message?

core/recovery_test.go
```diff
@@ -227,6 +227,9 @@ func generateBlock(ctx context.Context, t testing.TB, db pg.DB, timestamp time.T
 	}
 
 	c, err := protocol.NewChain(ctx, b1.Hash(), store, nil)
+	if err != nil {
+		return err
```
kr: errors.Wrap?

kr:
LGTM with a couple of little suggestions

cmd/md2html/main.go
jbowens: not sure. seems like it might not have been deliberate? i'll add it in

PR #781, opened 2017-03-17
core/pin: add unit tests

jbowens:
Improve the test coverage of the pin package. This brings statement
coverage from 27.9% to 82.1%.

### DIFF
--- a/core/pin/pin_test.go
+++ b/core/pin/pin_test.go
@@ -6,7 +6,62 @@ import (
 	"time"
 
 	"chain/database/pg/pgtest"
+	"chain/protocol/bc"
+	"chain/protocol/prottest"
+	"chain/testutil"
 )
+
+func TestCreatePin(t *testing.T) {
+	db := pgtest.NewTx(t)
+	ctx := context.Background()
+	store := NewStore(db)
+
+	err := store.CreatePin(ctx, "example", 100)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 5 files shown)


jbowens:
Writing the ProcessBlocks test made me realize that some of our block processors might have race conditions. I think some of them are assume that all block heights < _n_ are processed before the block processor for block _n_ is invoked. It is possible for blocks to be processed by the same processor in parallel. I think we allow a block processor to process up to 10 blocks in parallel. @erykwalder does this sound right?

kr:
So would it be fair to say that some processor P(n) for block n actually has a dependency on P(n-1)? Would it be appropriate to add that as an explicit dependency on itself to make it wait?

jbowens:
> So would it be fair to say that some processor P(n) for block n actually has a dependency on P(n-1)? Would it be appropriate to add that as an explicit dependency on itself to make it wait?

Yeah, I think that's right. I think the accounts and query processors both have this dependency. IIRC. the assets processor can process blocks in parallel but we might need to tweak the SQL query to make sure that we always save the earliest initial block height.

kr:
Or should we find a way to make it so all processors run serially by default, and a processor can only do pipelining if it explicitly indicates it's capable of doing so? Sort of like how unit tests run in series unless they call t.Parallel? https://godoc.org/testing#T.Parallel

kr:
LGTM

PR #782, opened 2017-03-17
cmd/cored: set up mux and add handlers later

tessr:
This is a prereq for using raft.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -37,6 +37,7 @@ import (
 	"chain/log/rotation"
 	"chain/log/splunk"
 	"chain/net/http/limit"
+	"chain/net/http/reqid"
 	"chain/protocol"
 	"chain/protocol/bc"
 )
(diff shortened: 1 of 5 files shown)


tessr:
Ran Dashboard against this and made some keys 🔑 

PTAL

kr:
:eyes:

cmd/cored/main.go
```diff
@@ -131,6 +132,61 @@ func runServer() {
...
+		Handler:      secureheader.DefaultConfig,
+		ReadTimeout:  httpReadTimeout,
+		WriteTimeout: httpWriteTimeout,
+		// Disable HTTP/2 for now until the Go implementation is more stable.
+		// https://github.com/golang/go/issues/16450
+		// https://github.com/golang/go/issues/17071
+		TLSNextProto: map[string]func(*http.Server, *tls.Conn, http.Handler){},
+	}
+
+	// The `ListenAndServe` call has to happen in its own goroutine because
```
kr: Nit: personally I would avoid markdown-style backticks in code comments like this, since nothing will ever render them as markdown and the font is already monospace. To me it's more readable without them. But not a big deal.

kr:
LGTM (with one small nit)

cmd/cored/main.go
tessr: I am going to confidently paint this bike shed the other color.

PR #783, opened 2017-03-18
core: fix processor dependencies

jbowens:
Fix a few block processor dependencies that were incorrectly missing or
unnecessary.

expire ACPs processor

The expire account control programs processor used to have a dependency
on the query package's transaction annotator and indexer. However, we no
longer use the account control programs for annotation. We only use the
account_utxo table. Instead, we just need to make sure that the account
indexer has successfully added the utxos to the account_utxo table
before we delete the control programs.

asset processor

A block processor can be run on multiple blocks at once. Previously, if
two consecutive blocks both contained transactions issuing a previously
unknown asset, there was a race condition for which value would be
recorded in the first_block_height column. We can fix this by always
updating with the lower block height.

annotated tx query processor

The query block processor is responsible for marking annotated outputs
as spent on every spend. Since a block processor can be run on multiple
blocks in parallel, the query block processor had a race condition. For
example, an output k can be created in block n-1 and spent in block n.
If the block processor processes block n first and then block n-1, the
processor would create the annotated output and never mark it as spent.
This would permanently corrupt balance queries. If we instead wait for
block n-1 to be processed by the query processor, we ensure that the
output has been created by the time we try to mark it as spent.

I tested this change locally against testnet, creating a few transactions
through the java SDK.

### DIFF
--- a/core/account/indexer.go
+++ b/core/account/indexer.go
@@ -103,7 +103,6 @@ func (m *Manager) ProcessBlocks(ctx context.Context) {
 	}
 	go m.pinStore.ProcessBlocks(ctx, m.chain, ExpirePinName, func(ctx context.Context, b *bc.Block) error {
 		<-m.pinStore.PinWaiter(PinName, b.Height)
-		<-m.pinStore.PinWaiter(query.TxPinName, b.Height)
 		return m.expireControlPrograms(ctx, b)
 	})
 	go m.pinStore.ProcessBlocks(ctx, m.chain, DeleteSpentsPinName, func(ctx context.Context, b *bc.Block) error {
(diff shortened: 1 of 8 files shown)


kr:
LGTM

jbowens:
Should these changes be made to the 1.1-stable branch too? Not sure if it's worthy of a point release all by its lonesome.

kr:
That's a good question. The three standards I have been looking for so far are:

- the bug has no workaround
- the bug is likely to be hit (e.g. >1% probability)
- the bug compromises important functionality (i.e. not cosmetic, not minor)

The one that I'm unsure of here is how likely is it to happen in production? @jbowens do you have a sense of that?

If it meets these standards, we can still throw it on the branch to be in the next point release, whenever that may be, but not actually trigger a release yet.

jbowens:
I think to exercise the bug, these conditions need to be met:
* multiple (_k_) blocks are indexed in parallel
* an output is created and soon after (≤ _k_ blocks) is spent
* the indexer happens to index the later block first

I think it's very unlikely to happen on a single Core network with spare capacity. If the Core is at capacity though, indexing can be slow and multiple blocks can be indexed in parallel.

On a multiple Core network, it's totally reasonable for the multiple blocks to be indexed in parallel because of a temporary network partition, high latency retrieving a block, syncing a new Core, restarting a Core, etc.

The other two conditions seem pretty high probability.

kr:
Ok then I would say yeah, we should at least cherry-pick this onto 1.1-stable.

PR #784, opened 2017-03-18
database/raft: new package

tessr:
This splits out the new `raft` package from #683.


tessr:
PTAL 💃

kr:
:eyes:

database/raft/raft.go
```diff
@@ -0,0 +1,923 @@
...
+	// The actual replicated data set.
+	stateMu   sync.Mutex
+	stateCond sync.Cond
+	state     *state.State
+	done      bool
+
+	// Current log position.
+	// access only from runUpdates goroutine
+	snapIndex uint64
+	confState raftpb.ConfState
```
kr: I'm pretty sure confState needs to be protected by stateMu (or some other synchronization). It is read in getSnapshot (inside of CreateSnapshot), which runs in http handler goroutines.

It looks like snapIndex is okay. As far as I can tell, it is only accessed in the runUpdates goroutine, as the comment says.

database/raft/raft.go
```diff
@@ -0,0 +1,923 @@
...
+	}
+}
+
+func runTicks(rn raft.Node) {
+	for range time.Tick(tickDur) {
+		rn.Tick()
+	}
+}
+
+//
```
kr: empty comment

database/raft/raft.go
```diff
@@ -0,0 +1,923 @@
...
+func (sv *Service) exec(ctx context.Context, instruction []byte) error {
+	prop := proposal{Wctx: randID(), Instruction: instruction}
+	data, err := json.Marshal(prop)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+	req := wctxReq{wctx: prop.Wctx, satisfied: make(chan bool, 1)} //buffered channel
```
kr: nit: we could probably do without this comment, it is just echoing the code.

database/raft/raft.go
```diff
@@ -0,0 +1,923 @@
...
+		return errors.New("raft shutdown")
+	}
+}
+
+// Set sets a value in the key-value storage.
+// If successful, it returns after the value is committed to
+// the raft log.
+// TODO (ameets): possibly RawNode in future to know whether Proposal worked or not
+func (sv *Service) Set(ctx context.Context, key string, val []byte) error {
+	// encode w/ json for now: b with a wctx rand id
```
kr: stale comment

database/raft/raft.go
```diff
@@ -0,0 +1,923 @@
...
+// Delete deletes a value in the key-value storage.
+// if successful, it returns after the value is deleted from the raft log.
+// TODO (ameets): is RawNode possible/applicable?
+func (sv *Service) Delete(ctx context.Context, key string) error {
+	b := state.Delete(key)
+	return sv.exec(ctx, b)
+}
+
+func (sv *Service) allocNodeID(ctx context.Context) (uint64, error) {
+	// encode w/ json for now: b with a wctx rand id
```
kr: stale comment

database/raft/raft.go
```diff
@@ -0,0 +1,923 @@
...
+			log.Printkv(context.Background(), "no-addr-for-peer", msg.To)
+			continue
+		}
+		sendmsg(addr, data)
+	}
+}
+
+// best effort. if it fails, oh well -- that's why we're using raft.
+func sendmsg(addr string, data []byte) {
+	resp, err := http.Post("http://"+addr+"/raft/msg", contentType, bytes.NewReader(data))
```
kr: This is gonna need to support TLS. See d49c4c2d7b0133d9f94e337c5bcb298ce89070e5. This will be simpler once @boymanjor's TLS-everywhere change goes in, then it can be hard-coded to https.

database/raft/state/state.go
```diff
@@ -0,0 +1,223 @@
...
+// It should be called with the retrieved snapshot
+// when bootstrapping a new node from an existing cluster
+// or when recovering from a file on disk.
+func (s *State) RestoreSnapshot(data []byte, index uint64) error {
+	s.appliedIndex = index
+	//TODO (ameets): think about having statepb in state for restore
+	snapshot := &statepb.Snapshot{}
+	err := proto.Unmarshal(data, snapshot)
+	s.peers = snapshot.Peers
+	s.state = snapshot.State //TODO (ameets): need to add version here
```
tessr: Yes. My next three follow up issues are:

1. Evicting nodes from the cluster when they go offline
2. Protecting against incompatible versions
3. Forcing corectl to use the API.

database/raft/raft.go
```diff
@@ -0,0 +1,923 @@
...
+	if err != nil {
+		log.Printkv(context.Background(), log.KeyWarning, err)
```
tessr: Hmm alright. I had been thinking that "warning" was sufficiently standard that it merited its own constant, but you're right, we don't use it anywhere else.

tessr:
This is ready for another pass, PTAL

database/raft/raft.go
```diff
@@ -0,0 +1,935 @@
...
+
+	// The actual replicated data set.
+	stateMu   sync.Mutex
+	stateCond sync.Cond
+	state     *state.State
+	done      bool
+
+	// Current log position, accessed only from runUpdates goroutine
+	snapIndex uint64
+	confState raftpb.ConfState
```
kr: Now that this is protected by stateMu this field should move up to that group just above.

database/raft/raft.go
```diff
@@ -0,0 +1,935 @@
...
+	}
+
+	if raft.IsEmptySnap(raftSnap) {
+		log.Printkv(ctx, "at", "joined", "appliedindex", 0, "msg", "(empty snap)")
+	} else {
+		err = sv.state.RestoreSnapshot(raftSnap.Data, raftSnap.Metadata.Index)
+		if err != nil {
+			return errors.Wrap(err)
+		}
+		sv.stateMu.Lock()
```
kr: Acquiring here isn't necessary because the http listener hasn't started yet, so this "happens before" any of the handlers can start. This is also why we don't hold the lock when calling RestoreSnapshot just above. Basically, we have a "setup" phase and then an "online" phase, and we don't take the lock during the setup phase because there's no concurrency in play.

I am ok with taking the lock on this line, but if we do it, then we should also do it consistently throughout the setup phase whenever we're modifying state fields.

tessr: ah, got it. I'll drop the lock.

tessr:
🔴 -y for another pass

kr:
LGTM

PR #785, opened 2017-03-20
docs/internal: document point release standards

kr:
This change attempts to codify the standards we use to
include changes in a point release. It'll help to have
them written down. We can refine them over time if we
think it's necessary.


kr:
PTAL cc @jeffomatic 

This has come up a couple of times (e.g. https://github.com/chain/chain/pull/783#issuecomment-287830729) and I keep having to dig around in the slack history to remember what they even are.

PR #786, opened 2017-03-20
core: fix processor dependencies

jbowens:
Fix a few block processor dependencies that were incorrectly missing or
unnecessary. This is the 1.1-stable version of #783. Because main and
`1.1-stable` have diverged some, this PR is slightly different. Most
notably, because account_utxos are marked as spent in the account block
processor, the account block processor must also wait for block n-1 to
be indexed.

I had to move the account package's PinWaiter calls into closures for the
tests.


kr:
LGTM

PR #787, opened 2017-03-20
database/pg: remove Uint32s

jbowens:
Use pq.Array([]uint32) instead of maintaining our own uint32-specific
PostgreSQL array serializer/deserializer.

Tested this change locally by connecting a new Core to testnet and
running the five-minute guide.


bobg:
LGTM

PR #788, opened 2017-03-20
protocol/validation: implement entries-based validation

bobg:
Rewrite block and transaction validation in terms of "entries."

This implements the spec changes that (at the moment) live on the `txgraph-spec` branch. Reviewers, please evaluate this PR against the documents there.

What used to be called the "well-formedness" check, a.k.a. "context-free validation," is now simply "validation" and lives in `protocol/validation`. Validation that depends on blockchain state is now called "applying" and lives in `protocol`.

The old tangle of different entrypoints to (different phases of) validation has hopefully been simplified.

The blockchain state's "issuance memory" is replaced by a "nonce set."


bobg:
rebased

bobg:
PTAL

protocol/bc/validation.go
```diff
@@ -0,0 +1,142 @@
...
+		blockchainID: initialBlockID,
+	}
+	return tx.checkValid(vs)
+}
+
+// ValidateBlock validates a block and the transactions within.
+// The consensus program is executed only if prev is non-nil and
+// runProg is true.
+func ValidateBlock(b, prev *BlockEntries, initialBlockID Hash, runProg bool) error {
+	if prev == nil {
```
kr: Looks like if the caller provides a prev block on block 1, this function will attempt to validate against it. I think it would be better to ignore it (or maybe return an error) in that case.

I also think it'd be easier to notice that sort of thing — easier to read the code — if it were flatter:

```go
if b.Body.Height > 1 && prev == nil {
	return errors.WithDetailf(errNoPrevBlock, "height %d", b.Body.Height)
}
if b.Body.Height > 1 {
	// validate against prev block
}

// validate current block
```

(It's okay to repeat part of the test in another if statement. We often have an urge to factor those out but it is just as often more readable if we don't. Check out these two excerpts from [TEOPS](https://www.amazon.com/Elements-Programming-Style-2nd/dp/0070342075).)

decisions1

and

decisions2

protocol/bc/validation.go
```diff
@@ -0,0 +1,142 @@
...
+		if b.Body.Height != 1 {
+			return errors.WithDetailf(errNoPrevBlock, "height %d", b.Body.Height)
+		}
+	} else {
+		if b.Body.Version < prev.Body.Version {
+			return errors.WithDetailf(errVersionRegression, "previous block verson %d, current block version %d", prev.Body.Version, b.Body.Version)
```
kr: For a long time, I have wished that the consensus-critical validation code didn't have to be cluttered up with all this error reporting. The error descriptions are a lot more than the actual logic by sheer volume.

What do you think of having two separate functions: `IsValidBlock(...) bool` and `AnalyzeInvalidBlock(...) error`. AnalyzeInvalidBlock would be handy to report back meaningful error messages to clients, but it'd never be called in the actual validator. It would have to duplicate most or all of the validation logic, but in return, IsValidBlock, which is much more sensitive to bugs, would become smaller and easier to read.

This is all pretty offtopic for this PR. I'm just thinking about it again because we're looking so closely at the validation code now.

protocol/bc/value.go
```diff
@@ -13,6 +15,53 @@ type ValueSource struct {
 	Entry `entry:"-"`
 }
 
+func (vs *ValueSource) CheckValid(vstate *validationState) error {
```
kr: Does this method need to be exported?

protocol/vm/ops.go
```diff
@@ -201,19 +201,20 @@ const (
...
-	OP_NONCE         Op = 0xcc
-	OP_NEXTPROGRAM   Op = 0xcd
-	OP_BLOCKTIME     Op = 0xce
+	OP_CHECKOUTPUT Op = 0xc1
+	OP_ASSET       Op = 0xc2
+	OP_AMOUNT      Op = 0xc3
+	OP_PROGRAM     Op = 0xc4
+	OP_MINTIME     Op = 0xc5
+	OP_MAXTIME     Op = 0xc6
+	OP_TXDATAHASH  Op = 0xc7
```
bobg: As far as I'm concerned, the naming of these opcodes (and the data fields they refer to) is not yet a settled question. (See https://github.com/chain/chain/pull/728#discussion_r105473953.) Although populating them with hashes is likely to be a common use case, nothing requires them to be hashes. I'd sooner keep the "ref" in the name and lose the "hash."

In any case, I want the code to reflect the names in the spec. No sense letting them diverge for the sake of saving a few diff lines in one PR.

protocol/state/snapshot.go
```diff
@@ -1,53 +1,110 @@
...
-// PruneIssuances modifies a Snapshot, removing all issuance hashes
+func (s *Snapshot) DeleteSpentOutput(id bc.Hash) error {
+	if !s.Tree.Contains(id[:]) {
+		return fmt.Errorf("invalid prevout %x", id[:])
+	}
+	s.Tree.Delete(id[:])
+	return nil
+}
+
+func (s *Snapshot) AddOutput(id bc.Hash) error {
```
bobg: Done. This (and two other functions, also now inlined) were there to satisfy a short-lived interface called `BlockchainState` used by `protocol/bc` when `Apply` functionality briefly lived there.

protocol/state/snapshot.go
```diff
@@ -1,53 +1,110 @@
...
 	}
 }
 
 // Copy makes a copy of provided snapshot. Copying a snapshot is an
-// O(n) operation where n is the number of issuance hashes in the
-// snapshot's issuance memory.
-func Copy(original *Snapshot) *Snapshot {
+// O(n) operation where n is the number of nonces in the snapshot's
+// nonce set.
+func (s *Snapshot) Copy() *Snapshot {
```
bobg: Reverted. An artifact of how this PR evolved (and of other changes that were already reverted).

protocol/cache.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/golang/groupcache/lru"
+
+	"chain/protocol/bc"
+)
+
+// maxCachedValidatedTxs is the max number of validated txs to cache.
+const maxCachedValidatedTxs = 1000
+
+type prevalidatedTxsCache struct {
+	mu  sync.Mutex
```
bobg: Not a change: https://github.com/chain/chain/blob/main/protocol/tx.go#L28

core/txdb/internal/storage/snapshot.proto
```diff
@@ -10,14 +10,23 @@ message Snapshot {
   repeated StateTreeNode nodes = 1;
 
   // Issuances contains the record of recent issuances for ensuring uniqueness
-  // of issuances.
-  repeated Issuance issuances = 2;
+  // of issuances. DEPRECATED in favor of nonces (below).
+  repeated Issuance issuances = 2 [deprecated=true];
```
bobg: Done. I knew that renaming a field would not affect the wire format, but I was less sure about renaming a message type. A little research makes me reasonably assured that's safe too.

protocol/bc/nonce.go
```diff
@@ -42,3 +47,27 @@ func (n *Nonce) SetAnchored(id Hash, entry Entry) {
 	n.Witness.AnchoredID = id
 	n.Anchored = entry
 }
+
+func (n *Nonce) checkValid(vs *validationState) error {
```
bobg: I'm considering your suggestion. I'm not opposed, but I do like the fact that the current pattern places each `checkValid` close to the definition of the data structure it's checking. Of course that will change when each entry type gets defined in a .proto file instead of a .go file.

> composing them (package txbuilder)

Note that when we retire the legacy data structures and therefore no longer have `MapTx`, we will need a new `protocol/bc`-level mechanism for composing txentry graphs. It will probably have a builder-like interface, which will blur the distinction between `protocol/bc` and `core/txbuilder`.

protocol/bc/nonce.go
```diff
@@ -42,3 +47,27 @@ func (n *Nonce) SetAnchored(id Hash, entry Entry) {
...
+	err = n.TimeRange.checkValid(&vs2)
+	if err != nil {
+		return errors.Wrap(err, "checking nonce timerange")
+	}
+
+	if n.TimeRange.Body.MinTimeMS == 0 || n.TimeRange.Body.MaxTimeMS == 0 {
+		return errZeroTime
+	}
+
+	if vs.tx.Body.Version == 1 && (n.Body.ExtHash != Hash{}) {
```
bobg: Looks weird to me, but OK.

protocol/bc/txissuance.go
```diff
@@ -53,3 +58,58 @@ func NewIssuance(anchor Entry, value AssetAmount, data Hash, ordinal int) *Issua
...
+		return errors.WithDetailf(errMismatchedAssetID, "asset ID is %x, issuance wants %x", computedAssetID[:], iss.Body.Value.AssetID[:])
+	}
+
+	err := vm.Verify(NewTxVMContext(vs.tx, iss, iss.Witness.AssetDefinition.IssuanceProgram, iss.Witness.Arguments))
+	if err != nil {
+		return errors.Wrap(err, "checking issuance program")
+	}
+
+	var anchored Hash
+	switch a := iss.Anchor.(type) {
```
bobg: I like this suggestion from a Go perspective, but on the other hand the spec is explicit about the set of types allowed as anchors and I can't decide whether the Go-cleanness of your suggestion is worth hiding the explicitness in the spec.

protocol/bc/value.go
bobg: Ha! No. I originally had all the `CheckValid` functions exported, and at some point I unexported them with:

`gorename -from '"chain/protocol/bc".ValidChecker.CheckValid' -to checkValid`

(I then also unexported `validChecker`.)

As I now understand, this worked on `Entry` types like `Mux` and `Spend` because the definition of interface `Entry` explicitly includes `validChecker`, but it skipped `ValueSource` and `ValueDestination` because nothing says that those types have to be `validChecker`s. Interesting!

protocol/bc/validation.go
bobg: Interesting thought. It might be possible to write the validation logic just once, in terms of (admittedly complex) interfaces that can be satisfied by a simple validation-checker on the one hand, and a validation-error-analyzer on the other. (Ask me to show you how I used this approach in my Trustless Digital Escrow demo.)

Agree about pushing this off to a later PR.

protocol/bc/validation.go
bobg: Your point is well-taken, but in this particular case I thought a slightly different approach to clarifiying to code was called for, LMKWYT.

protocol/bc/validation.go
```diff
@@ -0,0 +1,142 @@
...
+	vs := &validationState{
+		blockchainID: initialBlockID,
+	}
+	return tx.checkValid(vs)
+}
+
+// ValidateBlock validates a block and the transactions within.
+// The consensus program is executed only if prev is non-nil and
+// runProg is true.
+func ValidateBlock(b, prev *BlockEntries, initialBlockID Hash, runProg bool) error {
```
bobg: Done in https://github.com/chain/chain/pull/788/commits/6df0569eac2d9feba85fa81d8879b51ae1e381f5 but it doesn't feel like an improvement, feels about the same. I think I'd be slightly happier reverting that change, LMKWYT.

protocol/tx.go
```diff
@@ -1,61 +1,40 @@
...
+// performing full validation.
+func (c *Chain) ValidateTx(tx *bc.TxEntries) error {
+	err := c.checkIssuanceWindow(tx)
+	if err != nil {
+		return err
+	}
+	var ok bool
+	err, ok = c.prevalidated.lookup(tx.ID)
+	if !ok {
+		err = bc.ValidateTx(tx, c.InitialBlockHash)
```
bobg: Not without pain.

As you know, this PR started out much bigger, but I factored out a series of smaller lead-up PR's, each a logical stepping stone in getting to this one. Changing the validation API, however, did not happen in isolation from switching to entries. Instead, I wrote a whole new validate/apply API, trying simultaneously to make it line up with existing callers' expectations and to clarify (compared to what we had before) the boundaries and contracts of the different entrypoints.

In other words, restoring the previous validation API would be brand-new work, unlike all the lead-up PRs. It's not impossible, and I'll do it if you really think it's needed, but since the intent would be quickly to replace it with what's already in this PR, my strong preference is not to spend the effort.

bobg:
Updated (and a couple of open comments added), PTAL

protocol/bc/validation.go
kr: Ah, I see what you mean, but that wasn't exactly what I had in mind. However, I didn't notice it would require a bit more plumbing since the validator/signer doesn't call this function directly. I'm fine reverting 6df0569eac2d9feba85fa81d8879b51ae1e381f5 and maybe I can take a stab at my idea in a followup PR to see if it looks reasonable.

protocol/bc/blockheader.go
```diff
@@ -36,3 +36,12 @@ func NewBlockHeaderEntry(version, height uint64, previousBlockID Hash, timestamp
 	bh.Body.NextConsensusProgram = nextConsensusProgram
 	return bh
 }
+
+// checkValid does only part of the work of validating a block. The
+// rest is handled in ValidateBlock, which calls this.
+func (bh *BlockHeaderEntry) checkValid(_ *validationState) error {
```
kr: if you want, it's okay to leave out the parameter name entirely when all of the params are unused, e.g.

```go
func (bh *BlockHeaderEntry) checkValid(*validationState)
```

protocol/bc/nonce.go
bobg: Here's what your suggestion looks like: https://github.com/chain/chain/pull/788/commits/425e060fa45136f6ee032630bbd9f95fadb24ffe. Keep it? Revert it? Change it?

protocol/cache.go
kr: D'oh! My bad. It is a little unfortunate that this code shows up as changed in the diff.

protocol/bc/nonce.go
kr: Keep it! (We both agreed offline that we prefer it.)

protocol/cache.go
bobg: Agreed; fixed.

kr:
LGTM. Do you wanna wait for another reviewer?

kr:
Oh maybe the title should say `protocol/validation`

bobg:
> wait for another reviewer

Great idea.

bobg:
Ran it once. :tm: Did the five-minute tutorial and some extra stuff, trying and failing to fool validation.

core/txbuilder/finalize.go
```diff
@@ -40,8 +39,8 @@ func FinalizeTx(ctx context.Context, c *protocol.Chain, s Submitter, tx *bc.Tx)
 	<-c.BlockWaiter(1)
 
 	// If this transaction is valid, ValidateTxCached will store it in the cache.
```
jbowens: This comment looks like it needs to be updated?

core/txdb/internal/storage/snapshot.proto
```diff
@@ -9,11 +9,11 @@ message Snapshot {
   // The nodes are ordered according to a pre-order traversal.
   repeated StateTreeNode nodes = 1;
 
-  // Issuances contains the record of recent issuances for ensuring uniqueness
-  // of issuances.
-  repeated Issuance issuances = 2;
```
jbowens: Should this Issuance -> Nonce rename happen in a separate PR?

protocol/tx.go
```diff
@@ -10,17 +10,23 @@ import (
...
-	// Consult a cache of prevalidated transactions.
-	err, ok := c.prevalidated.lookup(tx.ID)
-	if ok {
+// ErrBadTx is returned for transactions failing validation
+var ErrBadTx = errors.New("invalid transaction")
+
+// ValidateTx validates the given transaction. A cache holds
+// per-transaction validation results and is consulted before
+// performing full validation.
+func (c *Chain) ValidateTx(tx *bc.TxEntries) error {
```
jbowens: Any particular reason for this rename? If the rename is orthogonal to the actual changes of the PR, I'd love it to be in its own PR.

protocol/block.go
```diff
@@ -63,58 +69,86 @@ func (c *Chain) GenerateBlock(ctx context.Context, prev *bc.Block, snapshot *sta
...
+// CommitBlock commits a block to the blockchain. The block
+// must already have been applied with ApplyValidBlock or
+// ApplyNewBlock, which will have produced the new snapshot that's
+// required here.
 //
 // This function:
 //   * saves the block to the store.
-//   * saves the state tree to the store (optionally).
+//   * sometimes saves the state tree to the store.
 //   * executes all new-block callbacks.
```
jbowens: While you're updating these comments, "executes all new-block callbacks" is no longer true. That happens asynchronously through the block-processor pins.

protocol/block.go
```diff
@@ -63,58 +69,86 @@ func (c *Chain) GenerateBlock(ctx context.Context, prev *bc.Block, snapshot *sta
...
+}
+
+// ValidateBlock validates an incoming block in advance of applying it
+// to a snapshot (with ApplyValidBlock) and committing it to the
+// blockchain (with CommitAppliedBlock).
+func (c *Chain) ValidateBlock(block, prev *bc.Block) error {
+	return c.validateBlock(block, prev, true)
+}
+
+func (c *Chain) validateBlock(block, prev *bc.Block, runProg bool) error {
```
jbowens: Do `ValidateBlock` and `validateBlock` only need to be methods of `Chain` in order to access `c.InitialBlockHash`? Maybe make that a parameter instead then?

protocol/state/snapshot.go
```diff
@@ -1,53 +1,96 @@
...
 	return &Snapshot{
-		Tree:      new(patricia.Tree),
-		Issuances: make(map[bc.Hash]uint64),
+		Tree:   new(patricia.Tree),
+		Nonces: make(map[bc.Hash]uint64),
+	}
+}
+
+// ApplyBlock updates s in place.
+func (s *Snapshot) ApplyBlock(block *bc.BlockEntries) error {
```
jbowens: Should `ApplyBlock` and `ApplyTx` be exported functions? As per Keith's TODO comment up above in `Copy`, we've been slowly moving towards making `Snapshot` an immutable type. Keeping in-place modifications out of package interfaces was philosophically aligned with Snapshot as immutable.

protocol/validation/validation.go
```diff
@@ -0,0 +1,466 @@
...
+		return errors.Wrapf(errMismatchedValue, "destination value %v disagrees with %v", src.Value, vd.Value)
+	}
+
+	return nil
+}
+
+// ValidateBlock validates a block and the transactions within.
+// The consensus program is executed only if prev is non-nil and
+// runProg is true.
+func ValidateBlock(b, prev *bc.BlockEntries, initialBlockID bc.Hash, runProg bool) error {
```
jbowens: Instead of using a boolean `runProg` flag, should we split into multiple functions? That way when you see a function's invocation, you can tell by it's name what it's doing. It's much harder to interpret a trailing `true` or `false`.

jbowens: It looks like the `runProg` flag is only used for a single if statement in `validateBlockAgainstPrev`?:

```
	if runProg {
		vmContext := NewBlockVMContext(b, prev.Body.NextConsensusProgram, b.Witness.Arguments)
		err := vm.Verify(vmContext)
		if err != nil {
			return errors.Wrap(err, "evaluating previous block's next consensus program")
		}
	}
```

Another option is to have an argument that can speak for itself. For ex, make the body of that if statement a function:

```
func ValidateConsensusProgram(b, prev *bc.BlockEntries, runProg bool) error { ... }
```

And then change `ValidateBlock`'s signature to take that as an argument? So then an invocation would look like:

```
ValidateBlock(b, prev, initialBlockID, validation.ValidateConsensusProgram)
```

IDK, just splitballing.

core/txbuilder/finalize.go
bobg: ...or just removed. Done.

core/txdb/internal/storage/snapshot.proto
bobg: I don't think so, because nonce entries, and replacement of the issuance memory with the nonce set, are an integral part of the same spec changes implemented by the rest of this PR.

protocol/tx.go
bobg: It's not so much a rename as it is part of a brand-new, simpler interface to validation that happens to look a lot like the old one. See the discussion at https://github.com/chain/chain/pull/788#discussion_r108548333

`protocol.Chain` has no tx-validation entrypoint that isn't cached, so there's no need to include `Cached` in the name.

protocol/block.go
bobg: It is a parameter in the `protocol/validation` layer, but only the `protocol` package interfaces with that now; everything else does validation through a `Chain` object.

protocol/state/snapshot.go
bobg: Given the current (and IMO gratifyingly clear) structure of the code, I think they should. Mutating Snapshots is no more than what the code was already doing. But I expanded on the TODO comment to point out that the `Apply` functions (and `PrunceNonces`) will have to change if Snapshots become immutable.

protocol/validation/validation.go
bobg: We tried something similar earlier in this code review and decided to revert and punt for now. See https://github.com/chain/chain/pull/788#discussion_r108581917.

protocol/block.go
jbowens: It could still be a `protocol.ValidateBlock` just not with a `Chain` receiver. Generally, imo, unnecessarily making functions methods complicates testing.

core/txdb/internal/storage/snapshot.proto
jbowens: Even if it's not philosophically orthogonal, it's definitely architecturally and practically orthogonal. I don't think there should be a single atomic PR implementing the spec changes.

bobg: In this case I don't agree, but I did it anyway: https://github.com/chain/chain/pull/788/commits/af4e95f8df9c1324b23f97569ca5be9788da70ed

protocol/block.go
bobg: Do you mean changing from taking a `Chain` as a receiver to taking it as a parameter? That seems like a pretty cosmetic change, not sure why you'd insist on that.

Or do you mean requiring callers to pull the initial block hash out of the `Chain` object first and then calling `protocol.ValidateBlock(block, prev, initialBlockHash)`? Why add that burden to the callers, which didn't previously have to know or care that an initial block hash can be found in the `Chain` object or is needed by validation? The cost of expanding the API surface in that way exceeds any small complication in testing, it seems to me.

bobg: Oh geez, hang on, you're talking about `validateBlock`, not `ValidateBlock`. I'm totally with you on that. Stand by.

bobg: Hang on again, looks like you _are_ talking about `ValidateBlock`, though this comment thread is on the `validateBlock` line. Well I split the difference: `validateBlock` is now a function, but `ValidateBlock` is still a method. I'll entertain changing that to a function that takes a `Chain` object if you really insist, but I strongly resist making that entrypoint require a separate initial block hash argument.

jbowens: I won't insist, but it's not requiring an additional parameter. Receivers are parameters themselves. It's really changing a parameter from a large composite type to the value you need.

It's analogous to taking a `http.Request` struct as a parameter when you really only need it's `Method` field. By requiring the smallest dependency possible, you make the function easier to use.

Practically speaking, it's not a big deal and is up to you. Just want to make sure the point isn't lost.

bobg: Haha, writing a response to this (that mentioned, among other things, that `Chain.ValidateTx` needs more from its `Chain` object than just the initial block hash, and it's a slippery slope as to when callers should just pass the composite object vs. pulling out fields and passing those) led me to realize that I had inadvertently clobbered the logic that used the tx validation cache during block validation. That's now restored, PTA(nother)L.

As to your point above, I do get it, and I agree, with the qualifications already discussed. In this instance, I think the de-methodization of `protocol.validateBlock` and the existence of `validation.ValidateBlock` (which can be used to bypass the `Chain` object entirely if needed) address your concerns about usability and testing complexity.

bobg: Incidentally, `validation.ValidateBlock` is only called from one place now, and I wonder whether it's worth adding that `validateTx` parameter to it, as I've just done, vs. inlining the whole thing in `protocol.validateBlock`. Opinions?

kr: I think @jbowens is right on about the general point. It's not necessarily accurate to characterize the alternative as "pulling out fields and passing those" since the caller might not have the composite object to begin with. It might be more work to gin up or plumb in that composite object vs just using the one or two relevant values.

But in this case it seems okay, since Chain.ValidateBlock is just a convenience wrapper around validation.ValidateBlock. There's potentially a valid debate around whether we ought to have that convenience wrapper at all… a more interesting question than how many places do we call validation.ValidateBlock is how many places (in non-test code) do we call Chain.ValidateBlock? If it's just one or two then I'd advocate for inlining the convenience wrapper.

bobg: Is this what you had in mind? https://github.com/chain/chain/pull/788/commits/3eb576e00bdbb13fc125c7b938794613a04c58c3

Because to me that looks like abstraction-boundary leakage that makes the code harder to read and understand.

Granted, it will get a little better soon when we don't have to call `bc.MapBlock` (because `bc.Block` will be replaced with `bc.BlockEntries` everywhere).

kr: Yeah something like that. I think it can be made a little cleaner by fiddling with the runProg parameter as @jbowens and I have suggested. Honestly I'm happy with this PR either way. It will be easier for us to discuss and evaluate these alternatives separately.

PR #789, opened 2017-03-20
core: unexport API.HealthSetter

jbowens:
Unexport API.HealthSetter. Now that the leader run function is in the
core package, HealthSetter doesn't need to be exported.


bobg:
LGTM

PR #790, opened 2017-03-20
protocol/bc: create AssetDefinition type, ComputeAssetID is its hash

bobg:
The spec revision in https://github.com/chain/chain/pull/728 specifies a new Asset Definition type. This change reflects that.

This is a carved-off piece of https://github.com/chain/chain/pull/788

### DIFF
--- a/cmd/multitool/multitool.go
+++ b/cmd/multitool/multitool.go
@@ -182,7 +182,10 @@ func assetid(args []string) {
 	if len(assetdefInp) > 0 {
 		assetdefHash = mustDecodeHash(assetdefInp)
 	}
-	assetID := bc.ComputeAssetID(issuance, initialBlock, 1, assetdefHash)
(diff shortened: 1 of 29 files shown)


bobg:
PTAL

jbowens:
Most of this PR seems to be propagating errors from `writeForHash`. Do we ever use `writeForHash` with a writer that's not a buffer? Besides `io.Writer` errors, it looks like the only other errors are reflection type errors.

Can we modify the `writeForHash` or `ComputeAssetID` interface to panic instead?

bobg:
> Can we modify the `writeForHash` or `ComputeAssetID` interface to panic instead?

We can do better, I reckon: we can tighten the contract of `writeForHash` to make errors impossible in the `ComputeAssetID` case. (Done.)

jbowens:
lgtm

PR #791, opened 2017-03-21
core: remove deprecated /rpc/get-blocks endpoint

jbowens:
The /rpc/get-blocks RPC endpoint was replaced with the /rpc/get-block
endpoint shortly before the 1.0.0 release.

See 0a590e07e9484872d2ee7a5b28d5b8b3a582764f.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -147,7 +147,6 @@ func (a *API) buildHandler() {
 	m.Handle(networkRPCPrefix+"submit", needConfig(func(ctx context.Context, tx *bc.Tx) error {
 		return a.submitter.Submit(ctx, tx)
 	}))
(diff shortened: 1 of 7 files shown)


jbowens:
> Does this PR need a special tag or does it need to update the release notes? Removing deprecated functionality is definitely something we need to communicate when 1.2 goes out.

Maybe, but in this case it's a _network_ RPC endpoint, so the only consumer should be other Cores. Also, we've never had an official release that used this RPC endpoint. It was deprecated _before_ 1.0.0.

kr:
> Also, we've never had an official release that used this RPC endpoint.

Cool, then as a matter of practicality perhaps we should not bump the network protocol version, but just describe this when we write the release notes for 1.2 but not make a big stink about it.

@jeffomatic thoughts?

jeffomatic:
SGTM

PR #792, opened 2017-03-21
core: add unit tests for waitForTxInBlock

jbowens:
Add unit tests for `waitForTxInBlock` to verify that the correct
block height is returned when a tx is confirmed and that the tx is
re-submitted for every block that does not contain the unconfirmed tx.

### DIFF
--- a/core/transact_test.go
+++ b/core/transact_test.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


core/transact_test.go
```diff
@@ -97,3 +98,85 @@ func TestRecordSubmittedTxs(t *testing.T) {
...
+	// Make n blocks but never include the transaction
+	// that we're looking for. The tx should be resubmitted
+	// to the generator each time.
+	for i := 0; i < timesToResubmit; i++ {
+		prottest.MakeBlock(t, c, []*bc.Tx{})
+	}
+
+	// Wait until the submitter records that the transaction has been
+	// re-submitted to the generator all n times.
+	wg.Wait()
```
bobg: So this will either succeed, or wait forever? Seems weird to have a test with no calls to `t.Error` et al.

jbowens: Agreed, it is weird, but I think it's what you need to do to test asynchronous things like this. See [this sync.WaitGroup stdlib test](https://github.com/golang/go/blob/b3bf38e79d71784d98a79f2c2e3af003f48c8052/src/sync/waitgroup_test.go#L14-L38).

I updated this individual test to add an explicit 2-second timeout. I worry about it making the test more fragile than relying on the `go test` default timeout.

bobg:
LGTM

PR #793, opened 2017-03-21
protocol/bc: move CalcMerkleRoot here

bobg:
This is part of "entries"-based validation, which moves much of the validation logic into the `protocol/bc` package.

### DIFF
--- a/protocol/bc/merkle.go
+++ b/protocol/bc/merkle.go
@@ -1,10 +1,9 @@
-package validation
+package bc
 
 import (
 	"math"
 
 	"chain/crypto/sha3pool"
-	"chain/protocol/bc"
 )
 
 var (
@@ -14,7 +13,7 @@ var (
 
 // CalcMerkleRoot creates a merkle tree from a slice of transactions
 // and returns the root hash of the tree.
-func CalcMerkleRoot(transactions []*bc.Tx) (root bc.Hash, err error) {
+func CalcMerkleRoot(transactions []*Tx) (root Hash, err error) {
 	switch {
 	case len(transactions) == 0:
 		sha3pool.Sum256(root[:], nil)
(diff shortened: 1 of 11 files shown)


bobg:
PTAL

kr:
Ideally the description for each commit would stand on its own. What is going on in this patch, why we are doing it, etc. Perhaps the motivation is just "this will make it easier to blah blah blah in the future", but we should spell out what "blah blah blah" is. Saying "Another carved-off piece of #788" is an interesting and useful fact to mention probably at the end, after the actual explanation.

bobg:
PTAL

bobg:
PTAL

protocol/bc/merkle.go
```diff
@@ -14,7 +13,7 @@ var (
 
 // CalcMerkleRoot creates a merkle tree from a slice of transactions
 // and returns the root hash of the tree.
-func CalcMerkleRoot(transactions []*bc.Tx) (root bc.Hash, err error) {
+func CalcMerkleRoot(transactions []*Tx) (root Hash, err error) {
```
kr: Would you be ok with renaming this to MerkleRoot? I'd prefer to name a function after what it returns, rather than what it does.

protocol/bc/merkle_test.go
```diff
@@ -1,18 +1,18 @@
-package validation
+package bc_test
```
kr: It looks like the uses of package vm here are just for some constants that have no relevance to the meaning of the tests. Seems like it would be pretty trivial to remove that dependency. In fact I just opened#799 to do that. It would be preferable imo to land #799 first, then do this pr without having to resort to the bc_test trick.

protocol/bc/merkle.go
```diff
@@ -14,7 +13,7 @@ var (
 
 // CalcMerkleRoot creates a merkle tree from a slice of transactions
```
kr: s/CalcMerkleRoot/MerkleRoot/

bobg:
rebased

kr:
LGTM with a nit

PR #794, opened 2017-03-21
protocol/prottest/memstore: move MemStore here

bobg:
`MemStore` is just a test utility, so it belongs under `prottest`.

### DIFF
--- a/protocol/block_test.go
+++ b/protocol/block_test.go
@@ -63,7 +63,7 @@ func TestNoTimeTravel(t *testing.T) {
 }
 
 func TestWaitForBlockSoonAlreadyExists(t *testing.T) {
-	c := prottest.NewChain(t)
+	c, _ := newTestChain(t, time.Now())
 	makeEmptyBlock(t, c) // height=2
 	makeEmptyBlock(t, c) // height=3
 
@@ -74,7 +74,7 @@ func TestWaitForBlockSoonAlreadyExists(t *testing.T) {
 }
 
 func TestWaitForBlockSoonDistantFuture(t *testing.T) {
-	c := prottest.NewChain(t)
+	c, _ := newTestChain(t, time.Now())
 
 	got := <-c.BlockSoonWaiter(context.Background(), 100) // distant future
 	want := ErrTheDistantFuture
@@ -92,7 +92,7 @@ func TestWaitForBlockSoonWaits(t *testing.T) {
 	//
 	// It's the best we can do.
 
-	c := prottest.NewChain(t)
+	c, _ := newTestChain(t, time.Now())
 	makeEmptyBlock(t, c) // height=2
 
 	go func() {
@@ -110,7 +110,7 @@ func TestWaitForBlockSoonWaits(t *testing.T) {
 }
 
 func TestWaitForBlockSoonTimesout(t *testing.T) {
-	c := prottest.NewChain(t)
+	c, _ := newTestChain(t, time.Now())
 	go func() {
 		makeEmptyBlock(t, c) // height=2
 	}()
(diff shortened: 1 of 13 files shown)


bobg:
PTAL

protocol/export_test.go
```diff
@@ -0,0 +1,9 @@
+package protocol
```
jbowens: Is there a way we can avoid moving to tests to a separate package?

protocol/export_test.go
```diff
@@ -0,0 +1,9 @@
...
+
+func SetChainHeight(c *Chain, height uint64) {
+	c.setHeight(height)
+}
+
+func GetChainHeight(c *Chain) uint64 {
+	return c.state.height
+}
```
jbowens: Can we use c.Height instead? Or do we need to explicitly not lock the protecting mutex?

protocol/prottest/memstore.go
```diff
@@ -1,8 +1,8 @@
-// Package memstore provides a Store implementation that
+// MemStore is a Store implementation that
 // keeps all blockchain state in memory.
 //
 // It is used in tests to avoid needing a database.
```
jbowens: Should get rid of this package comment if we're moving it into prottest. It no longer describes the package but the type.

protocol/tx_test.go
```diff
@@ -1,4 +1,4 @@
-package protocol
```
jbowens: Is this move to `protocol_test` necessary? If we really need to move it to a test package, then we should stop using `newTestChain` and use `prottest.NewChain`. `newTestChain` exists only to circumvent with the dependency cycle.

protocol/block_test.go
```diff
@@ -223,7 +224,7 @@ func newTestChain(tb testing.TB, ts time.Time) (c *Chain, b1 *bc.Block) {
 	if err != nil {
 		testutil.FatalErr(tb, err)
 	}
-	c, err = NewChain(ctx, b1.Hash(), memstore.New(), nil)
+	c, err = NewChain(ctx, b1.Hash(), prottest.NewMemStore(), nil)
```
jbowens: this newTestChain func is moot if we're moving it into protocol_test. We can use `prottest.NewChain` and friends.

protocol/tx_test.go
jbowens: IMO, the separate memstore package is cleaner than using `_test` packages. What about moving the memstore package to `prottest/memstore`?

protocol/export_test.go
bobg: In this PR it wouldn't be too hard, but this is preparation for other moving around that would make it much harder. (In short, `MemStore` will start depending directly on the `protocol` package, and [since it definitely wants to live in `prottest`] `protocol` can't also depend on `prottest`.)

bobg:
Updated, PTAL

protocol/tx_test.go
bobg: As I said, that'd be OK in this PR, but we'll only have to undo it in my next PR, when the memstore code starts depending on `protocol` (because the `Snapshot` type is moving there). We'd then have a `prottest/memstore` -> `protocol` -> `prottest` -> `prottest/memstore` dependency cycle.

Would you still like me to create a new `prottest/memstore` even if it doesn't let us get rid of `protocol_test`?

kr: I admit I would not have expected things like moving Snapshot to another package to fall out of rewriting package validation in terms of bc.Y instead of bc.X. Maybe we should pull on this thread a little more to understand the ultimate motivation for it?

bobg: I moved `Snapshot` into `protocol` based on the observation that the `Chain` and `Snapshot` objects together embody the blockchain state described by the spec; neither one does it alone. It was a small part of much larger streamlining.

However, I was easily able to [back out that one piece of it](https://github.com/chain/chain/commit/3872e6ede101fc0c38fffd2cbd2e6226141638c1) on the `entries-validation` branch and it does in fact clarify the dependencies, so I'll repeat that here momentarily.

bobg:
Updated (and considerably less consequential), PTAL

kr:
LGTM

PR #795, opened 2017-03-22
net/http/httperror: new package

jbowens:
Move error formatting, logging and response writing
from core to a new httperror package.

`signerd` can then be updated to use this package
instead of duplicating its logic.


bobg:
:eyes:

net/http/httperror/httperror.go
```diff
@@ -0,0 +1,101 @@
...
+// response format.
+type Formatter struct {
+	Default     Info
+	IsTemporary func(info Info, err error) bool
+	Errors      map[error]Info
+}
+
+// Format builds an error Response body describing err by consulting
+// the f.Errors lookup table. If no entry is found, it returns f.Default.
+func (f Formatter) Format(err error) (body Response, info Info) {
```
bobg: Why return `info` separately from `body`, if it's also always available as `body.Info`? Is it just a convenience? Does it conform to some interface?

core/transact.go
```diff
@@ -80,7 +81,13 @@ func (a *API) buildSingle(ctx context.Context, req *buildRequest) (*txbuilder.Te
 	maxTime := time.Now().Add(ttl)
 	tpl, err := txbuilder.Build(ctx, req.Tx, actions, maxTime)
 	if errors.Root(err) == txbuilder.ErrAction {
-		err = errors.WithData(err, "actions", errInfoBodyList(errors.Data(err)["actions"].([]error)))
+		// Format each of the inner errors contained in the data.
+		var formattedErrs []httperror.Response
+		for _, innerErr := range errors.Data(err)["actions"].([]error) {
```
bobg: Can `txbuilder.ErrAction` ever fail to have an `errors.Data(err)["actions"].([]error)`?

jbowens: Nope. `txbuilder.ErrAction` is only returned if there are 1+ action-specific errors.

net/http/httperror/httperror.go
jbowens: goooood q. I'll take a look to see if we can just return Response. I just blindly moved and adapted this function from `core.errInfo`

bobg:
LGTM pending a resolution to https://github.com/chain/chain/pull/795#discussion_r107560865

PR #796, opened 2017-03-22
errors: define new WrappedError interface

bobg:
The existing `wrapperError` type satisfies the new `WrappedError` interface, which is now also satisfied by `vm.Error`. This will allow finer-grained unit testing of the "entries"-based validation logic to follow.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -11,7 +11,13 @@ func New(text string) error {
 	return errors.New(text)
 }
 
-// wrapperError satisfies the error interface.
+// WrappedError is an error that can wrap another error.
+type WrappedError interface {
+	error
+	Wrapped() error
+}
+
+// wrapperError satisfies the WrappedError interface.
 type wrapperError struct {
 	msg    string
 	detail []string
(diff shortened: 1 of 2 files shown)


bobg:
PTAL

kr:
At first glance, I would think the validation unit tests could be written in terms of vm.Error directly. Along the lines of

```go
if ve, ok := errors.Root(err).(vm.Error); !ok || ve.Err != wantErr {
	t.Error(...)
}
```

Is that infeasible?

bobg:
Not infeasible, it's just a lot prettier when [it looks like this](https://github.com/chain/chain/blob/4855597892b4398b7cf26cf53835eb3bf46c2d32/protocol/bc/mux_test.go#L61-L63):

```
if errors.Root(err) != c.err {
  ...
}
```

without regard to whether `c.err` is a `vm.Error` or something else (as in the linked code sample above).

kr:
Can I suggest starting with a helper function in the tests' package?

If this becomes a common pattern, then maybe it should "graduate" into chain/errors. Or maybe we can find a good alternative that doesn't require adding more stuff here.

bobg:
> Can I suggest starting with a helper function in the tests' package?

OK, [done](https://github.com/chain/chain/commit/bb393f0d98742417f14b3ab49228c0a9609140ca) (on the `entries-validation` branch); withdrawing this PR.

PR #797, opened 2017-03-22
protocol/vm: remove the dependency on protocol/bc

bobg:
This resolves a long-standing TODO item, and will also allow `protocol/bc` to invoke VM validation (inverting the dependency) in the upcoming "entries"-based implementation of validation.

The dependency is broken by introducing a new object, `vm.Context`, that contains the execution context for the virtual machine.

The entrypoints to the VM are now consolidated into the single function `vm.Verify`, whose sole parameter is an appropriate `Context` object.

The unrelated `VMContexts` object in the `TxHashes` structure is no longer needed; it has been reduced to a list of txsighash values.

Some tests in `protocol/vm` depend on `protocol/bc` and have been moved into the new `vm_test` package to break the dependency cycle. Future work: remove even those files' dependence on `protocol/bc` by avoiding construction of full transaction and block objects.


bobg:
PTAL

kr:
:eyes:

protocol/vm/vm_test.go
```diff
@@ -1,4 +1,4 @@
-package vm
+package vm_test
```
kr: I see this is mentioned in the commit message

> Some tests in protocol/vm depend on protocol/bc and have been moved into the new vm_test package to break the dependency cycle.

The ideal way to handle this would be to make the tests not depend on protocol/bc at all, and have them set up context objects directly. Do you consider that future work? Should it be mentioned in a TODO or in the commit message? Or would it be easy enough to do right here in this PR?

kr: Basically I think using actual blocks and transactions here is more of an integration test, and we can probably exercise this package more efficiently (more coverage with less test code) if we focus on unit tests.

protocol/vm/vmcontext.go
```diff
@@ -0,0 +1,27 @@
+package vm
+
+type VMContext interface {
```
bobg: > The name vm.VMContext stutters, it should prob be vm.Context IMO.

Ha, I had it that way first but thought better of it when I considered the potential for confusion. Your point about `vm.Context` is well-taken, though.

> This type might work better as a struct with optional fields

[I tried it](https://github.com/chain/chain/commit/df6946c7e2d255222758ee2b720dbe3ccf1656b8) and expected to hate it, but it's a definite improvement. Thanks!

protocol/vm/vm_test.go
bobg: Yes, I consider this future work, and have updated the commit message.

bobg:
Updated, PTAL

protocol/vm/introspection.go
```diff
@@ -50,207 +38,163 @@ func opCheckOutput(vm *virtualMachine) error {
...
+}
 
+func opNonce(vm *virtualMachine) error {
 	err := vm.applyCost(1)
 	if err != nil {
 		return err
 	}
 
-	return vm.push(ii.Nonce, true)
+	if vm.context.Nonce == nil {
```
kr: Doing this check after applying the cost (rather than before) looks like a behavior change from what we had before. Is that intentional? Should it be mentioned in the commit message?

My assumption going in was that this PR would be pretty much a pure behavior-preserving refactoring.

protocol/vm/vm.go
```diff
@@ -47,94 +40,42 @@ var ErrFalseVMResult = errors.New("false VM result")
...
-			program:  prog,
-			runLimit: initialRunLimit,
-		}
-		for _, arg := range args {
-			err := vm.push(arg, false)
-			if err != nil {
-				return err
+		if r := recover(); r != nil {
+			if rErr, ok := r.(error); ok {
+				err = errors.Sub(ErrUnexpected, rErr)
```
kr: This looks like it's doing some new work to preserve error information if the panic value is an error. Is this really related to switching the dependency direction? I don't understand how. I think it's a good change, but maybe in its own PR?

protocol/vm/context.go
```diff
@@ -0,0 +1,31 @@
...
+	Code      []byte
+	Arguments [][]byte
+
+	TxVersion *uint64
+
+	BlockHash            *[]byte
+	BlockTimeMS          *uint64
+	NextConsensusProgram *[]byte
+
+	TxSigHash        *[]byte
```
kr: A slice is capable of being nil on its own, so if you want, the slice fields here don't have to be pointers. (Same as for the func field).

protocol/vm/context.go
```diff
@@ -0,0 +1,31 @@
...
+// By convention, variables of this type have the name context, _not_
+// ctx (to avoid confusion with context.Context).
+type Context struct {
+	VMVersion uint64
+	Code      []byte
+	Arguments [][]byte
+
+	TxVersion *uint64
+
+	BlockHash            *[]byte
```
kr: It would be nice to document how the optional fields work, maybe in the godoc for type Context or func Verify. Something brief would be fine, e.g. something like what I wrote in an earlier comment:

```
// Fields below this point are optional. If the corresponding opcode
// executes on a nil field, Verify will return ErrContext.
```

protocol/bc/vmcontext.go
```diff
@@ -0,0 +1,89 @@
...
+		assetID          = tx.Inputs[index].AssetID()
+		assetIDBytes     = assetID[:]
+		amount           = tx.Inputs[index].Amount()
+		inputRefDataHash = hashData(tx.Inputs[index].ReferenceData).Bytes()
+		txRefDataHash    = hashData(tx.ReferenceData).Bytes()
+	)
+
+	checkOutput := func(index uint64, refdatahash []byte, amount uint64, assetID []byte, vmVersion uint64, code []byte) (bool, error) {
+		if index >= uint64(len(tx.Outputs)) {
+			return false, vm.ErrBadValue
```
bobg: Actually, now that you mention it, [the spec is silent](https://github.com/chain/chain/blob/main/docs/protocol/specifications/vm1.md#checkoutput) on what to when `index` is out of range. (And the updated spec on the `txgraph-spec` branch is similarly silent but includes outdated language about "the number of outputs.")

I agree that just returning `false` would be preferable. @danrobinson @oleganza would this be acceptable, and if so will you clarify the spec?

kr: The spec says

> Fails if the number of outputs is less or equal to index.

which I interpret to explicitly require the current behavior.

protocol/vm/introspection.go
bobg: It's a pretty inconsequential behavior change that only manifests when both error conditions are true (it's the wrong context _and_ the runlimit is exhausted). The old code would detect one of the errors first, the new code detects the other. (It still works the same way in the case of zero errors or one.) I don't think it's worth mentioning or changing.

protocol/bc/vmcontext.go
oleganza: I think this line says that CHECKOUTPUT should fail execution if asked to find out-of-bounds output:

> 3. Fails if the number of outputs is less or equal to index.

Same as in txgraph spec.

oleganza: @bobg i guess "outputs" should be "destinations" and moved into appropriate part (under "if we are in the Mux").

protocol/vm/context.go
bobg: The slices that are hashes could be `nil` to mean "absent," because we know they should be non-empty if present. But other slices in this struct could be "empty and present," so those do need to be pointers to slices. My (weak) aesthetic preference was to handle all the slices that way.

protocol/bc/vmcontext.go
bobg: > the spec is silent

> Fails if the number of outputs is less or equal to index.

:blush:

Carry on...

protocol/vm/introspection.go
kr: Cool. Just something to keep in mind, stuff like this makes it harder to review. Because I'm asking myself (and then debating whether I should bring it up in review feedback) "does this look like it was on purpose? will it matter?" And anyone who reads this commit in the future will be left with the same questions (unless they find their way to this discussion thread).

If a PR is really a pure refactoring, then review is more straightforward, because there won't _be_ any behavior changes. And if there are, the reviewer can just flag them and then move on.

protocol/vm/context.go
kr: There's a difference between a nil slice (which would mean absent) and a zero-length but non-nil slice (which would mean empty and present). So they don't need to be pointers to slices. But I don't feel strongly about it, I just wanted to raise the possibility.

protocol/bc/vmcontext.go
kr: Did you have thoughts on my actual comment above?

> But since it can't [be a simple predicate], it might look a little cleaner to have it return `error` that's one of `vm.ErrBadValue` or `vm.NotFound` or `nil`.

protocol/vm/context.go
bobg: Oh good point. But, I don't know, I don't much like the idea of relying on the difference between `[]byte(nil)` and `[]byte{}` because I've seen one turn into the other too easily. Maybe just pilot error, but still.

kr:
LGTM

PR #798, opened 2017-03-23
dashboard: replace Map with object in flash message state

dominic:
Safari 8 and earlier does not natively support the `Map` object. Rather
than adding a polyfill, this replaces the single use of a `Map` as a 
data structure with a plain JavaScript object.

### DIFF
--- a/dashboard/src/features/app/reducers.js
+++ b/dashboard/src/features/app/reducers.js
@@ -4,9 +4,9 @@ import { combineReducers } from 'redux'
 import uuid from 'uuid'
 
 const flash = (message, title, type) => ({ message, title, type, displayed: false })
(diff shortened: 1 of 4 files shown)


dashboard/src/features/app/reducers.js
```diff
@@ -4,82 +4,84 @@ import { combineReducers } from 'redux'
 import uuid from 'uuid'
 
 const flash = (message, title, type) => ({ message, title, type, displayed: false })
-const success = (message, title) => flash(message, title, 'success')
-const error = (message, title) => flash(message, title, 'danger')
+const addFlash = (state, f) => ({...state, [uuid.v4()]: f})
```
jeffomatic: I slightly prefer the original function names, e.g. `success` and `error`.

PR #799, opened 2017-03-23
protocol/validation: remove import of chain/protocol/vm

kr:
We were importing chain/protocol/vm for some constant values that don't affect the meaning of the tests in merkle_test.go. It's useful to remove edges in the dependency graph where possible so that future refactoring is less unnecessarily constrained.

### DIFF
--- a/protocol/validation/merkle_test.go
+++ b/protocol/validation/merkle_test.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"chain/protocol/bc"
-	"chain/protocol/vm"
 )
 
 func TestCalcMerkleRoot(t *testing.T) {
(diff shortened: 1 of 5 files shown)


kr:
PTAL

protocol/validation/merkle_test.go
```diff
@@ -16,31 +15,31 @@ func TestCalcMerkleRoot(t *testing.T) {
 	}{{
 		witnesses: [][][]byte{
 			[][]byte{
-				vm.Int64Bytes(1),
+				{1, 0, 0, 0, 0, 0, 0, 0},
```
bobg: This works, but it isn't the same as `vm.Int64Bytes(1)` - that's simply `[]byte{1}`.

bobg:
LGTM but see comment.

protocol/validation/merkle_test.go
kr: Oops, okay.

PR #800, opened 2017-03-23
protocol/vm: preserve error info during panic recovery

bobg:
Use `errors.Sub` to preserve the panic error as a "sub-error" during recovery in `vm.Verify`.


bobg:
PTAL

protocol/vm/vm.go
```diff
@@ -42,8 +42,12 @@ var TraceOut io.Writer
 
 func Verify(context *Context) (err error) {
 	defer func() {
-		if panErr := recover(); panErr != nil {
-			err = ErrUnexpected
+		if r := recover(); r != nil {
+			if rErr, ok := r.(error); ok {
+				err = errors.Sub(ErrUnexpected, rErr)
+			} else {
+				err = ErrUnexpected
```
kr: Since we're trying to preserve information in this PR, maybe also consider on this line:

```
err = errors.Wrap(ErrUnexpected, r)
```

kr:
LGTM with a suggestion

PR #801, opened 2017-03-23
protocol/bc: adjust exports and field names

bobg:
The `Body` field of each entry type is now exported, and the `body` method of the `Entry` interface is not. Also, some entry fields that hold the IDs of other entries have been renamed fooID. These changes will allow better access to entry data during "entries"-based validation, coming soon.

### DIFF
--- a/generated/rev/RevId.java
+++ b/generated/rev/RevId.java
@@ -1,4 +1,4 @@
 
 public final class RevId {
-	public final String Id = "main/rev2845";
+	public final String Id = "main/rev2846";
 }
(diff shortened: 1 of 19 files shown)


bobg:
PTAL

protocol/vm/vm.go
```diff
@@ -42,8 +42,12 @@ var TraceOut io.Writer
 
 func Verify(context *Context) (err error) {
 	defer func() {
-		if panErr := recover(); panErr != nil {
-			err = ErrUnexpected
+		if r := recover(); r != nil {
```
kr: Presumably this chunk of the diff will disappear after a rebase.

kr:
LGTM

bobg:
rebased

PR #802, opened 2017-03-23
protocol/bc: add an Entry pointer to ValueSource, exclude it from hashing

bobg:
This change relieves the various Entry types that use `ValueSource` objects from each having to maintain its own separate `Entry` pointer (or list), and simplifies their construction. A struct tag is used to exclude the `ValueSource`'s embedded `Entry` from hashing.


bobg:
PTAL

protocol/bc/value.go
```diff
@@ -4,4 +4,8 @@ type ValueSource struct {
 	Ref      Hash
 	Value    AssetAmount
 	Position uint64 // what int do we actually want?
+
+	// The Entry corresponding to Ref, if available
+	// The struct tag excludes the field from hashing
```
kr: Nit: periods at the ends of the sentences?

protocol/bc/map.go
```diff
@@ -124,18 +129,14 @@ func mapTx(tx *TxData) (headerID Hash, hdr *TxHeader, entryMap map[Hash]Entry, e
...
 
-	mux := NewMux(Program{VMVersion: 1, Code: []byte{0x51}}) // 0x51 == vm.OP_TRUE, minus a circular dependency on protocol/vm
-	for _, src := range muxSources {
-		// TODO(bobg): addSource will recompute the hash of
-		// entryMap[src.Ref], which is already available as src.Ref - fix
-		// this (and a number of other such places)
-		mux.addSource(entryMap[src.Ref], src.Value, src.Position)
-	}
-	_, err = addEntry(mux)
+	mux := NewMux(muxSources, Program{VMVersion: 1, Code: []byte{0x51}}) // 0x51 == vm.OP_TRUE, minus a circular dependency on protocol/vm
```
kr: No more circular dependency as of #797?

protocol/bc/entry.go
```diff
@@ -118,7 +118,13 @@ func writeForHash(w io.Writer, c interface{}) error {
 		return nil
 
 	case reflect.Struct:
-		for i := 0; i < v.NumField(); i++ {
+		typ := v.Type()
+		for i := 0; i < typ.NumField(); i++ {
+			sf := typ.Field(i)
+			if sf.Tag.Get("entry") == "-" {
```
kr: Nice!

protocol/bc/value.go
```diff
@@ -4,4 +4,8 @@ type ValueSource struct {
 	Ref      Hash
 	Value    AssetAmount
 	Position uint64 // what int do we actually want?
+
+	// The Entry corresponding to Ref, if available
```
kr: Are there any rules for when this is or isn't set? It would be comforting to know what to expect.

bobg: Added a little commentary.

protocol/bc/entry.go
bobg: How do you like "entry" as the tag?

bobg:
Updated, PTAL

protocol/bc/entry.go
kr: Seems perfect to me. Tags aren't so pervasive that I would be worried about [clashing with something else](https://github.com/golang/go/wiki/Well-known-struct-tags), and it's perfectly clear in context.

kr:
LGTM

PR #803, opened 2017-03-24
dashboard: prefix flexbox styles

dominic:
Safari 8 does not support unprefixed flexbox directives.


jeffomatic:
LGTM

dominic:
We can do this with autoprefixer config instead. Closing for now.

PR #804, opened 2017-03-24
protocol/bc: remove TxHashes, introduce TxEntries

bobg:
This change removes the `TxHashes` object from a `bc.Tx` and replaces it with a `TxEntries` object. `TxHashes` contained selected information extracted from the "entries"-based representation of a transaction. `TxEntries` _is_ (a wrapper for) the "entries"-based representation.

This will allow validation to work on a `TxEntries` object directly, among other things.

Eventually, `TxEntries` will replace the `Tx` and `TxData` types entirely.


bobg:
PTAL

protocol/bc/transaction.go
```diff
@@ -19,35 +18,43 @@ const CurrentTransactionVersion = 1
...
 	if err != nil {
 		return errors.Wrap(err)
 	}
 
-	tx.TxHashes = *hashes
+	tx.TxEntries = txEntries
 	return nil
 }
 
+func (tx *Tx) IssuanceHash(n uint32) Hash {
```
kr: This parameter type should be int so we don't have to do conversions at the call sites. This was one of the NCC group recommendations from their first audit.

kr:
:eyes:

protocol/bc/txtransaction.go
```diff
@@ -7,6 +7,29 @@ import (
 	"chain/errors"
 )
 
+type TxEntries struct {
```
kr: It would be great if, when the dust all settles, the main type we end up dealing with for a transaction still ends up being called `Tx`. If this is destined to be that type, can we find a way to name it that now (e.g. by putting these fields in the existing Tx type) or rename it later?

protocol/bc/txtransaction.go
```diff
@@ -26,8 +49,8 @@ func ComputeOutputID(sc *SpendCommitment) (h Hash, err error) {
 	return h, nil
 }
 
-// ComputeTxHashes returns all hashes needed for validation and state updates.
-func ComputeTxHashes(oldTx *TxData) (hashes *TxHashes, err error) {
+// TxHashes returns all hashes needed for validation and state updates.
+func ComputeTxEntries(oldTx *TxData) (txEntries *TxEntries, err error) {
```
kr: Should this be named something like `Convert` or `Map` or `TxEntriesFromTxData`? Am I right to expect eventually this function will disappear?

protocol/bc/txtransaction.go
```diff
@@ -26,8 +49,8 @@ func ComputeOutputID(sc *SpendCommitment) (h Hash, err error) {
 	return h, nil
 }
 
-// ComputeTxHashes returns all hashes needed for validation and state updates.
-func ComputeTxHashes(oldTx *TxData) (hashes *TxHashes, err error) {
+// TxHashes returns all hashes needed for validation and state updates.
```
kr: s/TxHashes/ComputeTxEntries/

protocol/bc/txtransaction.go
bobg: It will definitely become `Tx` later. Added a comment to that effect.

bobg:
Updated, PTAL

protocol/bc/transaction.go
```diff
@@ -19,35 +18,43 @@ const CurrentTransactionVersion = 1
...
-	tx.TxHashes = *hashes
+	tx.TxEntries = txEntries
 	return nil
 }
 
+func (tx *Tx) IssuanceHash(n int) Hash {
+	return tx.TxEntries.TxInputIDs[n]
+}
+
+func (tx *Tx) OutputID(outputIndex uint32) Hash {
```
kr: Same recommendation here as for IssuanceHash, this param should be int. (Oh I see it was uint32 before. This isn't a new method, it just moved up here. But now seems like an ok time to fix it.)

kr:
LGTM after one fix

PR #805, opened 2017-03-25
protocol/bc: add Entry witness and destination data

bobg:
Add witness and destination data to various entry types in preparation for "entries"-based validation.


bobg:
PTAL

bobg:
PTAL

kr:
:eyes:

protocol/bc/map.go
```diff
@@ -68,14 +69,20 @@ func mapTx(tx *TxData) (headerID Hash, hdr *TxHeader, entryMap map[Hash]Entry, e
...
+			)
 
 			if len(oldIss.Nonce) == 0 {
 				if firstSpend == nil {
 					err = fmt.Errorf("nonce-less issuance in transaction with no spends")
 					return
 				}
-				nonce = firstSpend
+				anchor = firstSpend
+				setAnchored = func(id Hash, iss *Issuance) {
```
kr: This could be `setAnchored = firstSpend.SetAnchored` if you want. Same for nonce below.

protocol/bc/value.go
```diff
@@ -3,7 +3,17 @@ package bc
 type ValueSource struct {
 	Ref      Hash
 	Value    AssetAmount
-	Position uint64 // what int do we actually want?
+	Position uint64
+
+	// The Entry corresponding to Ref, if available
+	// The struct tag excludes the field from hashing
+	Entry `entry:"-"`
```
kr: Really this comment belongs on ValueDestination.Entry but that line isn't commentable.

Should ValueDestination.Entry have the `entry:"="` tag? It doesn't seem like we ever hash ValueDestination objects.

bobg:
Updated, PTAL

This PR is all that stands in the way of beginning the final "entries validation" code review.

kr:
:eyes:

kr:
LGTM

PR #832, opened 2017-03-30
dashboard: link docs to public site

dominic:


jeffomatic:
I think we should use specific version links, so that your version of dashboard is synched with the version of the docs that the links lead to.

dominic:
Right - it won't always be latest.

dominic:
@jeffomatic PTAL again

dashboard/src/features/app/components/Navigation/Navigation.jsx
```diff
@@ -100,11 +100,21 @@ class Navigation extends React.Component {
...
 export default connect(
-  state => ({
-    routing: state.routing, // required for <Link>s to update active state on navigation
-    showSync: state.core.configured && !state.core.generator,
-    production: state.core.production,
-  }),
+  state => {
+    let docVersion = ''
+
+    const versionComponents = state.core.version.match('^([0-9]\\.[0-9])\\.(.*)')
```
jeffomatic: I suspect you don't need the second set of parentheses in the regex here.

dominic: :+1:

PR #837, opened 2017-03-30
cmd/cored: remove docs from cored and related commands

dominic:
This reduces the binary size of `cored` from 60MB to 20MB. Includes Dashboard 
compiled to point to chain.com/docs for online docs.


kr:
Is this really `all`? It seems more like `generated/docs` or maybe `cmd/cored` (since the main stated benefit is reducing the size of cored).

jeffomatic:
What's the update to dashboard.go here?

dominic:
@kr I'm happy to change it to either. I thought the removal of the ops command, core HTTP responder, and generated docs folder made it a little ambiguous. `cmd/cored` seems the most fitting if it's not `all`.

dominic:
@jeffomatic The dashboard build includes the docs link pointing at `chain.com/docs` instead of `localhost:1999/docs`. To make it easier to backport, it might be easier to just do a generate PR next.

dominic:
@jeffomatic Removed Dashboard compile - PTAL again

PR #839, opened 2017-03-30
cmd/slashland: fix logic to avoid empty commits

### LINKED ISSUE #830: chainbot still adds empty auto-rev-id commits
The `git diff-index HEAD` command returns a nonempty diff when a file's mtime has changed even if its contents are the same. This might be a git bug, not sure. Some experimenting revealed that:

- Attempting to do a commit `git commit -m foo .` caused subsequent `diff-index HEAD` calls to return an empty diff.
- Running `git add .` also caused subsequent `diff-index HEAD` calls to return an empty diff.

Since we are about to commit everything anyway, it won't hurt to run git add just before we test for dirty working tree. This seems like it'll be an effective workaround.

kr:
Running 'git commit' (without --allow-empty) will fail
when the file contents haven't changed, regardless of
the files' time stamps. We want to detect this situation
ahead of time and skip committing when the tree is
clean.

Unfortunately, diff-index, which is what we use to check
for a clean tree, *does* consider timestamps when
computing the diff; fortunately, running 'git add' fixes
this and causes the behavior of diff-index to match what
commit looks for in its "nothing to commit" error
message.

So now we explicitly add the files before checking to
see if the tree is clean.

This also gives us enough confidence to remove the
--allow-empty check on commit. It would be preferable
for chainbot simply to fail, alerting us that there's a
problem, rather than producing an empty commit and
continuing.

Fixes #830.

### DIFF
--- a/cmd/slashland/main.go
+++ b/cmd/slashland/main.go
@@ -274,6 +274,22 @@ func commitRevIDs(landdir, baseBranch string) error {
 		}
 	}
 
+	// We have to add the files here for a weird reason:
+	// Running 'git commit' (without --allow-empty) will fail when
+	// the file contents haven't changed, regardless of the files'
+	// time stamps. We want to detect this situation ahead of time
+	// and skip committing when the tree is clean.
+	// Unfortunately, diff-index *does* consider timestamps when
+	// computing the diff; fortunately, running 'git add' fixes this
+	// and causes the behavior of diff-index to match what commit
+	// looks for in its "nothing to commit" error message.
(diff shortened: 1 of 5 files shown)


kr:
PTAL

I'm going to deploy this on heroku for a while before landing it, so we can test it out.

kr:
I think it worked! (Still PTAL.)

kr:
PTAL

bobg:
:eyes:

bobg:
LGTM

PR #840, opened 2017-03-30
protocol: rename Chain.CommitBlock to CommitAppliedBlock

bobg:
The intent here is to clarify the sequence of events: validate, apply, commit. It's part of a larger simplification of the validation API, most of which happened in (and leading up to) https://github.com/chain/chain/pull/788.


bobg:
PTAL

kr:
LGTM

PR #841, opened 2017-03-30
protocol/validation: move consensus program exec into a new function

kr:


kr:
Not ready for review yet; the callers still need to be updated. But you can get the idea.

kr:
PTAL.

It is a little scary that none of the unit tests broke when I changed the signature of a function called ValidateBlock in package chain/protocol/validation.

bobg:
...which sounds like there oughta be a couple of unit tests in this PR.

kr:
Maybe in a separate PR, yeah.

protocol/validation/validation.go
```diff
@@ -376,15 +376,21 @@ func checkValidDest(vs *validationState, vd *bc.ValueDestination) error {
 	return nil
 }
 
+// ValidateBlockSig runs the consensus program from prev on b.
+func ValidateBlockSig(b, prev *bc.BlockEntries) error {
+	vmContext := NewBlockVMContext(b, prev.Body.NextConsensusProgram, b.Witness.Arguments)
```
bobg: Suggest some bulletproofing here against the possibility that `prev == nil`.

kr: I will update the godoc to specify that both b and prev must be non-nil.

bobg: Y'know, haha, no bulletproofing would be needed if we did the thing here that we were talking about in [in this thread](https://github.com/chain/chain/pull/788#discussion_r108822154): have callers pass not `prev` but only its `NextConsensusProgram`, which is all this function wants.

kr:
Rebased. Need to update for the new unit tests.

protocol/validation/validation.go
kr: 😮  Will do!

kr:
PTAL

bobg:
LGTM

PR #842, opened 2017-03-30
protocol/state: replace "issuance memory" with "nonce set"

bobg:
The latest spec (still on the `txgraph-spec` branch at the moment) replaces a blockchain state's "issuance memory" (for ensuring uniqueness of issuances) with a "nonce set." This implements that change.

A deferred piece of https://github.com/chain/chain/pull/788.


bobg:
PTAL

PR #843, opened 2017-03-30
protocol/bc: make MapBlock correctly transform nil

kr:
The BlockEntries equivalent of a nil Block is nil. This
change makes MapBlock cover its entire domain -- it's
valid for all values of *Block, including nil.


kr:
PTAL @bobg

bobg:
LGTM

PR #848, opened 2017-04-03
bin/upload-docs: upload specific version of docs

chrisgarvin:


bin/upload-docs
```diff
@@ -2,22 +2,24 @@
 set -ef -o pipefail
 
 bucket="chain-staging.chain.com"
-if [ "$1" = "prod" ];
+if [ "$2" = "prod" ];
```
dominic: Let's document the updated command line arguments in https://github.com/chain/chain/blob/main/docs/README.md

bin/upload-docs
```diff
@@ -2,22 +2,24 @@
...
 then
   bucket="chain.com"
 fi
 
 bucket_url=s3://$bucket
 
+git checkout $1-stable
```
dominic: This has a small issue when the script is called with no arguments. It attempts to run:

```
git checkout -stable
```

Which gives an error when it tries to parse `-s` as a flag to git:

```
error: unknown switch `s'
usage: git checkout [<options>] <branch>
...
```

Let's check that there is at least one argument at the start of the script.

dominic: This also has an unexpected side effect of leaving your git repo checked out to a different repository than you started with. I think it's worth it to us to copy the `build-cored-release` system of cloning the git repo to a tmp directory, and doing our checkout and builds there:

https://github.com/chain/chain/blob/main/bin/build-cored-release#L28-L36

PR #850, opened 2017-04-03
protocol/bc: replace Hash with a protobuf-generated type

bobg:
This is in preparation for replacing legacy `protocol/bc` types with "entries," which will shortly be protobuf-generated types too.

Before that, `bc.AssetID` will get a similar conversion. A new helper type in this PR, `Byte32`, embodies the various marshaling/unmarshaling methods used by both `Hash` and `AssetID`.

This was easier, and turned out less ugly, than I expected.


bobg:
PTAL

Note that the Wercker cored test is presently failing because the `go generate` step produces a slightly different `protocol/bc/hash.pb.go`! I'll fix that but please don't let that stop you from beginning a review.

kr:
I like this, except I think Byte32 should be called Hash, and Hash should be called ProtoHash, and we should use Byte32 in more places since that's a more appropriate representation of the value. The 4 64-bit integers thing is a hack to do serialization more conviently+safely, and I'd rather keep it confined to the serialization code if possible.

For example, writing `bc.Hash{0xff00000000000000, 0, 0, 0}` is a weird way to notate what ought to be `bc.Hash{0xff}`.

And things like `h.ReadFrom(r)` is an undesirable level of indirection when it could be simply `r.Read(h[:])`.

Basically, I prefer the approach we used in the `raft` branch. This patch as it stands is relatively more disruptive to the codebase overall, and I'm not sure what the benefit is.

bobg:
> I'm not sure what the benefit is

The benefit is hypothetical at the moment. I'm trying to avoid having a mishmash of `bc.Hash` and `bc.ProtoHash` references all throughout the core, a situation I anticipate when the rest of the core starts wanting to reach into the tx graph structure and handling the hashes it finds there. The alternative is always to call `.Hash()` on a `bc.ProtoHash` whenever accessing one, but with nothing to enforce that I foresee `bc.ProtoHash` leaking out and coexisting confusingly with `bc.Hash`.

kr:
> I'm trying to avoid having a mishmash

I can understand this urge, but in practice, it seems like this patch actually makes/will make it worse. We will have a mishmash of Hash and Byte32 with no clear guidance on which one is more appropriate in a given situation. (That is in addition to the couple of drawbacks mentioned in my previous comment.)

I'd rather have ProtoHash and Hash, where the guidance is: use Hash wherever possible. (Encouraged by giving it the more palatable name.)

I anticipate it'll be more pleasant to call `.Hash()` on a `bc.ProtoHash` than to make the ProtoHash leak out, and that that will happen naturally. It's certainly possible that ProtoHash will spread, but even that seems less disruptive and more clear than (Hash+Byte32), since the distinction between ProtoHash and Hash is expressed in their names. One is for serialization. What is the difference between Hash and Byte32? It is arbitrary.

bobg:
I've written https://github.com/chain/chain/pull/863, which uses the `ProtoHash` approach (along with `ProtoAssetID` and `ProtoAssetAmount`). Take a look at it: I think the approach in _this_ PR is preferable. Although in principle it sounds straightforward to confine the `Proto` types to serialization logic, in practice the boundary is very fractal and the code is littered with conversions both to and from the `Proto` types (even though we're not using them for serialization yet at all!).

> I think Byte32 should be called Hash, and Hash should be called ProtoHash

The difference between your suggestion and what this PR does is not limited to naming. Your suggestion would prefer using the lower-level object everywhere; this PR prefers using the higher-level one. I think https://github.com/chain/chain/pull/863 demonstrates that it would be cleaner to carry the higher-level object around in most contexts. If you're still not convinced (and tbh I'm only 80% there myself) I can clarify the demonstration by writing another version of https://github.com/chain/chain/pull/863, one that uses this PR's approach to the support types.

kr:
Yeah, I see what you mean. Okay, I too am 80% sold on this.

Is there a way to get rid of or at least not export the Byte32 type? In that case I would be like 95% on board!

protocol/validation/validation_test.go
```diff
@@ -75,15 +75,15 @@ func TestTxValidation(t *testing.T) {
 		{
 			desc: "nonempty mux exthash",
 			f: func() {
-				mux.Body.ExtHash = bc.Hash{1}
+				mux.Body.ExtHash = bc.Hash{0x0100000000000000, 0, 0, 0}
```
kr: Maybe we should use little-endian so this can stay `bc.Hash{1}`???

protocol/bc/hash.go
```diff
@@ -111,12 +100,18 @@ func writeFastHash(w io.Writer, d []byte) error {
...
-	return int64(n), err
+	return h.Byte32().WriteTo(w)
 }
 
-func (h *Hash) readFrom(r io.Reader) (int, error) {
-	return io.ReadFull(r, h[:])
+// WriteTo satisfies the io.ReaderFrom interface.
+func (h *Hash) ReadFrom(r io.Reader) (int64, error) {
+	var b32 Byte32
+	n, err := b32.ReadFrom(r)
```
kr: I'm curious why this isn't

```go
var b [32]byte
n, err := io.ReadFull(r, b[:])
if err != nil {
    return n, err
}
h.Set(b)
return n, nil
```

protocol/bc/hash.go
```diff
@@ -1,103 +1,92 @@
...
-	return string(b)
+func (h Hash) Byte32() (b32 Byte32) {
+	binary.BigEndian.PutUint64(b32[0:8], h.V0)
+	binary.BigEndian.PutUint64(b32[8:16], h.V1)
+	binary.BigEndian.PutUint64(b32[16:24], h.V2)
+	binary.BigEndian.PutUint64(b32[24:32], h.V3)
+	return b32
+}
+
+func (h *Hash) FromByte32(b32 Byte32) {
```
kr: Could this be defined as

```go
func (h *Hash) Set(b [32]byte)
```

or even

```go
func NewHash(b [32]byte) Hash
```

?

(more interested in the signature than the name)

protocol/bc/hash.go
```diff
@@ -1,103 +1,92 @@
...
+}
+
+// Hash represents a 256-bit hash.  Data structure defined in
+// hash.proto.
 
-// String returns the bytes of h encoded in hex.
-func (h Hash) String() string {
-	b, _ := h.MarshalText() // #nosec
-	return string(b)
+func (h Hash) Byte32() (b32 Byte32) {
```
kr: And likewise, could this be

```go
func (h Hash) Byte32() [32]byte
```

(Is this even actually used anywhere? I don't see it.)

protocol/bc/hash.go
```diff
@@ -1,103 +1,92 @@
...
 
 	"chain/crypto/sha3pool"
 	"chain/encoding/blockchain"
-	"chain/errors"
 )
 
-// Hash represents a 256-bit hash.  By convention, Hash objects are
-// typically passed as values, not as pointers.
-type Hash [32]byte
+var EmptyStringHash Hash
```
kr: If it's not possible to initialize this in one line (`EmptyStringHash = NewHash(sha.Sum256(nil))`?) then can we have a comment here like `// initialized in func init below`?

cmd/multitool/multitool.go
```diff
@@ -510,5 +510,6 @@ func verify(args []string) {
 }
 
 func zerohash(_ []string) {
-	fmt.Println(bc.Hash{}.String())
+	var h bc.Hash
+	fmt.Println(h.String())
```
kr: Is this a superfluous diff?

core/asset/asset.go
```diff
@@ -118,7 +118,8 @@ func (reg *Registry) Define(ctx context.Context, xpubs []chainkd.XPub, quorum in
 		return nil, err
 	}
 
-	defhash := sha3.Sum256(rawDefinition)
+	var defhash bc.Hash
+	defhash.FromByte32(sha3.Sum256(rawDefinition))
```
kr: Having a NewHash function would make this code look more like what was here before:

```go
defhash := bc.NewHash(sha3.Sum256(rawDefinition))
```

core/transact_test.go
```diff
@@ -79,22 +79,24 @@ func TestRecordSubmittedTxs(t *testing.T) {
...
 	testCases := []struct {
-		hash   bc.Hash
+		hash   bc.Byte32
 		height uint64
 		want   uint64
 	}{
-		{hash: bc.Hash{0x01}, height: 2, want: 2},
-		{hash: bc.Hash{0x02}, height: 3, want: 3},
-		{hash: bc.Hash{0x01}, height: 3, want: 2},
+		{hash: bc.Byte32{0x01}, height: 2, want: 2},
```
kr: What do you think of 

```go
{hash: bc.NewHash([32]byte{0x01}), height: 2, want: 2},
```

or (if Hash is little-endian)

```
{hash: bc.Hash{0x01}, height: 2, want: 2},
```

protocol/bc/asset.go
```diff
@@ -10,14 +10,14 @@ import (
 
 // AssetID is the Hash256 of the issuance script for the asset and the
 // initial block of the chain where it appears.
-type AssetID [32]byte
+type AssetID Byte32
```
kr: Would it be easier to define this in terms of Hash somehow? Or does that have to be later.

protocol/bc/asset.go
```diff
@@ -10,14 +10,14 @@ import (
...
 
-func (a AssetID) String() string                { return Hash(a).String() }
-func (a AssetID) MarshalText() ([]byte, error)  { return Hash(a).MarshalText() }
-func (a *AssetID) UnmarshalText(b []byte) error { return (*Hash)(a).UnmarshalText(b) }
-func (a *AssetID) UnmarshalJSON(b []byte) error { return (*Hash)(a).UnmarshalJSON(b) }
-func (a AssetID) Value() (driver.Value, error)  { return Hash(a).Value() }
-func (a *AssetID) Scan(b interface{}) error     { return (*Hash)(a).Scan(b) }
+func (a AssetID) String() string                { return Byte32(a).String() }
```
kr: If we wanted to keep `type AssetID [32]byte` for now, these functions could look like

```go
func (a AssetID) String() string                { return NewHash(a).String() }
```

(assuming the existence of a NewHash function)

protocol/bc/byte32.go
```diff
@@ -0,0 +1,89 @@
...
+	"database/sql/driver"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"chain/errors"
+)
+
+type Byte32 [32]byte
```
kr: As far as I can tell, this type is mostly only used inside Hash's methods, and it would not be too hard to inline some of these or lightly rearrange things and get rid of this type entirely. How does that sound?

bobg: Perfect. I had thought that `Byte32` would contain functionality common to both `Hash` and `AssetID`, but on the `entry-protos4` branch I wrote `AssetID` in terms of `Hash` quite naturally, so there's no need for a separate type to factor anything out to.

protocol/bc/asset.go
bobg: Already done on the `entry-protos4` branch, looks great.

protocol/validation/validation_test.go
bobg: Went with your other suggestion (at least for now) of `bc.NewHash([32]byte{1})`

protocol/bc/hash.go
bobg: Used elsewhere in hash.go, and also in one spot in `signerd` (a separate PR).

protocol/bc/hash.go
bobg: Added `NewHash` for allocating a new object, kept `FromByte32` for populating an existing one.

protocol/bc/hash.go
bobg: Simply not to repeat what `byte32` is already doing (and which I was supposed to reuse in `asset.go` but hadn't, and now have).

cmd/multitool/multitool.go
bobg: No: the protobuf `String` method takes a pointer receiver and so needs an addressable object.

protocol/bc/asset.go
bobg: (Sorry, that comment from last night was a little rushed and unclear.) The protobuf version of `AssetID` will be defined in terms of `Hash`, but for this interim change I prefer to keep the representation of `AssetID` unchanged.

protocol/bc/byte32.go
bobg: (Clarifying another late-night comment): it's convenient in this PR to put some common code into `byte32`, but the very next PR will reimplement `AssetID` and remove `byte32`.

(Unless you think I should combine the next PR with this one?)

bobg:
Updated, PTAL

Wercker is still generating a different `hash.pb.go` than I am locally. The difference is that it's producing accessors for the `fixed64` fields, which in proto3 [it's not supposed to](https://developers.google.com/protocol-buffers/docs/reference/go-generated#singular-scalar-proto3). Does Wercker have an out-of-date `protoc`?

bobg:
Walking back [my comment above](https://github.com/chain/chain/pull/850#discussion_r109978447):

> the very next PR will reimplement `AssetID` and remove `byte32`

After spending a fair amount of time on https://github.com/chain/chain/pull/887 with `AssetID` defined as

```
message AssetID {
  Hash hash = 1;
}
```

I discovered that I didn't want such a basic type to have a pointer to another object in it (alas, protobufs do not allow embedding of message fields), chiefly since that prevents `==` working as we'd like, and consequently `AssetID` objects can't be map keys, etc. So `AssetID` became essentially a duplicate of `Hash`, both implemented in terms of `byte32`.

bobg:
rebased

kr:
I really want to look at this but I prob can't get to it today.

bobg:
PTAL

The next PR will convert AssetIDs to protobufs and [is ready and waiting](https://github.com/chain/chain/compare/protohash...protoasset?expand=1).

core/txdb/txdb_test.go
```diff
@@ -148,18 +148,12 @@ func TestGetBlock(t *testing.T) {
 		BlockHeader: bc.BlockHeader{
 			Version:           1,
 			Height:            1,
-			PreviousBlockHash: [32]byte{'1', '2', '3'},
+			PreviousBlockHash: bc.Hash{0x3132330000000000, 0, 0, 0},
```
kr: I think these would be way more readable as

```go
bc.NewHash([32]byte{'1', '2', '3'})
```

throughout this file in particular.

core/txdb/txdb_test.go
```diff
@@ -148,18 +148,12 @@ func TestGetBlock(t *testing.T) {
...
-					'A', 'B', 'C', 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-					0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-				},
-				AssetsMerkleRoot: bc.Hash{
-					'X', 'Y', 'Z', 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-					0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-				},
-				ConsensusProgram: []byte("test-output-script"),
+				TransactionsMerkleRoot: bc.Hash{0x4142430000000000, 0, 0, 0},
+				AssetsMerkleRoot:       bc.Hash{0x58595a0000000000, 0, 0, 0},
```
kr: Similarly, `bc.NewHash([32]byte{'X', 'Y', 'Z'})` here for sure.

protocol/bc/asset.go
```diff
@@ -12,12 +12,12 @@ import (
 // initial block of the chain where it appears.
 type AssetID [32]byte
 
-func (a AssetID) String() string                { return Hash(a).String() }
-func (a AssetID) MarshalText() ([]byte, error)  { return Hash(a).MarshalText() }
-func (a *AssetID) UnmarshalText(b []byte) error { return (*Hash)(a).UnmarshalText(b) }
-func (a *AssetID) UnmarshalJSON(b []byte) error { return (*Hash)(a).UnmarshalJSON(b) }
-func (a AssetID) Value() (driver.Value, error)  { return Hash(a).Value() }
-func (a *AssetID) Scan(b interface{}) error     { return (*Hash)(a).Scan(b) }
+func (a AssetID) String() string                { return byte32(a).String() }
```
kr: I strongly suggest defining AssetID operations in terms of Hash, rather than defining both of them in terms of some third type. Conceptually, an asset ID *is* a hash. Practically, it'll be less code and fewer moving parts that way.

For the purposes of this PR, I am pretty sure these methods can look like

```go
func (a AssetID) String() string                { return NewHash(a).String() }
```

For the future, when AssetID is defined as a protobuf, then we might get rid of all these and instead provide just one or two conversion routines like

```go
func (a AssetID) Hash() Hash { return Hash(a) }
func (a *AssetID) Set(v [32]byte) { (*Hash)(a).Set(v) }
```

something along those lines.

protocol/bc/asset.go
```diff
@@ -45,23 +45,31 @@ func ComputeAssetID(prog []byte, initialBlockID Hash, vmVersion uint64, data Has
 	return def.ComputeAssetID()
 }
 
+func (a *AssetID) ReadFrom(r io.Reader) (int64, error) {
```
kr: Can we hold off on adding these methods until AssetID is a protobuf? (Next PR right?)

protocol/bc/byte32.go
```diff
@@ -0,0 +1,94 @@
...
+	"database/sql/driver"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"chain/errors"
+)
+
+type byte32 [32]byte
```
kr: Given that (IMO) AssetID should be defined in terms of Hash both in the short term and in the long term, I would really like to see this type eliminated from this PR.

protocol/bc/hash.go
```diff
@@ -1,103 +1,92 @@
...
+
+func (h Hash) Byte32() (b32 [32]byte) {
+	binary.BigEndian.PutUint64(b32[0:8], h.V0)
+	binary.BigEndian.PutUint64(b32[8:16], h.V1)
+	binary.BigEndian.PutUint64(b32[16:24], h.V2)
+	binary.BigEndian.PutUint64(b32[24:32], h.V3)
+	return b32
+}
+
+func (h *Hash) FromByte32(b32 [32]byte) {
```
kr: This is bikeshed territory, but I think this method should be called either Set or Init, for the same reason the constructor isn't (and shouldn't be) NewHashFromByte32, and because it's a side-effect procedure so its name should be a verb saying what it does.

If we had several different types that could initialize a Hash, then it would make sense to call them SetFoo, SetBar, etc. But there is just one. (And even if there were several, the "main" one should still probably be just Set.)

kr:
Check out 2dd27b338131f7b4e3fcc0c5829210b86735df5f as an illustration of the main feedback in https://github.com/chain/chain/pull/850#discussion_r110484240.

protocol/bc/hash.go
kr: Actually, I think probably we should just delete this method. Any call site that looks like

```go
h.FromByte32(b)
```

can be rewritten as one of

```go
h = NewHash(b) // or
*h = NewHash(b)
```

core/txdb/snapshot_test.go
```diff
@@ -49,48 +49,48 @@ func TestReadWriteStateSnapshot(t *testing.T) {
 		deletedNonces []bc.Hash
 	}{
 		{ // add a single hash
-			inserts: []bc.Hash{{0x01}},
+			inserts: []bc.Hash{{0x0100000000000000, 0, 0, 0}},
```
bobg: Oops, I had already done this on the `entry-protos4` branch but not here. Done.

core/txdb/txdb_test.go
bobg: Same.

core/txdb/txdb_test.go
bobg: Same.

bobg:
Regarding how to implement `Hash` and `AssetID` (both before and after `AssetID` is protobuffified):

I'm torn. On the one hand, it's true that an asset ID is just a hash. For clarity and to avoid overabstraction, `AssetID` should simply be eliminated, and the type of all hash values should just be `Hash`. On the other hand, type safety requires that `AssetID` and `Hash` be distinct types. I don't really know which of those options I like better. But if we get rid of `AssetID` this particular issue goes away.

Assuming we keep both types (and, who knows, maybe even introduce other similar duplicate types):

Once both types are protobufs, it's true that we could implement marshaling/unmarshaling (etc) in just `Hash` and access those implementations from `AssetID` by typecasting. But that is not an appealing option because the equality of the implementation of the two types is incidental. `AssetID` could choose an entirely different representation (or simply different field names) than `Hash` and then any type conversion expression would be invalid. IOW depending on `Hash(assetID)` working feels like a hack.

It also doesn't pass the necessary/sufficient test: the choice of putting behavior in `Hash` and calling it from `AssetID` vs. putting it in `AssetID` and calling it from `Hash` is arbitrary.

(It'd be great if we could write a `.proto` file saying "use the definition of `Hash` to create a duplicate new type `AssetID`." We _could_ do `type AssetID Hash` in Go, but then we wouldn't have an `AssetID` message type for protobuf fields.)

_Until_ both types are protobufs, the idea of passing an `AssetID` to `NewHash` just to get an object on which we can call `MarshalText` etc. seems wasteful.

The only approach that avoids all these objections is to factor out the common behavior to a shared type, `byte32`. I'm curious to know what concerns you have about `byte32` that outweigh these objections.

bobg:
> The only approach that avoids all these objections is to factor out the common behavior to a shared type, `byte32`.

...or to duplicate the implementation in both `Hash` and `AssetID`, but that doesn't seem as good as factoring it out.

kr:
> Once both types are protobufs, it's true that we could implement marshaling/unmarshaling (etc) in just Hash and access those implementations from AssetID by typecasting. But that is not an appealing option because the equality of the implementation of the two types is incidental.

Hmm I don't see it that way. We are deliberately choosing the same representation in part because it makes the implementation more straightforward. Ideally we'd just write `type AssetID Hash` and be done with it. The only reason we can't is because both types are generated from protoc, but deliberately generating them both to have the same internal representation is conceptually equivalent and mechanically as close as we can get to actually writing `type AssetID Hash`. We should use that to our advantage.

> AssetID could choose an entirely different representation (or simply different field names) than Hash and then any type conversion expression would be invalid. IOW depending on Hash(assetID) working feels like a hack.

If we ever decided to use a different representation for AssetID (but why would we?) then in that case of course we'd change its implementation. But YAGNI. We're not trying to solve hypothetical future problems, we're solving actual current problems.

IOW we're only "depending" on Hash(assetID) working because it actually does work. We do this routinely, and no one has ever objected to it before. Again, I think we should try to treat this as effectively `type AssetID Hash` (or as close as we can get to it). And we're not stuck with that choice permanently. If in the future we wanted to make it stop working then sure, we'll update the code accordingly.

> It also doesn't pass the necessary/sufficient test: the choice of putting behavior in Hash and calling it from AssetID vs. putting it in AssetID and calling it from Hash is arbitrary.

It's not at all arbitrary. AssetID is, conceptually, a specialization of Hash. If we had several other subtypes, they would all be written in terms of Hash, not the other way around.

> (It'd be great if we could write a .proto file saying "use the definition of Hash to create a duplicate new type AssetID." We could do type AssetID Hash in Go, but then we wouldn't have an AssetID message type for protobuf fields.)

Yeah, that would be wonderful! So would being able to specify a fixed-size byte array. :(

> Until both types are protobufs, the idea of passing an AssetID to NewHash just to get an object on which we can call MarshalText etc. seems wasteful.

This is a textbook case where Knuth's famous admonition applies: "We should forget about small efficiencies, say about 97% of the time." Especially in this case since the NewHash call is about to be replace by a direct conversion to Hash in the very next PR.

> The only approach that avoids all these objections is to factor out the common behavior to a shared type, byte32. I'm curious to know what concerns you have about byte32 that outweigh these objections.

It's just more moving parts. I'm looking for ways to simplify things, reduce the amount of code, and reduce the amount of tracing through levels of indirection to figure out what's going on when you're reading the code. This byte32 type seems like excessive DRY that only serves to increase the amount of code and make it harder to follow. Consider that it takes just as much code in the Hash-as-wrapper type to call the byte32 implementations as it does to actually implement the logic directly. That's not a win.

bobg:
OK, sold, and thanks for hearing me out.

Updated, PTAL

kr:
> OK, sold, and thanks for hearing me out.

Any time. It is a pleasure. :)

:eyes:

protocol/bc/asset.go
```diff
@@ -53,11 +69,11 @@ type AssetAmount struct {
 func (a *AssetAmount) readFrom(r io.Reader) (int, error) {
 	n1, err := io.ReadFull(r, a.AssetID[:])
 	if err != nil {
-		return n1, err
+		return int(n1), err
 	}
 	var n2 int
 	a.Amount, n2, err = blockchain.ReadVarint63(r)
-	return n1 + n2, err
+	return int(n1) + n2, err
```
kr: Nit: these two `int(n1)` conversions can go back to just `n1`.

protocol/bc/asset.go
```diff
@@ -12,12 +12,28 @@ import (
...
+func (a AssetID) String() string               { h := NewHash(a); return h.String() }
+func (a AssetID) MarshalText() ([]byte, error) { return NewHash(a).MarshalText() }
+func (a AssetID) Value() (driver.Value, error) { return NewHash(a).Value() }
+func (a *AssetID) UnmarshalText(b []byte) error {
+	var h Hash
+	err := h.UnmarshalText(b)
+	*a = h.Byte32()
+	return err
+}
+
```
kr: Nit: I accidentally left an extra blank line here in my suggestion commit.

kr:
LGTM (couple of really tiny nits)

PR #855, opened 2017-04-04
protocol/validation: unit test the CheckOutput callback

kr:
This change converts some CHECKOUTPUT test cases from
chain/protocol/vm into unit tests here, and deletes the
corresponding cases from package vm.

This slghtly reduces the dependency from package vm's
tests on package bc. Ultimately, we will remove that
dependency completely.


protocol/validation/vmcontext_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"encoding/hex"
+	"fmt"
+	"testing"
+
+	"chain/errors"
+	"chain/protocol/bc"
+	"chain/protocol/vm"
+)
+
+func TestIntrospectionOps(t *testing.T) {
```
kr: oops, will rename

kr:
PTAL

kr:
PTAL

bobg:
:eyes:

protocol/validation/vmcontext.go
```diff
@@ -164,8 +110,66 @@ func NewTxVMContext(tx *bc.TxEntries, entry bc.Entry, prog bc.Program, args [][]
...
 		AnchorID:      anchorID,
 		SpentOutputID: spentOutputID,
-		CheckOutput:   checkOutput,
+		CheckOutput:   (&txContext{entry: entry}).checkOutput,
 	}
 
 	return result
 }
+
+type txContext struct {
```
bobg: I think I'd prefer a more specific name for this type such as `entryContext` or even `checkOutputContext`. "Tx" is a little overused, even allowing for the renaming simplification to come.

bobg:
LGTM with one suggestion.

PR #859, opened 2017-04-04
cmd/cored: replace prod and dev build tags with granular options

boymanjor:
This commit replaces the `prod` and `dev` build tags with
granular options to define specific behavior. This includes:
- enabling access tokens
- disabling https redirects (renamed to `disable_tls_only`)
- disabling the `/reset` endpoint
- disabling the mockhsm

The idea is to define a default, Chain development build config
and deviate as necessary for targeted builds, e.g. developer edition
and production server binaries.


boymanjor:
PTAL

kr:
:eyes:

cmd/cored/disable_mockhsm.go
```diff
@@ -1,22 +1,17 @@
-//+build prod
+//+build disable_mockhsm
```
kr: I'm not super fond of the `enable_foo` `disable_bar` naming scheme, because “enable” doesn't contribute any information. For the positive ones I'd rather call it just `foo`. That implies we should find an alternative to “disable” for the negative ones, so how about `no_` as a prefix? e.g.

```
no_loopback_authn
no_mockhsm
no_reset
plain_http
```

boymanjor: I like this. How about instead of `plain_http` we go with `no_https_redirect`. This would leave us with the default behavior of: disabled loopback authn, mockhsm, enabled reset endpoints and https redirects. The build tags would be:
```
loopback_authn
no_mockhsm
no_reset
no_https_redirect
```

boymanjor: Discussed out of band: no loopback authn will be considered a feature. The build tag will be `no_loopback_authn`.

boymanjor:
This is superseded by #864 #865 #896 & #897

PR #861, opened 2017-04-04
sdk/java: throw HTTPException from setTrustedCerts

boymanjor:
To accomodate this change, ChainException has been
extended to allow the wrapping of other exceptions. This
allows our public methods to stay consistent in throwing
subclasses of `ChainException`.


boymanjor:
PTAL

PR #862, opened 2017-04-04
cmd/cored: configure root CA pool

boymanjor:
By default, package crypto/tls uses the operating
system's set of global root CAs. This is unsuitable for
a typical production deployment, where the opeator will
run its own CA and provision its own certs.

This change provides a new environment variable:

ROOT_CA_CERTS=certs.pem
This variable holds a filename of a file containing a list
of PEM-encoded root X.509 certificates.

This is a back port of a change made in #684 with
updates to the logging call sites.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -4,11 +4,13 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/hex"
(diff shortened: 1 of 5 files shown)


boymanjor:
PTAL

kr:
LGTM, but can you include the original commit message here as well?

PR #863, opened 2017-04-04
protocol/bc: reimplement entries as protobufs

bobg:
This PR redefines the various entry types (and some support types) as protocol buffers. This will let us use `proto.Marshal` and `proto.Unmarshal` as our block serializer/deserializer when we retire the legacy `protocol/bc` types.

`Hash`, `AssetID`, and `AssetAmount` get their own protobuf-generated types: `ProtoHash`, `ProtoAssetID`, and `ProtoAssetAmount`. Unlike the approach in https://github.com/chain/chain/pull/850, these protobuf-generated types exist alongside the legacy types. Conversions between the two are performed as needed, with the legacy types preferred whenever possible.

This PR also removes `Entry` pointers from within entries themselves. For dereferencing one `Entry` object from another, a `map[Hash]Entry` has been added to the `TxEntries` wrapper type.


bobg:
Closing in favor of https://github.com/chain/chain/pull/887, which uses https://github.com/chain/chain/pull/850's approach to protobuffifying `Hash`, `AssetID`, and `AssetAmount` (i.e., it totally replaces the legacy types).

PR #865, opened 2017-04-04
cmd/cored: introduce reset build tag

boymanjor:
By default, resetting a core's database will not be allowed.
To enable this feature, cored should be built with the `reset`
build tag. This is commit finishes the work started in #859
and removes all remnants of the prod build tag.


boymanjor:
PTAL

cmd/cored/protected_db.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"chain/core/config"
+	"chain/database/pg"
+)
+
+func init() {
+	config.BuildConfig.ProtectedDB = true
+}
+
+func resetInDevIfRequested(_ pg.DB) {}
```
kr: nbd, but you can leave out the `_` if none of the parameter names in the list are used. So

```go
func resetInDevIfRequested(pg.DB) {}
```

cmd/cored/default.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"os"
+)
+
+var (
+	reset = env.String("RESET", "")
+)
+
+func init() {
+	config.BuildConfig.ProtectedDB = false
```
kr: I think the default values should be initialized here, or in config.go, but not both. If here, then there should be a comment in config.go saying "initialized in $CHAIN/cmd/cored/default.go".

boymanjor: This makes sense. We should probably opt for config.go because the build constraints will become the AND of all the build tag negations. Setting one of the build tags should exclude this file.

boymanjor:
screen shot 2017-04-10 at 10 45 56 am

boymanjor:
PTAL

boymanjor:
PTAL

docs/internal/api-spec.swagger.yml
```diff
@@ -1065,10 +1065,17 @@ definitions:
         type: string
         description: An RFC3339 timestamp indicating the last successful
           retrieval of the generator's block height.
-      is_production:
+      is_reset:
```
jbowens: I had no idea this existed. 😮

cmd/cored/reset.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"chain/core/config"
+	"chain/core/coreunsafe"
+	"chain/database/pg"
+	"chain/database/raft"
+	"chain/env"
+	"chain/log"
+	"context"
+	"fmt"
+	"os"
```
jbowens: nit: import grouping

jbowens:
lgtm, just an import grouping nit

PR #867, opened 2017-04-05
protocol/validation: unexport newBlockVMContext

kr:
Function NewBlockVMContext was exported only to satisfy tests in package vm. Instead, those tests now create their own Context object directly.

Future work:
- do the same for NewTxVMContext


kr:
PTAL

bobg:
LGTM

PR #868, opened 2017-04-05
bin/poke-ci: new utility script

jeffomatic:
So sometimes our CI system (Wercker et al) can be a bit
tempermental. Here's a script that should trigger a re-run of the
CI process.


jeffomatic:
Ha ha. PTAL.

PR #872, opened 2017-04-05
cmd/multitool: remove the multitool command

bobg:
This tool is little-used, partly obsolete, at odds with some design decisions we reached a while ago, and keeps annoyingly requiring us to track changes elsewhere in the codebase.

It will be easy and better to reimplement the pieces of this tool on an as-needed basis.


bobg:
PTAL

PR #874, opened 2017-04-05
docs: add version selector

dominic:
Dropdown offers choices sourced from https://chain.com/docs/versions.js,
a document controlled by the marketing site. 

If the current version from the URL does not match the latest released 
version in that document, an alert is shown to the user indicating a
newer version is available.


docs/js/init.js
```diff
@@ -1,10 +1,11 @@
 $(function () {
-	
+
```
jeffomatic: newline

docs/js/layout.js
```diff
@@ -9,9 +9,9 @@ function activateCurrentSidebarItem() {
 }
 
 function prepareSidebarMenu() {
-	
+
```
jeffomatic: newline

PR #883, opened 2017-04-06
docs: improve version string parsing

dominic:
Previous regex relied on the presence of `/core` in the
URL. This prevented the correct version from being detected
on multiple paths.

### DIFF
--- a/docs/js/layout.js
+++ b/docs/js/layout.js
@@ -167,11 +167,12 @@ function fixupSidenotes() {
 
 function loadVersionOptions() {
 	var currentVersion
-	var matchedVersion = location.pathname.match('/docs/([0-9].*)/core')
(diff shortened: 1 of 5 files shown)


docs/js/layout.js
```diff
@@ -167,11 +167,14 @@ function fixupSidenotes() {
 
 function loadVersionOptions() {
 	var currentVersion
-	var matchedVersion = location.pathname.match('/docs/([0-9].*)/core')
+	var matchedVersion = location.pathname.match('/docs/([\d+\\.\d+)/')
```
jeffomatic: I think you need to double-slash the `\d`, as in `\\d`. The JS quoting rules will swallow the backslash otherwise, e.g.:

```
jeff@jeff-chain ~/go/src/chain (main) $ node
> console.log('\d')
d
```

PR #886, opened 2017-04-06
protocol/vm: create a Context directly

kr:
This change converts the tests in vm_test.go to stop
using packages bc and validation to build test fixtures,
and instead define a Context object in one step. Code
coverage stays at 95.3% with less test code.

Future work: once all test files stop depending on
package bc, move the test code back into package vm
(from vm_test).


protocol/vm/vm_test.go
```diff
@@ -9,8 +9,6 @@ import (
 	"testing/quick"
 
 	"chain/errors"
-	"chain/protocol/bc"
-	"chain/protocol/validation"
 	. "chain/protocol/vm"
```
bobg: Can this file now return to being in `package vm`?

protocol/vm/vm_test.go
```diff
@@ -460,10 +416,18 @@ func TestVerifyTxInputQuickCheck(t *testing.T) {
...
-		verifyTx(tx, 0)
+
+		vctx := &Context{
+
+			VMVersion: 1,
+			Code:      program,
+			Arguments: witnesses,
+			MinTimeMS: new(uint64),
+			MaxTimeMS: new(uint64),
+			TxData:    &[]uint8{0xa7, 0xff, 0xc6, 0xf8, 0xbf, 0x1e, 0xd7, 0x66, 0x51, 0xc1, 0x47, 0x56, 0xa0, 0x61, 0xd6, 0x62, 0xf5, 0x80, 0xff, 0x4d, 0xe4, 0x3b, 0x49, 0xfa, 0x82, 0xd8, 0xa, 0x4b, 0x80, 0xf8, 0x43, 0x4a},
```
bobg: Can you add a comment explaining the significance of these numbers (or lack thereof)?

protocol/vm/vm_test.go
kr: It can, but doing that right will involve some code changes (such as using virtualMachine instead of VirtualMachine) that would obscure what we're doing here, and it's relatively easy, so I'd rather do it for all the test files together in a separate "remove export_test.go" PR.

kr:
PTAL

protocol/vm/vm_test.go
kr: I updated the description to mention this as future work.

kr:
PTAL

bobg:
LGTM

PR #887, opened 2017-04-06
protocol/bc: reimplement entries as protobufs

bobg:
This PR redefines the various entry types (and some support types) as protocol buffers. This will let us use proto.Marshal and proto.Unmarshal as our block serializer/deserializer when we retire the legacy protocol/bc types.

This PR also removes Entry pointers from within entries themselves. For dereferencing one Entry object from another, a `map[Hash]Entry` has been added to the `TxEntries` wrapper type.


bobg:
rebased

bobg:
rebased

bobg:
PTAL

kr:
What's the compatibility plan for this? At a minimum, we need to:

- migrate database objects to the new format
- provide the old format to old get-blocks clients
- increment the "network protocol" version, but see previous item — maybe we should rethink this
- allow an old-format cored and a new-format cored to coexist in the same Chain Core cluster from one version to the next

kr:
We don't need to continue doing all those things indefinitely, just long enough to let people transition. So I think in practice that means for one or two release cycles.

kr:
As far as I know we haven't yet started to figure out how to do this correctly. In a blockchain network, there are several types of parties:

- transaction creator
- transaction signer
- block generator
- block signer
- block+tx validator
- (more?)

In what order should they each be upgraded so that nothing breaks?

bobg:
> What's the compatibility plan for this?

That can be mostly punted to after this PR lands, but I have begun thinking about it. We'll want to make sure that everything can read the new format before anything starts writing it, of course. We should avoid replacing legacy-format data with new-format data - so rather than change the layout of existing RPC calls, we define new RPC calls. And rather than change the format of the `blocks.data` database column, we define a new column for protobuf-format data.

kr:
Rad, that sounds great.

> That can be mostly punted to after this PR lands

As a tactical matter, does this PR break any existing network interfaces? Because we are about to freeze for 1.2 (at the end of this week) and I don't know if we'll have time to do all the necessary followup work to ensure compatibility before then. So maybe we need to hold off merging this until the tree "opens" for 1.3.

bobg:
> does this PR break any existing network interfaces?

I believe not (though we should of course test that), since nothing is yet trying to serialize any of the new entry types for anything other than hashing, and the protobuf types `Hash` and `AssetID` and `AssetAmount` serialize the same as they always have via `MarshalText`. (And nothing outside of test code is calling `proto.Marshal` on them yet.)

kr:
Oh self-facepalm, this is just the txentries types, not the legacy tx types. Got it.

kr:
Coupe of random thoughts in there but LGTM.

core/txbuilder/witness_test.go
```diff
@@ -41,9 +41,10 @@ func TestInferConstraints(t *testing.T) {
 }
 
 func TestWitnessJSON(t *testing.T) {
+	assetID := bc.NewAssetID([32]byte{0xff})
 	si := &SigningInstruction{
 		AssetAmount: bc.AssetAmount{
-			AssetID: bc.NewAssetID([32]byte{0xff}),
+			AssetId: &assetID,
```
bobg: Yeah, I went back and forth on this. I'm similarly unsure that where I ended up is right, but as you say we can revisit this later.

protocol/bc/bc.proto
```diff
@@ -20,3 +25,148 @@ message AssetID {
...
+    Hash          ext_hash    = 6;
+  }
+  Body body = 1;
+}
+
+message Mux {
+  message Body {
+    repeated ValueSource sources  = 1; // issuances, spends, and muxes
+    Program              program  = 2;
+    Hash                 ext_hash = 3;
```
bobg: Mostly by hand, partly by Emacs. It would be nice (and wouldn't be hard) to write a `protofmt` tool, now that you mention it.

protocol/validation/validation.go
```diff
@@ -145,133 +147,154 @@ func checkValid(vs *validationState, e bc.Entry) error {
...
 			return errNonemptyExtHash
 		}
 
 	case *bc.Spend:
-		err := vm.Verify(NewTxVMContext(vs.tx, e, e.SpentOutput.Body.ControlProgram, e.Witness.Arguments))
+		spentOutputEntry, ok := vs.tx.Entries[*e.Body.SpentOutputId]
+		if !ok {
+			return errors.Wrapf(errMissingEntry, "entry for prevout %x not found", e.Body.SpentOutputId.Bytes())
+		}
+		spentOutput, ok := spentOutputEntry.(*bc.Output)
```
bobg: What do you think of https://github.com/chain/chain/pull/887/commits/adc6dd38801fe507a1daccb19bcecf9c4da100ca ?

kr: Hmm, either way.

PR #892, opened 2017-04-07
docker/ci: install Thales CodeSafe SDK

kr:
This will allow us to test our production build targets.

Future work: update the test harness in chainprv to
check production builds.


kr:
PTAL

One drawback to this is it roughly triples the size of our docker image. Maybe we can prune some of it out once we're using it for real tests.

boymanjor:
LGTM

Issue #891, opened 2017-04-07
move various wercker scripts into chain/bin


kr:
Where possible, scripts such as `core-tests` should go in `$CHAIN/bin` so we don't have to rebuild the whole docker image just to edit the wercker logic.

boymanjor:
Wdyt of leaving them in `docker/ci/bin` and adding that directory to the image `PATH`? We could introduce a subdirectory named `install` that could house the scripts we need for the image build. In the Dockerfile we'll only include this new directory. That way we don't pollute $CHAIN/bin with wercker specific scripts. We won't have to rebuild the image unless we update the install scripts.

boymanjor:
Fixed by #893

Issue #890, opened 2017-04-07
don't compile protoc


kr:
Our docker image for CI is compiling protoc from source, which takes an ungodly amount of time. But they have binary releases we can download.

https://github.com/google/protobuf/releases

Let's look in to that.

kr:
One possible reason: maybe they don't run on CentOS 6 (glibc 2.12).

boymanjor:
The official protobuf binaries are built with an incompatible version of glibc. What do you think of pulling the compiled binary out of the image and throwing it up on s3? We can add downloading it as a part of the image build.

bobg:
Worth mentioning: although generated Go code in proto3 is not supposed to contain getters for simple scalar fields (as documented [here](https://developers.google.com/protocol-buffers/docs/reference/go-generated#singular-scalar-proto3)), [a recent change](https://github.com/golang/protobuf/commit/da116c3771bf4a398a43f44e069195ef1c9688ef) to `protoc` re-added them. The resulting difference between `go generate` output on my computer vs. that on Wercker had me [briefly confounded](https://github.com/chain/chain/pull/850#issuecomment-291939082).

Whichever version of `protoc` we settle on for CI, let's make sure it's the same one we use for development.

boymanjor:
Fixed by #893

PR #893, opened 2017-04-07
docker/ci: optimize image build times

boymanjor:
This commit addresses #890 and #891. Previously, the image
building process involved compiling the protobuf compiler from
src. This takes a very long time, but was necessary due to glibc
incompatibilities between the official binaries and CentOS 6.

The artifacts of the compilation have been extracted from the
image and placed on s3 for download. This commit also separates
the image's install scripts from its test scripts and only loads the
former into the Dockerfile. This will allow updates to tests without
having to rebuild the image.


boymanjor:
PTAL

bobg:
LGTM, awesome

PR #894, opened 2017-04-07
cmd/slashland: support new GitHub merge API

kr:
GitHub removed support for the squash property in merge
requests. They replaced it with a new property called
merge_method, which is an enum rather than a bool.

See https://developer.github.com/changes/2017-04-07-end-merge-methods-api-preview/.


kr:
PTAL

boymanjor:
lgtm

PR #896, opened 2017-04-07
cmd/cored: introduce loopback_auth build tag

boymanjor:
By default, cored will authenticate requests made on the loopback
interface. To allow unauthenticated requests on this interface cored
should be built with the `loopback_auth` build tag. This is a part
of the work started in #859.

### DIFF
--- a/bin/build-cored-release
+++ b/bin/build-cored-release
@@ -48,7 +48,7 @@ DATE=${DATE:-`date +%s`} # date can be set via envvar
 ldflags="-X main.buildTag=$releaseRef -X main.buildCommit=$commit -X main.buildDate=$DATE"
 
 go build\
-  -tags 'plain_http unauthn_loopback'\
+  -tags 'plain_http no_loopback_authn'\
   -ldflags "$ldflags"\
   -o $outputDir/cored\
   chain/cmd/cored
--- a/bin/build-cored-release
+++ b/bin/build-cored-release
@@ -1,4 +1,7 @@
 #!/bin/bash
+#
+# Note: This script is intended for developer edition builds.
+# The build tags specified are not safe for production servers.
 set -e
 
 usage() {
--- a/bin/build-cored-release
+++ b/bin/build-cored-release
@@ -48,7 +48,7 @@ DATE=${DATE:-`date +%s`} # date can be set via envvar
 ldflags="-X main.buildTag=$releaseRef -X main.buildCommit=$commit -X main.buildDate=$DATE"
 
 go build\
-  -tags 'plain_http no_loopback_auth'\
+  -tags 'plain_http loopback_auth'\
   -ldflags "$ldflags"\
   -o $outputDir/cored\
   chain/cmd/cored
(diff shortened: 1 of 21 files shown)


boymanjor:
screen shot 2017-04-10 at 2 22 43 pm

boymanjor:
PTAL

cmd/cored/prod.go
```diff
@@ -15,6 +14,10 @@ var prod = true
 
 func resetInDevIfRequested(db pg.DB) {}
 
-func authLoopbackInDev(req *http.Request) bool {
-	return false
+func devEnableMockHSM(_ pg.DB) []core.RunOption {
+	return nil
+}
+
+func devHSM(_ pg.DB) (blocksigner.Signer, error) {
```
kr: I think these are left over from another rebase?

boymanjor: Yep, thanks.

boymanjor:
Updated, PTAL.

cmd/cored/no_loopback_authn.go
```diff
@@ -0,0 +1,18 @@
+//+build no_loopback_authn
```
kr: Normally I am all on board with using "authn" or "authz" to distinguish. However, in this case @tessr will shortly be doing the authz project and as part of that I'm pretty sure this flag will change from modifying authn behavior to modifying authz behavior.

We could update the tag name at that time… or we could cheat and just call it "auth" now. :stare:

cmd/deploy/main.go
```diff
@@ -87,7 +87,7 @@ func mustBuild(filename string) []byte {
 	commit = bytes.TrimSpace(commit)
 	date := time.Now().UTC().Format(time.RFC3339)
 	cmd = exec.Command("go", "build",
-		"-tags", "plain_http",
+		`-tags "plain_http no_loopback_authn"`,
```
kr: Ah, same thing about the quote marks here.

kr:
Review'd

cmd/cored/no_loopback_authn.go
boymanjor: I'm fine with using "auth" now.

boymanjor:
I updated this to use "auth" instead of "authn" and fixed the tags for `benchcore` and `deploy`. PTAL

boymanjor:
PTAL

bin/build-cored-release
```diff
@@ -48,7 +48,7 @@ DATE=${DATE:-`date +%s`} # date can be set via envvar
 ldflags="-X main.buildTag=$releaseRef -X main.buildCommit=$commit -X main.buildDate=$DATE"
 
 go build\
-  -tags 'plain_http'\
+  -tags 'plain_http loopback_auth'\
```
kr: Do we want this tag here? If this is for a production server, we should leave it out right?

boymanjor: I thought we were only using this script for developer edition builds. If not you're correct. @jeffomatic or @dominic any insight?

boymanjor: Discussed out of band: the product folks are using this for developer edition builds. A search of the codebase shows it only being used for this purpose as well. Should I add a comment about the flags here?

kr: Nah, I think it's good as-is. We are prob going to be reworking the release tools soon anyway.

PR #897, opened 2017-04-07
cmd/cored: introduce no_mockhsm build tag

boymanjor:
By default, the mockhsm will be enabled. To disable its usage,
the cored binary should be built with the `no_mockhsm` flag.
This is a part of the work started in #859.


boymanjor:
screen shot 2017-04-10 at 1 38 09 pm

screen shot 2017-04-10 at 1 37 38 pm

boymanjor:
PTAL

kr:
:eyes:

cmd/cored/main.go
```diff
@@ -73,6 +74,17 @@ var (
 
 	race          []interface{} // initialized in race.go
 	httpsRedirect = true        // initialized in plain_http.go
+
+	// By default, the mock HSM is enabled for use.
+	// This feature is turned off for production binaries
+	// with the no_mockshm build tag.
+	enableMockHSM = func(db pg.DB) []core.RunOption {
+		return []core.RunOption{core.MockHSM(mockhsm.New(db))}
```
kr: This is a great idea but I think we should still try to avoid linking the mockhsm package in this mode, so we probably need to back to two files, `mockhsm.go` and `no_mockhsm.go`. :/

And in that case, they can go back to ordinary function definitions.

boymanjor:
Updated, PTAL.

dashboard/src/features/app/components/Navigation/Navigation.jsx
```diff
@@ -56,7 +56,7 @@ class Navigation extends React.Component {
 
         <ul className={styles.navigation}>
           <li className={styles.navigationTitle}>services</li>
-          {!this.props.production &&
+          {!this.props.production && this.props.mockhsm &&
```
kr: Should we get rid of the production flag here? I would think now that there's a mockhsm flag that's the only thing that's relevant. Thoughts?

kr:
LGTM but I had a question up there

dashboard/src/features/app/components/Navigation/Navigation.jsx
boymanjor: I was trying to touch as little as necessary, but I think it makes sense to remove it now.

PR #898, opened 2017-04-08
all: support client cert authentication

boymanjor:
Support for client cert authentication has been implemented
in cored and all SDKs.  If a valid client cert is detected, basic
auth is bypassed. File paths to pem encoded X.509 certs and
keys are expected to be passed as env vars.

Initially the server will not require client certs and only verify
certs received from the client.


boymanjor:
PTAL

PR #899, opened 2017-04-08
cmd/cored/pkix: autogenerate development X.509 certs

boymanjor:
For internal development, we intend to require TLS for all HTTP
communication. This commit is in preparation for that goal. At
startup, if the core is configured to redirect HTTP requests to
HTTPS and no TLS config variables have been set, a self signed
CA will be generated. This CA will be used to issue a pair of X509
key pairs for server and client authn.

These certs will be used to continue initialization of the core.
Directions will also be printed for installing the root CA and client
certs into the user's local key store. These directions are initially
for macOS only. We only intend to require TLS for internal
development and this is the common OS of our team. Developer
edition builds will continue to allow plain HTTP connections.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -136,6 +136,15 @@ func runServer() {
 	ctx := context.Background()
 	env.Parse()
 
+	if *tlsCrt == "" && *tlsKey == "" && *rootCAs == "" {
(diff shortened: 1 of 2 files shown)


boymanjor:
PTAL

PR #903, opened 2017-04-10
protocol/bc: replace AssetID with a protobuf type

bobg:
The protobuf definition of `AssetID` is identical to that of `Hash`, allowing conversion between the two. `AssetID` gets much of its behavior from conversion to the `Hash` type. (See the discussion in https://github.com/chain/chain/pull/850.)


bobg:
PTAL

bobg:
PTAL

kr:
:eyes:

kr:
Need to `git rm hash.proto`? Otherwise LGTM

PR #904, opened 2017-04-10
database/sql: propagate context to standard lib

jbowens:
Propagate the context from our own database/sql package to the standard
library database/sql package.

For now, we still need our own database/sql wrapper for logging. We
might be able to replace it with a driver that performs the logging, but
composing drivers is tricky.


kr:
LGTM

PR #905, opened 2017-04-10
protocol/validation: add test of expansion flag to checkoutput

bobg:
The spec for `CHECKOUTPUT` says that when matching a retirement entry, `code` must be the empty string only if the expansion flag is false. The implementation was missing that check.


bobg:
PTAL

kr:
LGTM

PR #908, opened 2017-04-11
core/hsm: adhere to no_mockhsm tag

boymanjor:
This file should also be excluded when cored
is built using the `no_mockhsm` tag.


boymanjor:
PTAL

kr:
LGTM

PR #910, opened 2017-04-11
protocol/vm: add unit tests of checkoutput retirement cases

bobg:
Follow-on to https://github.com/chain/chain/pull/905.


bobg:
PTAL

kr:
I think these cases are mostly testing the behavior of the CheckOutput, so I would say they belong in package validation. For the ones that actually test code in package vm, maybe it would be easier to add them after converting these tests to stop using bc objects?

kr:
See also chain/protocol/validation/vmcontext_test.go.

PR #911, opened 2017-04-11
core/config: decode block pubkey in migration

jbowens:
Decode the signer public key before storing it in the Config protobuf.

```
app=cored buildtag=? processID=chain-mba.local-93848-76aec3fe66f6fd4ca3a8 at=block.go:142 t=2017-04-11T01:00:54.596164634Z error="key not found: misconfigured signer public key" signer="signer for key 32386463306330623539643434643164326565373063393838353231646164646632616138616364643634656263636138653932663139363062613066613833"
/Users/jackson/src/chain/errors/errors.go:185 - chain/errors.Sub
/Users/jackson/src/chain/core/blocksigner/blocksigner.go:56 - chain/core/blocksigner.(*BlockSigner).SignBlock
/Users/jackson/src/chain/core/generator/block.go:140 - chain/core/generator.getSig
/usr/local/go/src/runtime/asm_amd64.s:2197 - runtime.goexit
app=cored buildtag=? processID=chain-mba.local-93848-76aec3fe66f6fd4ca3a8 at=generator.go:131 t=2017-04-11T01:00:54.596256881Z error="sign: got 0 of 1 needed signatures"
```


bobg:
Maybe this isn't the PR for it, but I notice that `config.proto` stores one pubkey as a `string` (in `Config`) and one as a `bytes` (in `BlockSigner`). It would be better if we were consistent about this, and maybe pubkeys should get the `bc.Hash` treatment - protobuf message types of their own, with their own `Scan` and `Value` methods, etc.

kr:
+1 for being consistent (my first instinct would be to use `bytes` for both)

jbowens:
Updated to use `bytes` for the block pubkey in the proto. Heads up that this will break any local config protos. Please take a second look :)

I was a little unsure about the corectl flags. They should take the pubkey hex-encoded, yah?

tessr:
LGTM. I definitely forgot to test passing the pubkey to the generator when configuring it using `corectl`--your change sounds right to me.

PR #912, opened 2017-04-11
core: add ability to update account tags

jeffomatic:


jeffomatic:
PTAL, ran it using batch and singleton variants of this script: https://gist.github.com/jeffomatic/49e8f6a68e500009122d700226936d40

kr:
:eyes:

core/account/accounts.go
```diff
@@ -128,6 +131,67 @@ func (m *Manager) Create(ctx context.Context, xpubs []chainkd.XPub, quorum int,
...
+		}
+		alias = &a
+	} else {
+		signer, err = m.FindByAlias(ctx, *alias)
+		if err != nil {
+			return errors.Wrap(err, "get account by alias")
+		}
+	}
+
+	if err != nil {
```
kr: This seems to be superfluous?

jeffomatic: Yes, it's a legacy of a previous iteration. Nice catch!

PR #914, opened 2017-04-12
bin/wait-and-configure: new utility script

jeffomatic:
`wait-and-configure` waits until a local Chain Core daemon is ready, and
then attempts to configure it. You can use this script as part of a
quick reset-and-configure loop while iterating on code. For example,
some of our teammates use a Make target that executes the equivalent of
the following:

```
  go install $(CHAIN)/cmd/corectl
  go install $(REPO)/cmd/cored
  -dropdb core
  createdb core
  rm -rf ~/Library/Application\ Support/Chain\ Core/raft
  bin/wait-and-configure &
  cored
```

Previously, `corectl config-generator` could be used in place of `wait-and-configure` for fast iteration loops, but since it now requires a running instance of Chain Core, we need something that can run in the background.

### DIFF
--- a/bin/wait-and-configure
+++ b/bin/wait-and-configure
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+# TODO: move this wait functionality to corectl as a -wait flag.
 while ! `curl -s http://localhost:1999 | grep -q Found`
 do
   sleep 0.5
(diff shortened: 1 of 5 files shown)


jeffomatic:
PTAL

kr:
Should this be built in to corectl? `corectl config-generator -wait &`

Just a thought.

kr:
Anyway, LGTM

PR #915, opened 2017-04-12
core: add ability to update asset tags

jeffomatic:


jeffomatic:
PTAL, tested manually using this script: https://gist.github.com/jeffomatic/717f68f35e73f2910f26eb4a60eae54c

core/query/assets.go
```diff
@@ -29,6 +30,12 @@ func (ind *Indexer) SaveAnnotatedAsset(ctx context.Context, asset *AnnotatedAsse
 	return errors.Wrap(err, "saving annotated asset")
 }
 
+func (ind *Indexer) UpdateAnnotatedAssetTags(ctx context.Context, id bc.AssetID, tags json.RawMessage) error {
+	const q = "UPDATE annotated_assets SET tags = $1::jsonb WHERE id = $2"
+	_, err := ind.db.Exec(ctx, q, string(tags), id)
+	return errors.Wrap(err, "update query")
+}
```
jbowens: Can we use `SaveAnnotatedAsset` instead since `SaveAnnotatedAsset` is designed for an upsert?

jbowens: Should be able to update the `assets` table, call `Annotated(a)` to get the annotated asset and then save it with `SaveAnnotatedAsset`.

jeffomatic:
PTAL @jbowens, revised based on your suggestion.

I also added some cache updating...what's there now is the simple route. I think there is some opportunity for subtle drift between the cache and postgres unless I keep the cache locked during the postgres updates. Do you think this is a big deal?

jbowens:
lgtm

PR #917, opened 2017-04-12
core/config: set ConfiguredAt

boymanjor:
Before using raft to store core config, this value was directly
inserted into the config table. This fixes a bug causing this
value to default to 0.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -253,6 +253,7 @@ func Configure(ctx context.Context, db pg.DB, rDB *raft.Service, c *Config) erro
 		return errors.Wrap(err)
 	}
 	c.Id = hex.EncodeToString(b)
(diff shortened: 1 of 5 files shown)


boymanjor:
PTAL

jbowens:
I wonder if we should rename the field to `configured_at_ms` to match `max_issuance_window_ms` sometime?

boymanjor:
I agree. I'll do that in a follow up.

kr:
There's also https://github.com/google/protobuf/blob/master/src/google/protobuf/timestamp.proto that we might want to use across the board. Only blockchain times are required to be milliseconds.

kr:
err, maybe https://godoc.org/github.com/golang/protobuf/ptypes is the way to access it, not sure

PR #919, opened 2017-04-12
protocol/bc/legacy: new package for legacy types

bobg:
This PR is strictly[*] a mechanical move of legacy (non-txentries) types, functions, etc. to a new package. The entire `legacy` package will be removed eventually.

The sequence of events after this PR:

- Improve the naming of objects in `protocol/bc` now that conflicts with legacy names is not an issue
- Finalize the new `Tx` and `Block` data structures, including their (new) serialization
- For every calling context that serializes or deserializes a legacy object (RPC handlers, db queries, etc) add a parallel construction (new RPC endpoint, new db column, etc) that does it with the corresponding new object and its new serialization

The version of Chain Core at this point (A) can read both old and new format data, in theory. :crossed_fingers:

- Update callsites using the `legacy` package to use the corresponding updated logic from `bc`. Add that logic where it doesn't yet exist (e.g., building txentries-style transactions without mapping from a legacy-style transaction)
- Update tests

The version of Chain Core at this point can still read old data and respond to old RPC requests but will only ever write new data and make new RPC calls. It is safe to deploy wherever we're sure version (A) above is fully deployed.

- Write a migration tool to convert any old db data to the new format

Users who are comfortable with having a [flag day](https://en.wikipedia.org/wiki/Flag_day_(computing)) can run this migration in preparation for the future version of Chain Core (assuming there is one) that removes the `legacy` package and can therefore no longer read old-format data. Note: with a little extra effort we can avoid the need for a flag day if necessary, allowing the migration to run incrementally on a live database.

[*] - I took the liberty of removing the `NewBlockVersion` and `CurrentTransactionVersion` constants rather than move them.


bobg:
PTAL

bobg:
PTAL

bobg:
PTAL

bobg:
rebased

erykwalder:
LGTM,

one thought about `protocol/bc/legacy/blockheader.go`, should the filename be `block_header.go` to match `block_witness.go` and `block_commitment.go`?

PR #920, opened 2017-04-12
core: move authn and authz into their own packages

tessr:
This parses authn tokens from http requests and adds them to the request contexts. It also checks if the request came from a loopback address, and adds a flag to the request context for that as well. 

This is groundwork for the forthcoming authz work.


core/authn.go
```diff
@@ -34,26 +36,56 @@ type tokenResult struct {
...
 func (a *apiAuthn) handler(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		err := a.auth(req)
-		if err != nil {
-			errorFormatter.Write(req.Context(), rw, err)
+		token, err0 := a.tokenAuth(req)
+		ctx := req.Context()
+		if err0 == nil {
+			// if this request was successfully authenticated, pass the token along
+			if token != "" {
```
bobg: Suggest combining this `if` with the one above (via `&&`) to avoid unnecessary `if` nesting.

kr:
:eyes:

bobg:
LGTM with one suggestion.

core/authn.go
```diff
@@ -34,26 +36,54 @@ type tokenResult struct {
...
+
+func (a *apiAuthn) tokenAuth(req *http.Request) (string, error) {
 	user, pw, ok := req.BasicAuth()
+	// TODO(tessr): remove the following clause once ACLs are in place
 	if !ok && a.alt != nil && a.alt(req) {
-		return nil
+		return "", nil
 	}
 
+	// Is this the way we want to encode the token to be passed around? Or just user?
```
kr: Just user. I think a reasonable conceptual framework is that the "credentials" prove your "identity", and the authn handler checks the credentials but everything else works just in terms of identity, relying on authn to have done its thing.

So an access token e.g. `kr:abc123` would prove my identity as `user("kr")`.

Or a remote addr of `127.0.0.1:12345` would prove my identity as `localhost(true)`.

core/authn.go
```diff
@@ -34,26 +36,54 @@ type tokenResult struct {
...
 func (a *apiAuthn) handler(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		err := a.auth(req)
-		if err != nil {
-			errorFormatter.Write(req.Context(), rw, err)
+		token, err0 := a.tokenAuth(req)
+		ctx := req.Context()
+		if err0 == nil && token != "" {
+			// if this request was successfully authenticated, pass the token along
+			ctx = context.WithValue(ctx, "token", token)
```
kr: Check out the sample code in https://godoc.org/context#Context (on the Value method) and https://godoc.org/context#WithValue. I think we should try to follow that pattern: making an unexported type just for the context key guarantees it won't conflict with any other package.

So here it could be something like

```go
type authnKey int

var (
	tokenKey = iota
	localhostKey
)
```

and this line would be

```go
ctx = context.WithValue(ctx, tokenKey, token)
```

core/authn.go
```diff
@@ -34,26 +36,54 @@ type tokenResult struct {
...
+		}
+
+		err1 := a.localhostAuth(req)
```
kr: Instead of duplicating some of the logic from cmd/cored here, I think it would be better to delete it from cmd/cored entirely and rely just on this one.

core/authn.go
```diff
@@ -34,26 +36,54 @@ type tokenResult struct {
...
+		if err1 == nil {
+			ctx = context.WithValue(ctx, "localhost", true)
+		}
+
+		// TODO(tessr): move this to authz as part of ACL work
```
kr: I think it would be easy and really nice to actually do this TODO here. This PR can be a pure refactoring, with the goal of splitting apart authn and authz into two separate http handlers.

The existing logic (spread across chain/cmd/cored and chain/core) is:

- when compiled with `loopback_auth`:
   - if there's a valid access token, ok
   - if the connection is from localhost, ok
   - otherwise reject
- when compiled without `loopback_auth`:
   - if there's a valid access token, ok
   - otherwise reject

This would translate pretty directly to a trivial, hard-coded authz policy here:

```go
//+build loopback_auth
func authz(ctx context.Context) bool {
	return authnAccessToken(ctx) != "" || authnLocalhost(ctx)
}

//+build !loopback_auth
func authz(ctx context.Context) bool {
	return authnAccessToken(ctx) != ""
}
```

(I am assuming haveFoo is a helper function that basically just calls ctx.Value.)

So if there are two handlers, the authz one would just call this `authz` function above, and the authn one would look a lot like what's here currently, except it wouldn't do this check here at the end.

This would separate authn from authz, so that future PRs could operate entirely in authz land. It would also mean we can delete the `alt` hack immediately.

core/authn.go
tonychain: So there's a couple schools of thought here. One is what you're talking about, a.k.a. [claims-based identity](https://en.wikipedia.org/wiki/Claims-based_identity).

But there's also an entirely different school of thought around authorization-centric credentials which include a sort of floating policy in the credential themselves.

core/authn.go
tonychain: In the ACL library I wrote for Square, we hoisted all of the credentials out of the context/environment into a specific "credential" object (which otherwise lived in the context/environment):

https://github.com/square/rails-auth/blob/master/lib/rails/auth/x509/middleware.rb#L43

tonychain: Oh whoops, so I see the `CredentialStore` now. If I have one other suggestion it would be to separate it by credential types, and (perhaps in a separate PR) also toss the X.509 client certificate used in the request into there (if, for no other reason, so it's easy to introspect)

It's nice having all the different credential types in one place because sometimes you want to use them in combination (e.g. X.509 + token for [Token Binding](https://datatracker.ietf.org/wg/tokbind/charter/))

tessr:
@kr I think this is now more in line with the refactoring we were talking about yesterday. It doesn't implement any of @tonychain's suggestions about credential objects, but I think we can do that (and/or discuss it) in another PR, perhaps after the policy stuff gets a little more fleshed out. PTAL

kr:
:eyes:

cmd/cored/loopback_auth.go
```diff
@@ -2,17 +2,8 @@
...
-import (
-	"chain/core/config"
-	"net"
-	"net/http"
-)
+import "chain/core/config"
 
 func init() {
 	config.BuildConfig.LoopbackAuth = true
-	loopbackAuth = func(req *http.Request) bool {
```
kr: Do you want to drop a comment in here pointing the reader to where the logic is? Like

```go
// See $CHAIN/net/authz/loopback_authz.go for the implementation.
```

core/api.go
```diff
@@ -243,6 +243,28 @@ type page struct {
 	LastPage bool         `json:"last_page"`
 }
 
+func (a *API) authnHandler(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		req = (&authn.API{
+			Tokens:           a.accessTokens,
+			TokenMap:         make(map[string]authn.TokenResult),
```
kr: It looks like this is making a new map for every request, which would defeat the purpose of putting tokens in this map as a cache. Maybe pull this object out of the closure just before line 247?

It might also be a little clearer to define a constructor in package authn:

```go
func (a *API) authnHandler(handler http.Handler) http.Handler {
	auth := authn.NewAPI(a.accessTokens, networkPrefix)
		req = auth.Authenticate(req)
		handler.ServeHTTP(rw, req)
	})
}
```

Doing this might also allow us to un-export authn.TokenResult???

core/api.go
```diff
@@ -243,6 +243,28 @@ type page struct {
...
+
+		handler.ServeHTTP(rw, req)
+	})
+}
+
+func authzHandler(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		if authz.Authorized(req.Context()) {
+			handler.ServeHTTP(rw, req)
+		} else {
```
kr: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

This suggests this should look like:

```go
if !authz.Authorized(req.Context()) {
	errorFormatter.Write(req.Context(), rw, errNotAuthenticated)
	return
}
handler.ServeHTTP(rw, req)
```

core/api.go
```diff
@@ -243,6 +243,28 @@ type page struct {
...
+			errorFormatter.Write(req.Context(), rw, errNotAuthenticated) // perhaps should have different error?
```
kr: Yeah good call, we should prob also have an `errNotAuthorized` I guess.

net/authz/validators.go
```diff
@@ -0,0 +1,47 @@
...
+	return LocalhostFromContext(ctx)
+}
+
+// NewContextWithToken sets the token in a new context and returns the context.
+func NewContextWithToken(ctx context.Context, token string) context.Context {
+	return context.WithValue(ctx, tokenKey, token)
+}
+
+// TokenFromContext returns the token stored in the context, if there is one.
+func TokenFromContext(ctx context.Context) (string, bool) {
```
kr: I'd say these accessors and the `key` type belong in package authn, and authz should depend on authn rather than the other way around. Your user name (your identity) is part of and defined by authentication, it is used by authorization.

I think this would also mean the NewContextFoo functions can be un-exported inside of authn.

As a guess, I bet that having the dependency go from authn->authz is to avoid a dependency from something outside of http on something inside of that directory? I'd say maybe let's not worry about it for now or just move this package in there as well. I can think of a bunch of tricks we might do to pull more code out of that subdirectory but not sure if it's worth it rn.

net/http/authn/authn.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"chain/core/accesstoken"
+	"chain/errors"
+	"chain/net/authz"
+)
+
+const tokenExpiry = time.Minute * 5
+
+type API struct {
+	Tokens           *accesstoken.CredentialStore
```
kr: Putting this concrete type here takes a hard dependency from chain/net/http/authn on chain/core/accesstoken. We shouldn't do that — as a "general-purpose" library, packages outside of chain/core/... shouldn't depend on core packages.

Prob the right thing to do is define a one-method interface here that can be satisfied by *accesstoken.CredentialStore.

Maybe also, as a potential TODO item, is to take all the code in that package (chain/core/accesstoken) and move it all in here.

core/api.go
tessr: Ha, oops. And yeah, I think the constructor will be very helpful.

net/http/authn/authn.go
tessr: How would you feel if I only added that TODO? I feel like moving the CredentialStore code is the better solution, but I'm not sure I want to move so much code right now.

tessr:
Alright, I've made the changes we discussed (except for breaking the `accesstoken` dependency--added a todo). PTAL!

net/http/authn/context.go
```diff
@@ -0,0 +1,36 @@
...
+	localhostKey
+)
+
+// newContextWithToken sets the token in a new context and returns the context.
+func newContextWithToken(ctx context.Context, token string) context.Context {
+	return context.WithValue(ctx, tokenKey, token)
+}
+
+// TokenFromContext returns the token stored in the context, if there is one.
+func TokenFromContext(ctx context.Context) (string, bool) {
```
kr: 🚲 🏠:

- Should these accessors should be named `Token` and `Localhost`?
- Should this function return just a `string` (the empty string if there was no valid token)?

Then the usage might look like

```go
func authzToken(ctx context.Context) bool {
	// TODO(tessr): compare against Policies
	return authn.Token(ctx) != ""
}
```

kr:
LGTM with some faint suggestions

net/http/authn/context.go
tessr: I like the color of this shed.

PR #922, opened 2017-04-13
analytics: add indexing of custom columns

jbowens:
Add ability to index custom columns extracted from JSON reference data.

This does not yet handle configuration of these custom columns. I'm not
sure what the best interface for configuration is?

Some of the logic to follow json paths got a little gross. I was trying to avoid
unmarshaling the entire transaction object a second time into a 
`Map<String,Object>`, hence the `JsonPath.extract` methods for `Transaction`,
`Transaction.Input` and `Transaction.Output`.


erykwalder:
:eyes:

erykwalder:
LGTM

Issue #924, opened 2017-04-13
Bug: submitting transaction with unbalanced outputs/inputs returns CH000


jeffomatic:
On recent builds of `main`, submitting a transaction with unbalanced inputs and outputs will yield `CH000` instead of `CH735` (Transaction rejected). For example, the following transaction will fail:

- issue one unit of `gold`
- control 100 units of `gold` in `alice`'s account

This should be considered a userland error rather than a server error.

As a side note, this causes the Node SDK integration tests to fail.

bobg:
We lost the `ErrBadTx` umbrella error that previously was wrapping validation errors (and that is then converted to a `CH735`). Will fix.

jeffomatic:
Thanks @bobg !

PR #925, opened 2017-04-13
protocol: wrap validation errors in an ErrBadTx

bobg:
In https://github.com/chain/chain/pull/788 we lost logic that wrapped validation errors in `ErrBadTx`, which `core/txbuilder` was expecting in order to produce a friendly "transaction rejected" message for the caller. This PR restores that and adds a regression test.

Fixes https://github.com/chain/chain/issues/924.

### DIFF
--- a/core/txbuilder/finalize_test.go
+++ b/core/txbuilder/finalize_test.go
@@ -20,6 +20,7 @@ import (
 	"chain/protocol/prottest"
 	"chain/protocol/prottest/memstore"
 	"chain/protocol/state"
+	"chain/protocol/vm"
 	"chain/testutil"
 )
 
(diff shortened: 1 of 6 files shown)


bobg:
PTAL

core/txbuilder/finalize_test.go
```diff
@@ -162,6 +163,49 @@ func TestConflictingTxsInPool(t *testing.T) {
...
+			},
+		},
+		Outputs: []*bc.TxOutput{
+			bc.NewTxOutput(bc.AssetID{}, 2, nil, nil),
+		},
+		MinTime: 1,
+		MaxTime: 2,
+	})
+	err := FinalizeTx(ctx, c, nil, badTx)
+	if errors.Root(err) != ErrRejected {
```
jeffomatic: Should this be asserting that `ErrBadTx` is returned? I think that's what will cause the appropriate user error.

bobg: The conversion of `ErrBadTx` to `ErrRejected` happens in `FinalizeTx`.

jeffomatic: Understood...should the error message on the following line be more like this?

```
t.Errorf("got error %s, want %s", errors.Root(err), ErrRejected)
```

bobg: Oops, good catch, thanks.

jeffomatic:
LGTM

PR #927, opened 2017-04-13
core/query: add `block_transactions_count` to tx

### LINKED ISSUE #921: API: add block_transactions_count transaction annotation
This is a stretch goal for 1.2. If it doesn't make it out in that release, it can go into a subsequent minor-version release, i.e. 1.3.

We'd like to add a `block_transactions_count` field to the annotated transaction object. This would provide some context when iterating over ascending-order lists of transactions, such as in transaction feeds.

In particular, you could perform actions upon reaching the last transaction in a block, and you could do so without knowledge of the subsequent transaction in the list. This is nice, since you cannot have complete knowledge of blockchain state at a given block height without having examined all transactions in the block.

Implementation is straightforward--the block-related annotations in a transaction would be as follows:

- `block_height`: height of the block
- `block_id`: hash of the block
- `timestamp`: RFC3339 serialization of block time
- `position`: position of transaction in the containing block
- `block_transactions_count` (NEW): the total number of transactions in the block

#### Questions

How will this affect existing annotated transactions? Is there a simple migration that can supplement existing annotations with this value?

jbowens:
Add a `block_transactions_count` to annotated transactions.

If we wanted to backfill, we would need to not only backfill the `block_tx_count` column but also add the element to the json in the `data` column. Couple options:

1. Write a Go migration to backfill and then SET NOT NULL on the `block_tx_count` column.
2. If we required PostgreSQL 9.5.1, we could write some intense SQL to count the number of  annotated_txs at each block height and use the `||` operator to merge the new json element into the `data` column's json blob. This wouldn't respect the field ordering for the `block_transactions_count` field in backfilled annotated transaction.

Fix #921.


bobg:
:eyes:

bobg:
Not sure this addresses all of https://github.com/chain/chain/issues/921. In particular, whether to do a backfill appears to be an open question.

jbowens:
> Not sure this addresses all of #921. In particular, whether to do a backfill appears to be an open question.

Please see the PR description. IMO, neither option is worth it but not sure if there's additional context requiring a backfill.

bobg:
LGTM but maybe we shouldn't close https://github.com/chain/chain/issues/921 (or we should open a new issue for just the possible backfill).

kr:
In #921 backfill is just a "nice to have" idea in a footnote. I think it's fine to close #921 with this PR and not bother opening another issue unless there's actual demand for a backfill from somewhere.

PR #907, opened 2017-04-11
protocol/vm: create a Context directly

kr:
This change converts the tests in crypto_test.go to stop
using packages bc and validation to build test fixtures,
and instead define a Context object in one step. Code
coverage stays at 95.3%.

This is a continuation of #886.

Future work: once all test files stop depending on
package bc, move the test code back into package vm
(from vm_test).

PR #928, opened 2017-04-13
protocol/vm: create a Context directly

kr:
This change converts the tests in introspection_test.go
to stop using packages bc and validation to build test
fixtures, and instead define a Context object in one
step. Code coverage stays at 95.3%.

This completes the work in #886 and #907. Now no files
in chain/protocol/vm depend on packages validation or bc.

Future work: move the test code back into package vm
(from vm_test).


bobg:
LGTM

PR #929, opened 2017-04-13
crypto/ed25519/math: EC math primitives

bobg:
A more Go-idiomatic interface to the routines in `ed25519/internal/edwards25519` for doing point and scalar (mod L) arithmetic.

### DIFF
--- a/crypto/ed25519/math/point.go
+++ b/crypto/ed25519/math/point.go
@@ -1,0 +1,92 @@
+package math
+
+import (
+	"crypto/subtle"
+
+	"chain/crypto/ed25519/internal/edwards25519"
+)
+
+type (
+	// Point is a point on the ed25519 curve.
+	Point edwards25519.ExtendedGroupElement
+
+	// EncodedPoint is a serialized representation of a point.
+	EncodedPoint [32]byte
+)
+
+// ZeroPoint is the zero point on the ed25519 curve.
+var ZeroPoint Point
+
+// Add adds the points in x and y, storing the result in z and
+// returning that. Any or all of x, y, and z may be the same pointers.
+func (z *Point) Add(x, y *Point) *Point {
+	var y2 edwards25519.CachedGroupElement
+	(*edwards25519.ExtendedGroupElement)(y).ToCached(&y2)
+
+	var z2 edwards25519.CompletedGroupElement
+	edwards25519.GeAdd(&z2, (*edwards25519.ExtendedGroupElement)(x), &y2)
+
+	z2.ToExtended((*edwards25519.ExtendedGroupElement)(z))
+	return z
+}
+
+// Sub subtracts y from x, storing the result in z and
+// returning that. Any or all of x, y, and z may be the same pointers.
+func (z *Point) Sub(x, y *Point) *Point {
+	var y2 edwards25519.CachedGroupElement
+	(*edwards25519.ExtendedGroupElement)(y).ToCached(&y2)
+
+	var z2 edwards25519.CompletedGroupElement
+	edwards25519.GeSub(&z2, (*edwards25519.ExtendedGroupElement)(x), &y2)
+
+	z2.ToExtended((*edwards25519.ExtendedGroupElement)(z))
+	return z
+}
+
+// ScMul multiplies the EC point x by the scalar y, placing the result
+// in z and returning that. X and z may be the same pointer.
+func (z *Point) ScMul(x *Point, y *Uint256le) *Point {
+	return z.ScMulAdd(x, y, &Zero)
+}
+
+// ScMulBase multiplies the ed25519 base point by x and places the
+// result in z, returning that.
+func (z *Point) ScMulBase(x *Uint256le) *Point {
+	edwards25519.GeScalarMultBase((*edwards25519.ExtendedGroupElement)(z), (*[32]byte)(x))
+	return z
+}
+
+// ScMulAdd computes xa+yB, where B is the ed25519 base point, and
+// places the result in z, returning that.
+func (z *Point) ScMulAdd(a *Point, x, y *Uint256le) *Point {
+	// TODO: replace with constant-time implementation to avoid
+	// sidechannel attacks
+
+	var p edwards25519.ProjectiveGroupElement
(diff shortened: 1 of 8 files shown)


bobg:
Hey, remember this code? Shall we land this before the new chainkd implementation, and write that in terms of this?

PTAL

crypto/ed25519/math/point.go
```diff
@@ -0,0 +1,92 @@
...
+func (z *Point) ScMulAdd(a *Point, x, y *Uint256le) *Point {
+	// TODO: replace with constant-time implementation to avoid
+	// sidechannel attacks
+
+	var p edwards25519.ProjectiveGroupElement
+	edwards25519.GeDoubleScalarMultVartime(&p, (*[32]byte)(x), (*edwards25519.ExtendedGroupElement)(a), (*[32]byte)(y))
+
+	var buf [32]byte
+	p.ToBytes(&buf)
+	(*edwards25519.ExtendedGroupElement)(z).FromBytes(&buf)
```
bobg: It's OK to ignore the return value from `ExtendedGroupElement.FromBytes` here, right? (The `bool` indicating a legal value.)

crypto/ed25519/math/point.go
```diff
@@ -0,0 +1,92 @@
...
+type (
+	// Point is a point on the ed25519 curve.
+	Point edwards25519.ExtendedGroupElement
+
+	// EncodedPoint is a serialized representation of a point.
+	EncodedPoint [32]byte
+)
+
+// ZeroPoint is the zero point on the ed25519 curve.
+var ZeroPoint Point
```
kr: In the godoc this is going to look like ZeroPoint is the zero value of Point, which is misleading since it's initialized to something else in func init. Can we either initialize it in an assignment here or at least put a comment calling this out?

crypto/ed25519/math/point.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"chain/crypto/ed25519/internal/edwards25519"
+)
+
+type (
+	// Point is a point on the ed25519 curve.
+	Point edwards25519.ExtendedGroupElement
+
+	// EncodedPoint is a serialized representation of a point.
+	EncodedPoint [32]byte
```
kr: Hmm just thinking out loud… how much benefit do we get out of naming this type instead of using [32]byte directly?

crypto/ed25519/math/point.go
```diff
@@ -0,0 +1,92 @@
...
+	var p edwards25519.ProjectiveGroupElement
+	edwards25519.GeDoubleScalarMultVartime(&p, (*[32]byte)(x), (*edwards25519.ExtendedGroupElement)(a), (*[32]byte)(y))
+
+	var buf [32]byte
+	p.ToBytes(&buf)
+	(*edwards25519.ExtendedGroupElement)(z).FromBytes(&buf)
+	return z
+}
+
+func (z *Point) Encode() (e EncodedPoint) {
```
kr: Nit: it's better to leave out the return value names when they don't contribute information. https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

So I would recommend

```go
func (z *Point) Encode() EncodedPoint {
	var e EncodedPoint
	(*edwards25519.ExtendedGroupElement)(z).ToBytes((*[32]byte)(&e))
	return e
}
```

or

```go
func (z *Point) Encode() [32]byte {
	var e [32]byte
	(*edwards25519.ExtendedGroupElement)(z).ToBytes(&e)
	return e
}
```

crypto/ed25519/math/point.go
kr: Could be as direct as

```
var ZeroPoint Point = zeroPoint()
```

as long as there's *some* indication it has a non-"zero" value.

crypto/ed25519/math/point.go
kr: (I know I've done this the `(e EncodedPoint)` way before. I'm trying to break my habit lol.)

crypto/ed25519/math/point.go
tonychain: For what it's worth, curve25519-dalek has named types for everything (calling this one `CompressedEdwardsY`):

https://github.com/isislovecruft/curve25519-dalek/blob/master/src/curve.rs#L109

crypto/ed25519/math/point.go
bobg: 99.9% sure it's OK. Added a TODO for the other 0.1%.

crypto/ed25519/math/point.go
bobg: Removed `EncodedPoint` for now, but I suspect we might want to add it back - not for the sake of this package, which barely uses it, but because the CA code will be full of a few different kinds of `[32]byte` values, and it may make things clearer to name them distinctly.

oleganza:
@bobg great idea. If first thought, since ChainKD does not do a lot of ECC voodoo (unlike CA), I'd rather land it as-is, and then rewrite in a separate PR in terms of these primitives. But I'd like to get it reviewed more thoroughly before landing, so we can land this, and rewrite ChainKD in its current PR.

oleganza:
Also: i'd like to use these primitives for CArev2 implementation that i start, like, right now :-P

crypto/ed25519/math/point.go
```diff
@@ -0,0 +1,91 @@
...
+	// TODO(bobg): double-check that it's OK to ignore the return value
+	// from ExtendedGroupElement.FromBytes here. (It's a bool indicating
+	// that its input represented a legal value.)
+	(*edwards25519.ExtendedGroupElement)(z).FromBytes(&buf)
+	return z
+}
+
+func (z *Point) Encode() [32]byte {
+	var e [32]byte
+	(*edwards25519.ExtendedGroupElement)(z).ToBytes((*[32]byte)(&e))
```
kr: Nit: unnecessary type conversion here.

crypto/ed25519/math/point.go
```diff
@@ -0,0 +1,91 @@
...
+	return e
+}
+
+func (z *Point) Decode(e [32]byte) (*Point, bool) {
+	ok := (*edwards25519.ExtendedGroupElement)(z).FromBytes((*[32]byte)(&e))
```
kr: Same nit here.

kr:
LGTM with a nit

PR #931, opened 2017-04-13
sdk/node: add access control methods

dominic:


sdk/node/src/api/accessControl.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * More info: {@link https://chain.com/docs/core/learn-more/authentication}
+ * @typedef {Object} Grant
+ * @global
+ *
+ * @property {String} guard_type
+ * Type of credential, either 'access_token' or 'x509'.
+ *
+ * @property {Object} guard_data
+ * Object containing data needed to identify the incoming credential.
```
jeffomatic: Folks will probably need a bit more hinting here, e.g.:

```
Data used by the guard to identity incoming credentials.

If guard_type is "access_token", you should provide an instance of {AccessTokenGuardData}, which identifies access tokens by ID.

If guard_type is "x509", you should...
```

sdk/node/src/api/accessControl.js
```diff
@@ -0,0 +1,72 @@
...
+const accessControl = (client) => ({
+  /**
+   * Create a new access grant.
+   *
+   * @param {Object} params - Parameters for access grant creation.
+   * @param {String} params.guard_type - Type of credential to guard with, either 'access_token' or 'x509'.
+   * @param {Object} params.guard_data - Object containing data needed to identify the incoming credential.
+   * @param {String} params.policy - Authorization polciy to attach to specific grant. See {@link Grant} for a list of available policiies.
+   * @param {objectCallback} [callback] - Optional callback. Use instead of Promise return value as desired.
+   * @returns {Promise} Status of created object.
```
jeffomatic: Maybe s/Status of created object/Success message/

sdk/node/src/api/accessControl.js
```diff
@@ -0,0 +1,72 @@
...
+   */
+  create: (params , cb) =>
+    shared.create(client, '/create-acl-grant', params, {skipArray: true}, cb),
```
jeffomatic: I think `cb` should be a field in the options object. Have you run this code yet?

sdk/node/src/api/accessControl.js
```diff
@@ -0,0 +1,72 @@
...
+    cb
+  ),
+
+  /**
+   * Get all access grants.
+   *
+   * @param {pageCallback} [callback] - Optional callback. Use instead of Promise return value as desired.
+   * @returns {Promise<Array<Grant>>} Requested page of results.
+   */
+  query: (cb) =>
```
jeffomatic: Should there be a `queryAll` equivalent? Or did we get rid of pagination?

sdk/node/src/api/accessControl.js
dominic: We use this language elsewhere; should I do an everything update here?

sdk/node/src/api/accessControl.js
dominic: I've been running the promise version; good catch.

sdk/node/src/api/accessControl.js
dominic: Surprisingly, my vote is for no pagination right now.

sdk/node/src/api/accessControl.js
dominic: Actually, scratch that. I totally updated markdown in this branch too.

dominic:
@jeffomatic PTAL again.

re: integration tests, yes. There are no APIs landed in core yet. I'm happy to hold this back until we have them though.

dominic:
Feedback from Jeff: let's rename `query` to `list`.

dominic:

jeffomatic:
LGTM but would you mind waiting until we get tests?

sdk/node/src/api/mockHsmKeys.js
```diff
@@ -5,7 +5,7 @@ const shared = require('../shared')
   * Cryptographic private keys are the primary authorization mechanism on a
   * blockchain. For development environments, Chain Core provides a convenient
   * MockHSM.
-  * <br/><br/>
+  *
```
kr: It looks to me like almost 25% of this PR is unrelated edits like this one. It would be nice to keep each PR focused on one thing when possible.

dominic: Yup. In this case, changing https://github.com/chain/chain/pull/931/files#diff-84c1d078debc432e4577dbe04061249aR16 to support https://github.com/chain/chain/pull/931/files#diff-f138fcf4ae92f3121a6c5988001b5f6fR4 broke a bunch of formatting elsewhere. I can split it up easily enough if that's preferred.

PR #932, opened 2017-04-13
dashboard: ability to update existing asset/account tags

chrisgarvin:


jeffomatic:
I get this error when I run it:

screen shot 2017-04-13 at 5 31 46 pm

chrisgarvin:
@jeffomatic i think you may need to delete your node modules folder and npm install again

dashboard/src/features/accounts/components/AccountShow.jsx
```diff
@@ -51,7 +52,10 @@ class AccountShow extends BaseShow {
             actions={[
               <button key='show-txs' className='btn btn-link' onClick={this.props.showTransactions.bind(this, item)}>Transactions</button>,
               <button key='show-balances' className='btn btn-link' onClick={this.props.showBalances.bind(this, item)}>Balances</button>,
-              <RawJsonButton key='raw-json' item={item} />
+              <RawJsonButton key='raw-json' item={item} />,
+              <Link to={`/accounts/${item.id}/tags`} key='edit-account-tags' className='btn btn-link'>
```
jeffomatic: Now that we can't edit aliases, calling this "Edit" feels weird. Without consideration for implementation time, which of these do you prefer:

1. Changing this to "Edit tags"
2. Add an "edit" link in the right column of the "Tags" row.

dashboard/src/features/accounts/components/AccountUpdate.jsx
```diff
@@ -0,0 +1,119 @@
...
+  constructor(props) {
+    super(props)
+
+    this.submitWithErrors = this.submitWithErrors.bind(this)
+
+    this.state = {}
+  }
+
+  submitWithErrors(data) {
+    return new Promise((resolve, reject) => {
```
jeffomatic: I realize this is copy/paste of other code, but as a re-read it, I wonder if there's a simpler pattern here:

```
submitWithErrors(data) {
  return this.props.submitForm(data, this.props.item.id).catch(err => {
    throw {_error: err}
  })
}
```

dashboard/src/features/accounts/components/AccountUpdate.jsx
```diff
@@ -0,0 +1,119 @@
...
+        this.setState({notFound: true})
+      }
+    })
+  }
+
+  renderIfFound(view) {
+    if (this.state.notFound) {
+      return(<NotFound />)
+    } else if (view) {
+      return(view)
```
jeffomatic: extra parentheses

dashboard/src/features/accounts/components/AccountUpdate.jsx
```diff
@@ -0,0 +1,119 @@
...
+    } else {
+      return(<div>Loading...</div>)
+    }
+  }
+
+  render() {
+    const item = this.props.item
```
jeffomatic: I think this may have been copied from another component, but you may be able to get away with one less function if your `render` looked like this:

```
render() {
  if (this.state.notFound) {
    return <NotFound />
  }

  if (!this.props.item) {
    return <div>Loading...</div>
  }

  // stuff for rendering the form
}
```

dashboard/src/features/accounts/components/AccountUpdate.jsx
```diff
@@ -0,0 +1,126 @@
...
+      handleSubmit,
+      submitting
+    } = this.props
+
+    const title = <span>
+      {'Edit account tags '}
+      <code>{item.alias ? item.alias :item.id}</code>
+    </span>
+
+    const hint = <div>
```
jeffomatic: I notice that the hint disappears when there is an error (e.g. invalid JSON). I think the hint should persist regardless of errors. We may have to change it from a hint to a div that captions the whole form, rather than a single field?

dashboard/src/features/accounts/components/AccountUpdate.jsx
```diff
@@ -0,0 +1,126 @@
...
+      <p>Updating tags will overwrite existing tags. Contents must be represented as a JSON object.</p>
+      <p>
+        Account tags are used to annotate transactions where relevant. Changing the tags will only be reflected for future transactions,
+        and will not affect annotations for transactions that already exist.
```
jeffomatic: I think we'll probably want to document this more thoroughly. Can we put a placeholder link or a TODO just to make sure we don't forget about putting a link to docs here?

dashboard/src/features/assets/components/AssetUpdate.jsx
```diff
@@ -0,0 +1,126 @@
...
+  }
+  return {}
+}
+
+export default BaseUpdate.connect(
+  mapStateToProps,
+  BaseUpdate.mapDispatchToProps('asset'),
+  reduxForm({
+    form: 'updateAssetForm',
+    fields,
```
jeffomatic: How do you feel about inlining `fields` and `validate` here?

dashboard/src/features/shared/actions/update.js
```diff
@@ -0,0 +1,36 @@
...
+
+      return function(dispatch) {
+        return promise.then(() => clientApi.updateTags({
+          id: id,
+          tags: JSON.parse(data.tags),
+        }).then((resp) => {
+          dispatch(updated(resp))
+
+          let postCreatePath = listPath
+          if (options.redirectToShow) {
```
jeffomatic: I suspect we will always redirect to the show page. In that case, it might be good to remove this option altogether.

dashboard/src/features/shared/routes.js
```diff
@@ -2,7 +2,7 @@ import { RoutingContainer } from 'features/shared/components'
 import { humanize } from 'utility/string'
 import actions from 'actions'
 
-const makeRoutes = (store, type, List, New, Show, options = {}) => {
+const makeRoutes = (store, type, List, New, Show, Update, options = {}) => {
```
jeffomatic: I suspect this will break any invocation of `makeRoutes` that passes a value for `options`. Can you double-check to make sure?

jeffomatic: This shouldn't block the PR, but passing both an options object and having a bunch of optional components seems redundant. It might be better to take all arguments as part of `options`.

dashboard/src/features/shared/components/KeyValueTable/KeyValueTable.jsx
```diff
@@ -45,18 +45,31 @@ class KeyValueTable extends React.Component {
...
         actions={this.props.actions} >
         <table className={styles.table}>
           <tbody>
-            {this.props.items.map((item) =>
-              <tr key={`${item.label}`}>
-                <td className={styles.label}>{item.label}</td>
-                <td className={styles.value}>{this.renderValue(item)}</td>
-              </tr>
+            {this.props.items.map((item) => {
+              if (item.label == 'Tags') {
```
jeffomatic: Would it be difficult to change this so that instead of checking for a specific label value, we expand `item` to optionally include an `editUrl` property. If it is defined, then the link component will be included, and point to the given URL.

dashboard/src/features/shared/components/KeyValueTable/KeyValueTable.jsx
```diff
@@ -45,18 +45,31 @@ class KeyValueTable extends React.Component {
...
+                return <tr key={`${item.label}`}>
+                  <td className={styles.label}>{item.label}</td>
+                  <td className={styles.value}>{this.renderValue(item)}
+
+                      <Link to={`/${this.props.object}s/${this.props.id}/tags`} className={styles.edit} key='edit-tags'>
```
jeffomatic: 1. Looks like a double-indent here.
2. Does this actually require a `key`?

dashboard/src/features/shared/routes.js
dominic: That'll passing all the nulls a lot cleaner. I like it.

dashboard/src/features/shared/components/KeyValueTable/KeyValueTable.jsx
```diff
@@ -45,19 +45,22 @@ class KeyValueTable extends React.Component {
...
     return(
       <Section
         title={this.props.title}
         actions={this.props.actions} >
         <table className={styles.table}>
           <tbody>
-            {this.props.items.map((item) =>
-              <tr key={`${item.label}`}>
+            {this.props.items.map((item) => {
+              return <tr key={`${item.label}`}>
```
jeffomatic: Nit: I don't think you need to use a quoted string here...`item.label` is fine.

PR #934, opened 2017-04-14
protocol/vm: eliminate package vm_test

kr:
Now that there is no dependency cycle, all the tests can
move back into package vm.


kr:
PTAL

kr:
I think this change also fixes a couple of test bugs.

screen shot 2017-04-13 at 7 30 12 pm

Pretty sure it was wrong to call `vm.FalseResult` before; it should have been `gotVM.FalseResult` after saving the new exported vm object (first return parameter from Run).

protocol/vm/crypto_test.go
```diff
@@ -84,7 +82,7 @@ func TestCheckSig(t *testing.T) {
 			if err != nil {
 				t.Errorf("case %d: expected ok result, got error %s", i, err)
 			}
-		} else if !vm.FalseResult() {
```
kr: I think this PR also fixes a couple of test bugs, such as this line. Pretty sure this should have been saving a gotVM variable on line 78 and using that here as `gotVM.FalseResult`.

The new code I believe is correct, because the existing `vm` variable is updated.

bobg:
LGTM

PR #937, opened 2017-04-14
sdk/ruby: add account/asset tag updating

jeffomatic:


sdk/ruby/lib/chain/account.rb
```diff
@@ -53,6 +53,21 @@ def create_batch(opts)
         client.conn.batch_request('create-account', opts) { |item| Account.new(item) }
       end
 
+      # @param [Hash] opts Options hash specifiying account creation details.
+      # @option opts [String] id The ID of the account. Either an ID or alias should be provided, but not both.
+      # @option opts [String] alias The alias of the account. Either an ID or alias should be provided, but not both.
+      # @option opts [Hash] tags A new set of tags, which will replace the existing tags.
+      # @return [Hash] a success message.
+      def update_tags(opts)
+        client.conn.singleton_batch_request('update-account-tags', [opts])
```
dominic: I just realized it's funny that `singleton_batch_request` requires you to pass an array in.

jeffomatic: It's for accomodating build-transaction, which takes as its argument an object (rather than an array) with a timeout and an array of transactions. We can likely rethink this when we move to the new transaction template.

PR #938, opened 2017-04-14
dashboard: improve field errors and hint text

dominic:
Field errors now correctly check for the `touched` property, and display similarly on
both components. Hint text is rendered in markdown to support more complex hints.


dashboard/src/features/shared/components/SelectField.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import FieldLabel from './FieldLabel/FieldLabel'
 import pick from 'lodash/pick'
+var ReactMarkdown = require('react-markdown')
```
jeffomatic: Can we `import` here?

dominic: Ugh. I made the change and didn't push it before landing. 🚀 space cadet today.

PR #939, opened 2017-04-14
crypto/ed25519/math: Rename math.Uint256le to ecmath.Scalar for brevity

oleganza:
This renames `Uint256le` to `Scalar` since it's easier to read. Consequently, to avoid the spelling of `math.Scalar` as deceivingly abstract, the package is renamed to `ecmath`.

PTAL @bobg @kr @danrobinson


kr:
I am :+1: on this change. LGTM

PR #941, opened 2017-04-14
sdk/java: add ability to update account/asset tags

jeffomatic:


jeffomatic:
PTAL @boymanjor @dominic

sdk/java/src/main/java/com/chain/api/Account.java
```diff
@@ -255,6 +266,66 @@ public Builder setTags(Map<String, Object> tags) {
   }
 
   /**
+   * Use this class to update an account's tags.
+   */
+  public static class TagUpdateBuilder {
+    @SerializedName("alias")
```
boymanjor: nit: Are the @SerializedName annotations necessary?

sdk/java/src/test/java/com/chain/integration/UpdateTagsTest.java
```diff
@@ -0,0 +1,199 @@
...
+            .addFilterParameter(account2.id)
+            .execute(client);
+
+    List<Account> accountList = new ArrayList<>();
+    while (accounts.hasNext()) {
+      accountList.add(accounts.next());
+    }
+
+    assertEquals(accountList.get(0).tags.get("y"), "three");
+    assertEquals(accountList.get(1).tags.get("x"), "three");
```
boymanjor: nit: can we add a comment about the query results being in reverse chron. order of creation? I first questioned if we could just rely on this ordering until I checked.

boymanjor:
LGTM with some nits.

sdk/java/src/main/java/com/chain/api/Account.java
jeffomatic: Nice catch

sdk/java/src/test/java/com/chain/integration/UpdateTagsTest.java
jeffomatic: Good call

PR #942, opened 2017-04-14
sdk: remove control programs as first-class objects

jeffomatic:
Control programs are no longer presented to SDK users as first-class objects; instead, they occur implicitly via receivers and transaction outputs. This commit removes functionality that was deprecated in 1.1.


jeffomatic:
PTAL, ran the Node/Ruby integration tests locally

PR #943, opened 2017-04-14
core: protect API with policies, guards and grants 

tessr:


kr:
:eyes:

core/grants.go
```diff
@@ -0,0 +1,194 @@
...
+		// if there aren't any grants associated with this policy, go ahead
+		// and chuck this into raftdb
+		gList := &authz.GrantList{
+			Grants: []*authz.Grant{&g},
+		}
+		val, err := proto.Marshal(gList)
+		if err != nil {
+			return errors.Wrap(err)
+		}
+		err = a.raftDB.Set(ctx, grantPrefix+x.Policy, val)
```
kr: I suggest using Insert instead of Set here, so it will fail in case someone concurrently added another grant for this policy instead of clobbering the other value.

core/grants.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// create new grant and append to
+	grants = append(grants, &g)
+	gList := &authz.GrantList{Grants: grants}
+	val, err := proto.Marshal(gList)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+	err = a.raftDB.Set(ctx, grantPrefix+x.Policy, val)
```
kr: Can you add a todo here to make this safe for concurrent updates? Right now if two things are trying to append to the same list concurrently, we might lose one of them depending on the interleaving of the get+set operations. Pretty sure fixing this will require changing the raft interface though, and not worth it in this PR.

core/grants.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		grantList := new(authz.GrantList)
+		err = proto.Unmarshal(data, grantList)
+		if err != nil {
+			return nil, errors.Wrap(err)
+		}
+		for _, g := range grantList.GetGrants() {
+			var data chainjson.Map
+			err = data.UnmarshalJSON(g.GuardData)
```
kr: I think we don't need chainjson.Map here, can it be `data := json.RawMessage(g.GuardData)`?

core/grants.go
```diff
@@ -0,0 +1,194 @@
...
+			grants = append(grants, grant)
+		}
+	}
+
+	return map[string]interface{}{
+		"items": httpjson.Array(grants),
+	}, nil
+}
+
+func (a *API) revokeGrant(ctx context.Context, x struct {
```
kr: Should this use the apiGrant type?

core/grants.go
```diff
@@ -0,0 +1,194 @@
...
+	chainjson "chain/encoding/json"
+	"chain/errors"
+	"chain/net/http/authz"
+	"chain/net/http/httpjson"
+)
+
+// an api-friendly representation of a grant
+type apiGrant struct {
+	GuardType string        `json:"guard_type"`
+	GuardData chainjson.Map `json:"guard_data"`
```
kr: In order to get consistent serialization (e.g. sorting map keys) we have to completely deserialize the map here and then reserialize it again. Our chainjson.Map type is really just json.RawMessage with an extra check that the json is a map, so it preserves the original serialization which is not what we want. This line needs to be

```go
GuardData map[string]interface{} `json:"guard_data"`
```

or

```go
GuardData interface{} `json:"guard_data"`
```

core/grants.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// before we go any further, make sure this token is real
+	var gd map[string]string
+	err = json.Unmarshal(guardData, &gd)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+	id, ok := gd["id"]
+	if !ok || !a.accessTokens.CheckID(ctx, id) {
```
kr: We only want to do this if if the guard type is access token, right?

core/api.go
```diff
@@ -250,14 +253,17 @@ func (a *API) authnHandler(handler http.Handler) http.Handler {
...
 		}
 		handler.ServeHTTP(rw, req)
 	})
 }
 
-func authzHandler(handler http.Handler) http.Handler {
+func (a *API) authzHandler(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		if !authz.Authorized(req.Context()) {
+		auth := authz.NewAuthorizer(a.raftDB, grantPrefix, policyByRoute)
```
kr: Since the authorizer is basically stateless, this would work fine, but still I think it would be better to define this outside the closure.

net/http/authz/authorizer.go
```diff
@@ -0,0 +1,96 @@
...
+	return &Authorizer{
+		raftDB:        rdb,
+		raftPrefix:    prefix,
+		policyByRoute: policyMap,
+	}
+}
+
+func (a *Authorizer) Authorize(req *http.Request) error {
+	policies := a.policyByRoute[req.RequestURI]
+	if policies == nil {
```
kr: Maybe `len(policies) == 0` here to be defensive.

net/http/authz/authorizer.go
```diff
@@ -0,0 +1,96 @@
...
+		return errors.New("missing policy on this route")
```
kr: I think this should be a CH000. Prob this package should export `ErrNotAuthorized` (replacing `errNotAuthorized` from chain/core) that core can put in the errors table as a 401, but any other errors would show up as 500s.

net/http/authz/authorizer.go
```diff
@@ -0,0 +1,96 @@
...
+		return errors.New("not authorized")
+	}
+
+	return nil
+}
+
+func authzToken(ctx context.Context, grants []*Grant) bool {
+	for _, g := range grants {
+		if g.GuardType == "access_token" {
+			if accessTokenGuardData(g) == authn.Token(ctx) {
```
kr: I suggest combining these nested ifs into an `&&` expression.

net/http/authz/authorizer.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, g := range grants {
+		if g.GuardType == "localhost" {
+			return true
+		}
+	}
+	return authn.Localhost(ctx)
+}
+
+func accessTokenGuardData(grant *Grant) string {
+	var data map[string]string
```
kr: Consider using a struct type here and taking advantage of the fact that the value doesn't get set when there's an error:

```go
var v struct{ ID string }
json.Unmarshal(grant.GuardData, &v) // ignore error, returns "" on failure
return v.ID
```

core/api.go
tessr: Ha, oops, yes

core/api.go
```diff
@@ -168,8 +168,11 @@ func (a *API) buildHandler() {
 		}
 	}))
 
-	m.Handle("/create-access-token", jsonHandler(a.createAccessToken))
+	m.Handle("/list-acl-grants", jsonHandler(a.listGrants))
+	m.Handle("/create-acl-grant", jsonHandler(a.createGrant))
+	m.Handle("/revoke-acl-grant", jsonHandler(a.revokeGrant))
 	m.Handle("/list-access-tokens", jsonHandler(a.listAccessTokens))
+	m.Handle("/create-access-token", jsonHandler(a.createAccessToken))
```
tessr: This was a @dominic change, I think - Dom, any objection?

dominic: Of course. My goal was for `*-acl-grant` and `*-access-token` to be in sync.

dominic: (no objection, that is).

net/http/authz/authorizer.go
```diff
@@ -0,0 +1,96 @@
...
+		if g.GuardType == "access_token" {
+			if accessTokenGuardData(g) == authn.Token(ctx) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func authzLocalhost(ctx context.Context, grants []*Grant) bool {
```
tessr: note: use this to replace the authn localhost flag

core/grants.go
tessr: :+1: (The raft interface is going to need to support conditional writes for Iago, too.)

net/http/authz/authorizer.go
tessr: Gave myself a TODO.

core/grants.go
tessr: (Done.)

core/authz.go
```diff
@@ -0,0 +1,63 @@
...
+	"/list-acl-grants":     []string{"client-readwrite", "client-readonly"},
+	"/create-acl-grant":    []string{"client-readwrite"},
+	"/revoke-acl-grant":    []string{"client-readwrite"},
+	"/create-access-token": []string{"client-readwrite"},
+	"/list-access-tokens":  []string{"client-readwrite", "client-readonly"},
+	"/delete-access-token": []string{"client-readwrite"},
+	"/configure":           []string{"client-readwrite"},
+	"/info":                []string{"client-readwrite", "client-readonly"}, // also network, maybe?
+
+	"/debug/vars":          []string{"monitoring"}, // should monitoring endpoints also be available to any other policy-holders?
```
kr: Yes I think so.

core/authz.go
```diff
@@ -0,0 +1,63 @@
...
+	networkRPCPrefix + "block-height":      []string{"network"},
+
+	"/list-acl-grants":     []string{"client-readwrite", "client-readonly"},
+	"/create-acl-grant":    []string{"client-readwrite"},
+	"/revoke-acl-grant":    []string{"client-readwrite"},
+	"/create-access-token": []string{"client-readwrite"},
+	"/list-access-tokens":  []string{"client-readwrite", "client-readonly"},
+	"/delete-access-token": []string{"client-readwrite"},
+	"/configure":           []string{"client-readwrite"},
+	"/info":                []string{"client-readwrite", "client-readonly"}, // also network, maybe?
```
kr: Yeah, network and also monitoring I think.

core/authz.go
```diff
@@ -0,0 +1,63 @@
...
+var policies = []string{
+	"client-readwrite",
+	"client-readonly",
+	"network",
+	"monitoring",
+	"internal",
+}
+
+var policyByRoute = map[string][]string{
+	"/create-account":           []string{"client-readwrite"},
```
kr: Nit: these `[]string{...}` can all be just `{...}`.

core/grants.go
```diff
@@ -0,0 +1,187 @@
...
+	CreatedAt string                 `json:"created_at"`
+}
+
+// ErrMissingTokenID is returned when a token does not exist.
+var errMissingTokenID = errors.New("id does not exist")
+
+func (a *API) createGrant(ctx context.Context, x apiGrant) error {
+	id, ok := x.GuardData["id"]
+	if ok {
+		if strID, ok := id.(string); !ok || !a.accessTokens.Exists(ctx, strID) {
```
kr: This could avoid a level of nesting by using the unconditional map lookup (instead of the comma-ok form):

```go
if id, _ := x.GuardData["id"].(string); !a.accessTokens.Exists(ctx, id) {
	return errMissingTokenID
}
```

core/grants.go
```diff
@@ -0,0 +1,187 @@
...
+			return errMissingTokenID
+		}
+	} else {
+		return errMissingTokenID
+	}
+
+	guardData, err := json.Marshal(x.GuardData)
```
kr: I think it would be good to leave a note here that we rely on the consistent serialization format to compare grants for equality.

```
// NOTE: package json produces consistent serialization output,
// effectively an ad hoc canonical form. We rely on this to
// grant data for equality.
```

core/grants.go
```diff
@@ -0,0 +1,187 @@
...
+		if data == nil {
+			continue
+		}
+
+		grantList := new(authz.GrantList)
+		err = proto.Unmarshal(data, grantList)
+		if err != nil {
+			return nil, errors.Wrap(err)
+		}
+		for _, g := range grantList.GetGrants() {
```
kr: Nit: instead of GetGrants it would be fine (and IMO slightly preferable) to use the field itself `grantList.Grants` unless we need the extra nil check on the receiver (which we don't here).

core/grants.go
```diff
@@ -0,0 +1,187 @@
...
+			toRemove = index
+		}
+	}
+
+	// If there's no matching grant, return success
+	if toRemove == -1 {
+		return nil
+	}
+
+	grants = append(grants[:toRemove], grants[toRemove+1:]...)
```
kr: This whole section (lines 163-175) is okay as-is, but it might be more straightforward to copy the array as we go and omit the match. (That would also be more robust/defensive in case somehow some duplicate values got in there.)

```go
var keep []*Grant
for _, g := range grantList.Grants {
	if g.GuardType != x.GuardType || !bytes.Equal(g.GuardData, guardData) {
		keep = append(keep, g)
	}
}
if len(keep) == len(grantList.Grants) {
	return nil // no change; nothing to do
}
val, err := proto.Marshal(&authz.GrantList{Grants: keep})
etc.
```

net/http/authn/authn.go
```diff
@@ -46,6 +46,7 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 		ctx = newContextWithToken(ctx, token)
 	}
 
+	// TODO(tessr): Remove this in favor of localhost grants.
```
kr: This todo should go on func authorized in net/http/authz/loopback_authz.go, no?

net/http/authz/authorizer.go
```diff
@@ -0,0 +1,87 @@
...
+		return errors.New("missing policy on this route")
+	}
+
+	grants, err := a.grantsByPolicies(policies)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	if !authorized(req.Context(), grants) {
+		return errors.New("not authorized")
```
kr: This should be ErrNotAuthorized right?

core/api.go
```diff
@@ -250,15 +252,21 @@ func (a *API) authnHandler(handler http.Handler) http.Handler {
...
+	auth := authz.NewAuthorizer(a.raftDB, grantPrefix, policyByRoute)
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		if !authz.Authorized(req.Context()) {
-			errorFormatter.Write(req.Context(), rw, errNotAuthorized)
+		err := auth.Authorize(req)
+		if err != nil {
+			// TODO(tessr/dominic): Switch the following to:
+			// errorFormatter.Write(req.Context(), rw, err)
+			// once dashboard can handle ErrNotAuthorized as well as errNotAuthenticated
+			errorFormatter.Write(req.Context(), rw, errNotAuthenticated)
```
kr: I think this makes some 500s look like 401s. Should this be

```go
if errors.Root(err) == authz.ErrNotAuthorized {
	err = errors.Sub(errNotAuthenticated, err)
}
if err != nil {
	errorFormatter.Write(req.Context(), rw, err)
}
```

And maybe that's not necessary if https://github.com/chain/chain/pull/943/files#r111655629 is fixed?

net/http/authn/authn.go
tessr: Oh, yes. My brain.

net/http/authz/authorizer.go
tessr: Yes.

core/api.go
tessr: I think it's still necessary, because dashboard doesn't know how to interpret errNotAuthorized, but this is better.

net/http/authz/authorizer.go
```diff
@@ -0,0 +1,87 @@
...
+		if g.GuardType == "access_token" && accessTokenGuardData(g) == authn.Token(ctx) {
+			return true
+		}
+	}
+	return false
+}
+
+func authzLocalhost(ctx context.Context, grants []*Grant) bool {
+	for _, g := range grants {
+		if g.GuardType == "localhost" {
```
kr: I'm a little confused about the logic here. I would have expected

```go
for _, g := range grants {
    if g.GuardType == "localhost" && authn.Localhost(ctx) {
        return true
    }
}
return false
```

kr:
LGTM

PR #944, opened 2017-04-14
analytics: load Config from Oracle database

jbowens:
Load the configuration of custom columns from the Oracle database
on application start up. There's not yet a way to populate the
configuration besides manually through sqlplus.

Applied some of IntelliJ's static analysis suggestions too.


kr:
Is there (should there be) a rough description of how to set this up as a user?

jbowens:
> Is there (should there be) a rough description of how to set this up as a user?

Maybe. I was planning on writing a tool for configuring it next, so the user shouldn't need to set it up manually through SQL*Plus.

Issue #674, opened 2017-03-02
internal rpcs fail with tls configured


kr:
cored uses a hard-coded http (not https) scheme to connect to other cored instances. If cored is configured in production mode, it redirects all http requests to https, but the client side of the RPC doesn't follow redirects.

jbowens:
Fixed in `1.1-stable` with #675 but still outstanding on main.

jbowens:
Seems like the fix for `main` will probably involve exclusively using TLS. We can talk about it today at the 1.1 retro.

jbowens:
Since the TLS everywhere work is delayed until 1.3, we'll need to do a fix similar to what we did for 1.1.x in 1.2. I think we can just pass in a `core.TLS` `RunOption` if `*tlsCrt != ""` which tells `forwardToLeader` to use https.

kr:
SGTM, but I think maybe this should wait until after #898 lands. Even if we don't have TLS everywhere, we might have TLS everywhere-in-production, and the way to detect it will be different (i.e. a build tag).

kr:
Fixed on main in #675.

PR #946, opened 2017-04-17
core: support request forwarding over TLS

jbowens:
@boymanjor does this seem right? Or should it default to https and be set to http in cmd/cored/plain_http.go?

Also slightly tangential and it might be covered as part of the authorization work, but we could probably pin the certificate here? Instead of using one ROOT_CAS for all RPCs (internal & external), we could propagate a certpool and configure the chain/core/rpc.Client with it?

For #674.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -295,6 +295,9 @@ func launchConfiguredCore(ctx context.Context, raftDB *raft.Service, db *sql.DB,
 
 	opts = append(opts, core.IndexTransactions(*indexTxs))
 	opts = append(opts, enableMockHSM(db)...)
+	if *tlsCrt != "" {
+		opts = append(opts, core.ForwardUsingTLS)
+	}
 	// Add any configured API request rate limits.
 	if *rpsToken > 0 {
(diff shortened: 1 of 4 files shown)


boymanjor:
> @boymanjor does this seem right? Or should it default to https and be set to http in cmd/cored/plain_http.go?

I would prefer the latter and renaming the flag to `forwardUsingHTTP` or `forwardUsingPlainHTTP`. I really like using the `RunOption` over adding the `TLS` flag to the config object. I don't think we can change this based on the build tag though. Unless we want the `plain_http` flag to effectively disable TLS usage.

> Also slightly tangential and it might be covered as part of the authorization work, but we could probably pin the certificate here? Instead of using one ROOT_CAS for all RPCs (internal & external), we could propagate a certpool and configure the chain/core/rpc.Client with it?

Are you suggesting to take in a separate config var for internal vs. external RPCs or just pinning it here for convenience? I like the idea of pinning the certificate here if we will have different cert chains for the separate use cases. I could see a signer using their in-house PKI for intranode comms but being issued a cert from the generator for the external RPCs.

kr:
I think the approach here is pretty good. @boymanjor is right, we can't set this based on the plain_http build tag, so detecting using the configured cert is still the right way to go.

PR #947, opened 2017-04-17
protocol/bc: simplify names

bobg:
Now that legacy types and functions are in their own package, we can simplify the remaining `bc` identifiers and filenames without fear of collision.

Well, mostly: Renaming `TxEntries.TxInputs` to `Tx.Inputs` causes a field-name conflict via embedding in the `legacy.Tx` type. Rather than work around that, this PR removes the `TxEntries.TxInputs` field altogether in favor of lookups via `Tx.Entries` and `Tx.InputIDs` (for which a couple of new convenience functions have been introduced).

This PR also unexports `Entry.Type` (as `Entry.typ`).


bobg:
PTAL

kr:
LGTM

PR #948, opened 2017-04-17
docker/ci: Add Rust tools to Dockerfile

tonychain:
Installs Rust nightly (locked to 2017-04-16 snapshot) systemwide via rustup.sh:

https://github.com/rust-lang/rustup.sh

(Note that rustup.sh is technically deprecated, but presently the only way to
install Rust systemwide)

Also installs the following development tools:

- rustfmt: Lint syntax to ensure it follows standard style (ala gofmt)
- clippy: Lint for non-idiomatic Rust or other minor mistakes
- cargo-audit: Audit Cargo.lock for crates containing security vulnerabilities

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -25,6 +25,7 @@ RUN yum -y update \
     && install-java && rm -f /usr/bin/install-java \
     && install-maven && rm -f /usr/bin/install-maven \
     && install-ruby && rm -f /usr/bin/install-ruby \
+    && install-rust && rm -f /usr/bin/install-rust \
     && install-protobuf && rm -f /usr/bin/install-protobuf \
     && install-gas && rm -f /usr/bin/install-gas \
     && go get -u google.golang.org/grpc \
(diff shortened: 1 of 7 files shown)


kr:
Maybe also do the steps in https://github.com/chain/chain/blob/main/docker/ci/Readme.md and throw that in this PR as well? Specifically:

> ```
> TAG=chaindev/ci:`date +%Y%m%d`
> docker build --tag $TAG .
> docker push $TAG
> ```
>
> Then edit $CHAIN/wercker.yml to use the new tag.

So there would be a one-line  diff at the top of wercker.yml.

kr:
That would also cause this PR itself to test the new image.

tonychain:
The build succeeded (after a few attempts). Tagged as `20170417`

kr:
Do you mind also editing wercker.yml to pick up the new image here?

tonychain:
@kr done

kr:
LGTM

PR #951, opened 2017-04-17
cmd/cored: read TLS cert and key from files

kr:


kr:
PTAL

cmd/cored/main.go
```diff
@@ -255,6 +246,31 @@ func runServer() {
 	select {}
 }
 
+// maybeUseTLS loads the TLS cert and key (if so configured)
+// and wraps ln in a TLS listener.
+func maybeUseTLS(ln net.Listener) (net.Listener, bool, error) {
```
tessr: nit: either name the return values or explain the bool in the comment

tessr:
👀

cmd/cored/main.go
```diff
@@ -255,6 +246,31 @@ func runServer() {
...
+	keyFile := filepath.Join(*dataDir, "tls.key")
+
+	_, certErr := os.Lstat(certFile)
+	_, keyErr := os.Lstat(keyFile)
+	if os.IsNotExist(certErr) && os.IsNotExist(keyErr) {
+		return ln, false, nil // files don't exist; don't want TLS
+	}
+
+	config := &tls.Config{
+		NextProtos: []string{"http/1.1", "h2"},
```
tessr: maybe explain this too

cmd/cored/main.go
```diff
@@ -255,6 +246,31 @@ func runServer() {
 	select {}
 }
 
+// maybeUseTLS loads the TLS cert and key (if so configured)
+// and wraps ln in a TLS listener.
+func maybeUseTLS(ln net.Listener) (net.Listener, bool, error) {
+	certFile := filepath.Join(*dataDir, "tls.crt")
```
boymanjor: Do you think we should create a subdir for the certs?

kr:
I'm going to update this to also support the env vars for now, and we'll call them deprecated, and not remove them until the next release at the earliest.

cmd/cored/main.go
kr: That wouldn't be my first choice. I would rather they be more visible.

cmd/cored/main.go
kr: Oh actually I'm going to remove that return param.

kr:
Updated to address feedback. PTAL.

boymanjor:
LGTM

PR #952, opened 2017-04-17
core: fix log statement to print tx id correctly

jbowens:
Fix the transaction rejected log line to include the transaction hex.

It was including the bc.Hash struct:
```
app=cored buildtag=? processID=chain-f2473408badf-1312-75849f6537c61139932c reqid=8e02eefc091511e1137a subreqid=8a9f632c4d23ebc7fd76 at=api.go:406 t=2017-04-17T21:29:52.208568849Z status=400 chaincode=CH738 path=/submit-transaction error="tx {%!s(uint64=9959100764939771229) %!s(uint64=15390949537693335959) %!s(uint64=18383847051763105326) %!s(uint64=11411281842560845055)}: no tx sighash attempted"
```

Should we define a `String` method on the value receiver too? I think
the protobuf type only has one on the pointer receiver.

### DIFF
--- a/core/transact.go
+++ b/core/transact.go
@@ -142,7 +142,7 @@ func (a *API) submitSingle(ctx context.Context, tpl *txbuilder.Template, waitUnt
 
 	err := a.finalizeTxWait(ctx, tpl, waitUntil)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


kr:
Oh that's a bummer. Yeah maybe.

kr:
LGTM

PR #954, opened 2017-04-17
core: work around existing auth response handling

kr:
Dashboard knows how to interpret errNotAuthenticated
(CH009) but not the new authz.ErrNotAuthorized (CH011).
For now, just return the authn error even for authz
failures. We can remove this workaround once dashboard
is updated.


kr:
PTAL

boymanjor:
lgtm

PR #955, opened 2017-04-17
sdk/java: add tx rejection integration tests

jbowens:
Add integration tests testing various transaction rejection conditions
and their error codes.


kr:
LGTM

PR #956, opened 2017-04-17
core: use Set instead of Insert when storing grants

tessr:
Policy keys are occasionally getting set without any data, which leads us to this situation where we try to fetch the policy's grant list, see nothing, and then try to conditionally write (`Insert`) a new grant list--which fails, because the key has already been set. 

In the future, we will have more safe conditional writes, but this will fix it for now.


kr:
LGTM

PR #957, opened 2017-04-18
core: update access token endpoints to work with policies

tessr:


kr:
:eyes:

cmd/corectl/main.go
```diff
@@ -237,6 +237,10 @@ func createToken(db *sql.DB, args []string) {
 		fatalln("error:", err)
 	}
 	fmt.Println(tok.Token)
+
+	if *flagNet {
+		fmt.Println("warning: the network flag is deprecated")
```
kr: This message should go to stderr. Prob `fmt.Fprintln(os.Stderr, ...)` is the way to go.

core/migrate/data.go
```diff
@@ -60,4 +60,7 @@ var migrations = []migration{
 	{Name: `2017-04-13.0.query.block-transactions-count.sql`, SQL: `
 		ALTER TABLE annotated_txs ADD COLUMN block_tx_count INTEGER;
 	`},
+	{Name: `2017-04-17.0.core.null-token-type.sql`, SQL: `
+		ALTER TABLE access_tokens ALTER COLUMN type DROP NOT NULL;
```
kr: Don't we need to keep this so we can return the token's type if the user created it that way?

tessr: This just drops the constraint, not the column.

kr: Aha, my bad. :+1:

kr:
LGTM

PR #958, opened 2017-04-18
cmd/cored: support client cert authn in cored

boymanjor:
Clients can now authenticate themselves with `cored` using
x509 client certificates.


boymanjor:
PTAL. Tested using self signed certs.

boymanjor:
Updated, PTAL

net/http/authn/context.go
```diff
@@ -1,14 +1,32 @@
...
+	subjectNameKey
 )
 
+// newContextWithSubjectName sets the cert subject name in a new context and returns that context
+func newContextWithSubjectName(ctx context.Context, name *pkix.Name) context.Context {
+	return context.WithValue(ctx, subjectNameKey, name)
+}
+
+// SubjectName returns the cert subject name stored in the context, if it exists.
+func SubjectName(ctx context.Context) *pkix.Name {
```
kr: I think the name for this should be `X509SubjectName` or `X509Certs` (if the cert list is in the context).

net/http/authn/context.go
```diff
@@ -1,14 +1,32 @@
...
 type key int
 
 const (
 	tokenKey key = iota
 	localhostKey
+	subjectNameKey
 )
 
+// newContextWithSubjectName sets the cert subject name in a new context and returns that context
+func newContextWithSubjectName(ctx context.Context, name *pkix.Name) context.Context {
```
kr: I suggest inlining this.

net/http/authn/authn.go
```diff
@@ -60,6 +66,13 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 	return req.WithContext(ctx), nil
 }
 
+func (a *API) certAuthn(req *http.Request) (*pkix.Name, error) {
+	if req.TLS != nil && len(req.TLS.PeerCertificates) > 0 {
+		return &req.TLS.PeerCertificates[0].Subject, nil
```
kr: I would actually go all the way and put `req.TLS.PeerCertificates` itself in the context.

net/http/authn/authn.go
```diff
@@ -60,6 +66,13 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 	return req.WithContext(ctx), nil
 }
 
+func (a *API) certAuthn(req *http.Request) (*pkix.Name, error) {
```
kr: Prob cleaner to have this function return the new request after wrapping the context, then there's no need for an error value (which will only be dropped anyway).

net/http/authz/authorizer.go
```diff
@@ -69,6 +73,35 @@ func accessTokenGuardData(grant *Grant) string {
...
+		}
+	}
+	json.Unmarshal(grant.GuardData, &v)
+	return pkix.Name{
+		CommonName:         v.Subject.CommonName,
+		OrganizationalUnit: v.Subject.OrganizationalUnit,
+	}
+}
+
+func validateX509GuardData(a, b pkix.Name) bool {
```
kr: Nit: how about naming this `equalX509Name`?

boymanjor:
PTAL

net/http/authn/context.go
```diff
@@ -1,14 +1,27 @@
...
 	tokenKey key = iota
 	localhostKey
+	x509CertsKey
 )
 
+// X509Certs returns the cert stored in the context, if it exists.
+func X509Certs(ctx context.Context) []*x509.Certificate {
+	c, ok := ctx.Value(x509CertsKey).([]*x509.Certificate)
+	if !ok {
+		return []*x509.Certificate{}
```
kr: Nit: since c gets the zero value of a failed type assertion using the comma-ok form, we can just return c (which will be nil).

```go
c, _ := ctx.Value(x509CertsKey).([]*x509.Certificate)
return c
```

net/http/authz/authorizer.go
```diff
@@ -69,6 +74,35 @@ func accessTokenGuardData(grant *Grant) string {
 	return v.ID
 }
 
+func x509GuardData(grant *Grant) pkix.Name {
+	var v struct {
+		Subject struct {
+			CommonName         string   `json:"cn"`
+			OrganizationalUnit []string `json:"ou"`
```
kr: Can we leave a todo comment here about this mapping json to pkix.Name? We need some more comprehensive plan for it. The `cn` and `ou` thing were just ad hoc examples in the docs. I am pretty sure the spec did not intend for that to be definitive.

cc @jeffomatic @tonychain

kr:
Oh, but LGTM if that todo is added.

net/http/authz/authorizer.go
tonychain: I think we should support the standard X.500 attributes (at least the MUSTs). From RFC 3280:

https://datatracker.ietf.org/doc/html/rfc3280#section-4.1.2.4

```

   Standard sets of attributes have been defined in the X.500 series of
   specifications [X.520].  Implementations of this specification MUST
   be prepared to receive the following standard attribute types in
   issuer and subject (section 4.1.2.6) names:

      * country,
      * organization,
      * organizational-unit,
      * distinguished name qualifier,
      * state or province name,
      * common name (e.g., "Susan Housley"), and
      * serial number.

   In addition, implementations of this specification SHOULD be prepared
   to receive the following standard attribute types in issuer and
   subject names:

      * locality,
      * title,
      * surname,
      * given name,
      * initials,
      * pseudonym, and
      * generation qualifier (e.g., "Jr.", "3rd", or "IV").
```

PR #959, opened 2017-04-18
core/rpc: let the caller provide an HTTP client

kr:
This will be used to supply a TLS client cert for requests across corectl and cored.


kr:
PTAL

PR #960, opened 2017-04-18
cmd/cored: allow intra-node client auth

boymanjor:
`cored` can now authenticate itself during client calls
to other cored's using a tls certificate.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -166,6 +166,7 @@ func main() {
 	// That is the second phase.
 	mux := http.NewServeMux()
 	// TODO(tessr): authenticate raft endpoints
+	mux.Handle("/raft/", raftDB)
 
 	var handler http.Handler = mux
 	handler = reqid.Handler(handler)
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -193,6 +193,10 @@ func main() {
 
 	if useTLS {
 		http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{
+			// This is the default set of protocols for package http.
+			// DefaultTransport sets this automatically if TLSClientConfig
+			// is null. Since we are defining our own config, we must set it.
+			NextProtos:   []string{"http/1.1", "h2"},
 			Certificates: []tls.Certificate{*cert},
 			RootCAs:      loadRootCAs(*rootCAs),
 		}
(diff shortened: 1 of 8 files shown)


cmd/cored/main.go
```diff
@@ -139,15 +140,17 @@ func main() {
...
 	}
-	listener, err = maybeUseTLS(listener)
+	listener, cert, err := maybeUseTLS(listener)
 	if err != nil {
 		chainlog.Fatalkv(ctx, chainlog.KeyError, err)
 	}
-	_, useTCP := listener.(*net.TCPListener) // no TLS?
+	if cert != nil {
+		useTLS = true
+	}
```
kr: Instead of an if statement here can we define useTLS one one line:

```go
useTLS := cert != nil
```

cmd/cored/main.go
```diff
@@ -265,17 +271,18 @@ func maybeUseTLS(ln net.Listener) (net.Listener, error) {
...
+		cert, err = tls.LoadX509KeyPair(certFile, keyFile)
 	}
 	if err != nil {
-		return nil, errors.Wrap(err)
+		return nil, nil, errors.Wrap(err)
 	}
+	config.Certificates = []tls.Certificate{cert}
 	ln = tls.NewListener(ln, config)
-	return ln, nil
+	return ln, &cert, nil
```
kr: Nit: the diff would be smaller if we wrote

```go
return ln, config.Certificates[0], nil
```

here.

core/run.go
```diff
@@ -34,6 +34,10 @@ const (
 // RunOption describes a runtime configuration option.
 type RunOption func(*API)
 
+// ForwardUsingTLS configures the Core to use TLS when communicating between
+// Core processes.
+var ForwardUsingTLS RunOption = func(a *API) { a.forwardUsingTLS = true }
```
kr: Should this take a bool to turn it on or off?

boymanjor:
PTAL

core/run.go
boymanjor: Oooh. Totally.

boymanjor:
Updated. PTAL

kr:
LGTM

core/api.go
```diff
@@ -77,6 +77,7 @@ type API struct {
 	remoteGenerator *rpc.Client
 	indexTxs        bool
 	forwardUsingTLS bool
+	internalClient  *http.Client
```
kr: Two small bikeshed suggestions:

- call this httpClient
- add an end-of-line comment `// normally nil; only set in tests`

kr:
LGTM with a little suggestion

PR #961, opened 2017-04-18
cmd/corectl: present a TLS client cert if available

kr:
Read the Chain Core TLS cert/key pair from the data dir.
This allows corectl to connect to a running local cored
even when that cored is built with loopback auth disabled.


kr:
PTAL

boymanjor:
LGTM

PR #962, opened 2017-04-18
sdk/node: support client cert authn

boymanjor:
The Client object can be passed an `https.Agent` with
its `ca`, `key`, and `cert` attributes set as file paths.

### DIFF
--- a/sdk/node/src/connection.js
+++ b/sdk/node/src/connection.js
@@ -69,10 +69,10 @@ class Connection {
    * @param {String} agent   https.Agent used to provide TLS config.
    * @returns {Client}
    */
-  constructor(baseUrl, token = '', agent = {}) {
+  constructor(baseUrl, token = '', agent) {
     this.baseUrl = baseUrl
     this.token = token || ''
-    this.agent = agent || {}
+    this.agent = agent
   }
 
   /**
@@ -115,7 +115,7 @@ class Connection {
       req.headers['Authorization'] = `Basic ${btoa(this.token)}`
     }
 
-    if (Object.keys(this.agent).length !== 0) {
+    if (this.agent) {
       req.agent = this.agent
     }
 
--- a/sdk/node/src/connection.js
+++ b/sdk/node/src/connection.js
@@ -66,11 +66,13 @@ class Connection {
    *
    * @param {String} baseUrl Chain Core URL.
    * @param {String} token   Chain Core client token for API access.
+   * @param {String} agent   https.Agent used to provide TLS config.
    * @returns {Client}
    */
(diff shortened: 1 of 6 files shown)


boymanjor:
PTAL. Tested with https://gist.github.com/boymanjor/c6eae9847b7858c2aa9316b0ca35fa88

jeffomatic:
Hot take: I think we're at the point that we should offer an alternative constructor that looks like this:

```
c = new Chain.Client({
  baseUrl: "https://...",
  token: ...,
  agent: ...
})
```

boymanjor:
@jeffomatic I prefer that pattern of constructor function. Is that change a pre-requisite for landing this?

jeffomatic:
@boymanjor Yeah, I feel bad about blocking here, but I'd rather not introduce a calling convention that we're going to deprecate right away.

Are you comfortable updating the existing constructor so it permissively accepts the old convention (one or two string params) and the new convention (a single object param)? If not I can sneak in a PR ahead of yours that introduces that.

boymanjor:
@jeffomatic it's cool. I'll handle it.

boymanjor:
@jeffomatic PTAL. Updated and tested with the following config:
```
const https = require('https')
const fs = require('fs')
const chain = require('chain-sdk')

let agent = new https.Agent({
	ca: fs.readFileSync(process.env.ROOT_CA_CERTS),
	cert: fs.readFileSync(process.env.TLSCRT),
	key: fs.readFileSync(process.env.TLSKEY)
})
const client = new chain.Client({
  baseUrl: 'https://localhost:1999',
  agent: agent
})
```

jeffomatic:
LGTM but let's get @tonychain's sign-off here.

boymanjor:
@jeffomatic totally.

sdk/node/src/connection.js
```diff
@@ -66,11 +66,13 @@ class Connection {
...
    * @param {String} baseUrl Chain Core URL.
    * @param {String} token   Chain Core client token for API access.
+   * @param {String} agent   https.Agent used to provide TLS config.
    * @returns {Client}
    */
-  constructor(baseUrl, token = '') {
+  constructor(baseUrl, token = '', agent = {}) {
     this.baseUrl = baseUrl
     this.token = token || ''
+    this.agent = agent || {}
```
jeffomatic: I think if you do `this.agent = agent` with no defaulting, you can avoid checking for object key count in the request function below.

tonychain:
Haha, unfortunately my node-fu is not quite up to my Ruby. From what I can tell this seems fine though?

PR #963, opened 2017-04-18
core: add function TLSConfig

kr:


jbowens:
👀

cmd/cored/main.go
```diff
@@ -140,11 +136,11 @@ func main() {
 	if err != nil {
 		chainlog.Fatalkv(ctx, chainlog.KeyError, err)
 	}
-	listener, cert, err := maybeUseTLS(listener)
+	listener, tlsConfig, err := maybeUseTLS(listener)
```
jbowens: should `tlsConfig` be used on the http.Server constructed down below?

core/tls.go
```diff
@@ -0,0 +1,80 @@
...
+// If certFile and keyFile do not exist or are empty
+// and the environment vars are both unset,
+// TLSConfig returns ErrNoTLS.
+func TLSConfig(certFile, keyFile, rootCAs string) (*tls.Config, error) {
+	config := &tls.Config{
+		// This is the default set of protocols for package http.
+		// ListenAndServeTLS and Transport set this automatically,
+		// but since we're supplying our own TLS config,
+		// we have to set it here.
+		NextProtos: []string{"http/1.1", "h2"},
```
jbowens: just want to call out that currently cmd/cored disables http2:

```
// Disable HTTP/2 for now until the Go implementation is more stable.
// https://github.com/golang/go/issues/16450
// https://github.com/golang/go/issues/17071
TLSNextProto: map[string]func(*http.Server, *tls.Conn, http.Handler){},
```
If we do use this tls.Config on the http.Server it will start enabling it. That sounds fine to me but just wanted to call it out.

cmd/cored/main.go
```diff
@@ -191,12 +187,7 @@ func main() {
 		chainlog.Fatalkv(ctx, chainlog.KeyError, errors.Wrap(err, "Serve"))
 	}()
 
-	if useTLS {
-		http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{
-			Certificates: []tls.Certificate{*cert},
-			RootCAs:      loadRootCAs(*rootCAs),
-		}
-	}
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = tlsConfig
```
jbowens: wdyt about ferrying the config into the `chain/core` package, creating a new `http.Client` and installing it at  `core.API.httpClient`? Then we'd consistently use  `core.API.httpClient` in tests and in production.

I think manipulating `http.DefaultTransport` here could be a race with `raft.Start` which might already be using `http.DefaultClient` to make RPCs to other nodes?

cmd/cored/main.go
kr: Actually no: the Server's TLSConfig is used by ListenAndServeTLS, which we don't call. Instead, we make our own *tls.Listener and provide that to Server.Serve. It is confusing. :/

jbowens: ah ok, wdyt about updating the `TLSConfig` struct comment to explain that?

The mention of server tripped me up:
```
// TLSConfig returns a TLS config suitable for use
// as a Chain Core client and server.
```

core/tls.go
kr: So this part of the HTTP package is really hard to follow and not well documented. But that slice of strings is only used to initialized this map if the map is unset; our own assignment in cmd/cored will take precedence even if we set TLSConfig on the Server.

See https://github.com/golang/go/blob/48163968b2927247213fca7a6f4678d3c93855dc/src/net/http/server.go#L2946-L2948.

kr: *However*, it's confusing and unnecessary to be specifying h2 here when our intent is not to use it. So I will take this out. Good catch.

cmd/cored/main.go
kr: Yeah I think we should go in that direction, but I was hesitant to do too much all at once.

> could be a race with `raft.Start`

😮 Indeed. Will fix in the next PR.

PR #965, opened 2017-04-18
core: revoke grants when tokens are deleted

tessr:


tessr:
PTAL

core/grants.go
```diff
@@ -195,3 +195,54 @@ func (a *API) revokeGrant(ctx context.Context, x apiGrant) error {
...
+			if err != nil {
+				return errors.Wrap(err)
+			}
+
+			if id, _ := data["id"].(string); id != token {
+				keep = append(keep, g)
+			}
+		}
+
+		// We didn't match any grants, don't need to do an update. Return success
```
kr: Nit: "Return success"?

core/grants.go
```diff
@@ -195,3 +195,54 @@ func (a *API) revokeGrant(ctx context.Context, x apiGrant) error {
...
 }
+
+func (a *API) revokeGrantsByAccessToken(ctx context.Context, token string) error {
+	for _, p := range policies {
+		data, err := a.raftDB.Get(ctx, grantPrefix+p)
+		if err != nil {
+			return errors.Wrap(err)
+		}
+
+		if data == nil {
```
kr: This check seems unnecessary right?

tessr: yes, but it's kind of nice to short-circuit the rest of this loop, with all the marshalling and unmarshalling, if we don't need it?

core/access_tokens.go
```diff
@@ -87,6 +88,15 @@ func (a *API) deleteAccessToken(ctx context.Context, x struct{ ID string }) erro
...
+	err := a.accessTokens.Delete(ctx, x.ID)
+	if err != nil {
+		return err
+	}
+
+	err = a.revokeGrantsByAccessToken(ctx, x.ID)
+	if err != nil {
+		// well, technically we did delete the access token, so don't return the error
+		log.Printkv(ctx, log.KeyError, "revoking grants for access token %s", x.ID)
+	}
```
jbowens: if we did return an error here, clients should automatically retry and eventually succeed in cleaning up the grants too. not sure how important it is to revoke the grants too?

tessr: it's not _that_ important, but I suppose we could return an error. it would be weird to return an error if the token actually got deleted though (and we have no concept of, like, database transactions across raftdb and postgres).

kr: Yeah I think it should be a TODO to somehow make this atomic. (Would certainly be doable if they were all in raft or all in postgres, or we could try other trickery.)

core/grants.go
kr: I dunno, the rest of the function gracefully does nothing when data is nil. Are you thinking about performance? Or readability? I'd be surprised if it's appreciably faster this way.

core/access_tokens.go
kr: Oh but in the meantime, I am not sure whether this should return an error. We don't have a way to indicate "hey your thing worked, but you should retry it anyway".

I would lean toward not returning an error for the reasons @tessr mentioned. The primary operation actually did complete.

core/grants.go
tessr: I had been thinking about performance, but yes, you're right.

core/access_tokens.go
tessr: alright, I'll add a TODO. Do you think it's better to return an error when the token gets deleted from postgres, or to fail to delete the grant from raft? 

In general, I'd lean towards simply failing to delete the grant, except we're making some assumptions about the number of grants we store, so if someone were cycling through access tokens regularly we might cause performance problems.

kr: Yeah I agree silently failing is better.

kr:
LGTM

PR #966, opened 2017-04-18
sdk/ruby: support client cert authn

boymanjor:
File paths to the client cert, key and trusted CAs
can be passed through the opts param of the Client
constructor.

### DIFF
--- a/sdk/ruby/lib/chain/connection.rb
+++ b/sdk/ruby/lib/chain/connection.rb
@@ -189,10 +189,12 @@ module Chain
         if @opts.key?(:root_ca_certs_path)
           @http.ca_file = @opts[:root_ca_certs_path]
         end
-        if @opts.key?(:client_cert_path) && opts.key?(:client_key_path)
+        if @opts.key?(:client_cert_path)
           cert = File.read(@opts[:client_cert_path])
-          key = File.read(@opts[:client_key_path])
           @http.cert = OpenSSL::X509::Certificate.new(cert)
+        end
+        if opts.key?(:client_key_path)
+          key = File.read(@opts[:client_key_path])
           @http.key = OpenSSL::PKey::RSA.new(key)
         end
       end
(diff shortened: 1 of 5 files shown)


boymanjor:
PTAL. Tested with https://gist.github.com/boymanjor/001c35769ebb3ec963ef15be86d9bd75

jeffomatic:
What do you think about making the parameters a bit more self-documenting:

- `root_ca_certs_path`
- `client_cert_path`
- `client_key_path`

boymanjor:
SGTM

boymanjor:
Updated, PTAL.

sdk/ruby/lib/chain/connection.rb
```diff
@@ -186,6 +186,15 @@ def http
       if @url.scheme == 'https'
         @http.use_ssl = true
         @http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+        if @opts.key?(:root_ca_certs_path)
+          @http.ca_file = @opts[:root_ca_certs_path]
+        end
+        if @opts.key?(:client_cert_path) && opts.key?(:client_key_path)
```
jeffomatic: Nit: I have the feeling that unless we do more robust sanity checks here (i.e. raise an exception if one of the cert or key is missing), it may be safe to check for and set each of these in their own `if` blocks.

boymanjor: I'll updated to check in their own `if` blocks.

tonychain:
Whoops, didn't see this before it landed.

The typical approach I like doing for this sort of thing is using a `params` hash that maps directly to `#set_params=` on `OpenSSL::SSL::SSLContext`:

https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-set_params

In practice this looks like:

```ruby
ssl_params: { ca_path: certs, cert: client_cert, key: client_key }
```

This gives you the full flexibility of all of the options on `OpenSSL::SSL::SSLContext`

jeffomatic:
@tonychain sounds good, thanks for the heads-up

PR #970, opened 2017-04-19
core: always make an explicit http Client

kr:
This eliminates a data race in cmd/cored/main.go; we
were configuring DefaultTransport while using it
concurrently (to send raft RPCs).


kr:
PTAL

boymanjor:
lgtm

PR #972, opened 2017-04-19
core: trust the current process cert

kr:
A Chain Core TLS config is used by cored peers to talk
to each other, and by corectl to talk to cored. All
those processes have the same identity, so now we
automatically trust the local cert, with the expectation
that the peer will also be using it. This makes
authn misconfiguation impossible as long as the cert and
key files are loaded successfully.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bytes"
 	"context"
-	"crypto/x509"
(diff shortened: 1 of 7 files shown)


kr:
PTAL

kr:
Note that corectl was already doing this, but cored is picking up new behavior here.

jbowens:
lgtm

PR #973, opened 2017-04-19
net/http/authz: apply built-in localhost authz as a grant

kr:
This is mainly to prepare for another built-in grant to
trust the current process cert for internal access.


kr:
Next one: https://github.com/chain/chain/compare/lorefactor...self-authz

kr:
One after that: https://github.com/chain/chain/compare/self-authz...self-authz-plumbing

kr:
PTAL

jbowens:
lgtm

net/http/authz/authorizer.go
```diff
@@ -16,19 +16,28 @@ import (
 
 var ErrNotAuthorized = errors.New("not authorized")
 
+var builtinGrants []*Grant // initialized in loopback_authz.go
+
 type Authorizer struct {
 	raftDB        *raft.Service
 	raftPrefix    string
 	policyByRoute map[string][]string
+	extraGrants   map[string][]*Grant
```
tessr: 🚲 🏠 : I'm not crazy about this name, can we pick something more descriptive? (`builtinGrants` seems good, actually...)

tessr: or even `builtinGrantsByPolicy`

kr: Hmm, okay. (They're not all necessarily built-in, though. See next PR https://github.com/chain/chain/compare/lorefactor...self-authz.)

tessr: Well, by built-in, I meant in memory rather than in raftdb.

tessr: which is maybe the problem with all these names.

tessr: (of course, not a big deal--if we don't have another name handy, let's not worry about it)

kr: The name doesn't have to be an essay; it shouldn't try to convey everything about how the field is going to be used. It has to be unique and evocative enough to remind you what it is when seen in context, and to make it easy to find its definition.

Specifically, in this case, what I'm going for is a way to distinguish this name from just `grants`, since, ordinarily, grants are loaded from the db. So `extraGrants` says that. But I'm specifically not trying to describe what data structure we're using here or really anything about how it's organized, because that's less germane (and more readily inferred from the code).

So:

```go
extraGrants map[string][]*Grant // by policy
```

or

```go
builtinGrants map[string][]*Grant // by policy
```

These both seem pretty good to me. (I slightly prefer "extra" but don't feel strongly about it.) Actual uses will look like `a.builtinGrants[policy]` so it'll be clear what the key is.

If it's not clear from the code what the meaning of the field is, I'm happy to write a comment explaining more about its purpose. (But I would rather write that explanation in a comment than inside the variable name.)

kr: (Actually, I might have called it just `extra` since the fact that it holds grants is also obvious from context.)

PR #974, opened 2017-04-19
sdk/ruby: configure client authn with ssl_params hash

boymanjor:
Previously, the Client constructor accepted cert and key
data as separate attributes on its opts config parameter.
It will now accept tls options on an `ssl_params` hash.
The options supported map to `OpenSSL::SSL:Context` params.
Currently, we support `ca_file`, `cert` & `key`.

### DIFF
--- a/generated/rev/RevId.java
+++ b/generated/rev/RevId.java
@@ -1,4 +1,4 @@
 
 public final class RevId {
-	public final String Id = "main/rev2952";
+	public final String Id = "main/rev2953";
 }
--- a/generated/rev/revid.go
+++ b/generated/rev/revid.go
@@ -1,3 +1,3 @@
 package rev
 
-const ID string = "main/rev2952"
+const ID string = "main/rev2953"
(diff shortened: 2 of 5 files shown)


boymanjor:
PTAL. Tested with this config:
```
require_relative './src/chain/sdk/ruby/lib/chain'

cert = OpenSSL::X509::Certificate.new(File.read(ENV['TLSCRT']))
key = OpenSSL::PKey::RSA.new(File.read(ENV['TLSKEY']))
ca_file = ENV['ROOT_CA_CERTS']
c = Chain::Client.new(url: "https://localhost:1999", ssl_params: { cert: cert, key: key, ca_file: ca_file })
```

jeffomatic:
alley-oop pass to @tonychain

tonychain:
Wow, I guess `Net::HTTP` really doesn't provide a way to configure an `OpenSSL::SSL::SSLContext` other than `instance_eval`. Lame!

Anywho, LGTM 👍

PR #975, opened 2017-04-19
net/http/authz: allow caller to add an extra grant

kr:
Package chain/core will use this to authorize other
cored and corectl processes using the same Subject Name.

### DIFF
--- a/generated/rev/RevId.java
+++ b/generated/rev/RevId.java
@@ -1,4 +1,4 @@
 
 public final class RevId {
-	public final String Id = "main/rev2961";
+	public final String Id = "main/rev2962";
 }
--- a/generated/rev/revid.go
+++ b/generated/rev/revid.go
@@ -1,3 +1,3 @@
 package rev
 
-const ID string = "main/rev2961"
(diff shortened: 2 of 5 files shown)


kr:
PTAL

kr:
Should this set CreatedAt?

tessr:
Yes. Dashboard uses it for sorting.

tessr:
Actually, maybe not. Will this get displayed in dashboard?

kr:
It won't show up in the API response from this code, since core reads it directly from raft.

I would like to make it show up some day. I'll add the timestamp anyway, it won't hurt.

kr:
Updated. PTAL

tessr:
This LG, did we resolve how we're going to format x509 cert data?

kr:
> did we resolve how we're going to format x509 cert data?

Not yet. I decided to do some research and then promptly got pulled off to some meetings.

kr:
Assuming we redesign our x509 grant data format, yeah we'll have to update this new code. It's right alongside the other code so hopefully it will be hard to miss.

PR #977, opened 2017-04-19
sdk/java: new Client.Builder constructor

boymanjor:
The new constructor takes a Client object as an argument. The private
members of the Client are copied into the resulting builder.
To facilitate this change, a private OkHttpClient member was added to
the builder (no setter method was provided). A new builder method was
also added to allow a user to set a list of URLs.

The impetus for this change was a bug I found when adding client cert
authn to the SDK. Our current method of creating a Client for the MockHsm
(`getSignerClient`) does not persist any configurations set on the httpClient.
The user can get around this by creating a separate Client with the appended
"/mockhsm" string but our provided mechanism is broken as is.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -56,7 +56,7 @@ public class Client {
   }
 
   public Client(Builder builder) {
-    List<URL> urls = new ArrayList<URL>(builder.urls);
+    List<URL> urls = new ArrayList<>(builder.urls);
     if (urls.isEmpty()) {
       try {
         urls.add(new URL("http://localhost:1999"));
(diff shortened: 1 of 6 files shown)


sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -379,19 +379,24 @@ public void setProxy(Proxy proxy) {
   }
 
   private OkHttpClient buildHttpClient(Builder builder) {
-    OkHttpClient httpClient = new OkHttpClient();
+    OkHttpClient httpClient = builder.httpClient.clone();
+    httpClient.setFollowRedirects(false);
```
jeffomatic: Is this property carried across a `clone`? If so, it may be better to call this when the `httpClient` member is instantiated, back in the builder.

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -379,19 +379,24 @@ public void setProxy(Proxy proxy) {
...
     }
-
-    httpClient.setFollowRedirects(false);
-    httpClient.setReadTimeout(builder.readTimeout, builder.readTimeoutUnit);
-    httpClient.setWriteTimeout(builder.writeTimeout, builder.writeTimeoutUnit);
-    httpClient.setConnectTimeout(builder.connectTimeout, builder.connectTimeoutUnit);
-
-    httpClient.setConnectionPool(builder.pool);
-
+    if (builder.readTimeoutUnit != null) {
```
jeffomatic: It feels like if we're instantiating `httpClient` in the builder, these timeouts should be set directly on the builder's `httpClient` object. In other words, maybe `buildHttpClient` is something that moves into the `Builder` class?

boymanjor: Discussed out of band: `Client#buildHttpClient` sets a couple of parameters on the httpClient that require an accumulation of dependent properties, i.e. we have to set a logger and log level before we can add a request interceptor. In the future, we will need to set trusted ca’s and client certs before we can build the SslSocketFactory. In lieu of adding explicit create calls for Builder fields, we will keep this as is and update the name of `Builder#httpClient` to `Builder#baseHttpClient` so it's clear that the field is a base that can be set and not the actual object the builder setters are operating upon.

boymanjor:
Updated, PTAL.

jeffomatic:
LGTM

PR #978, opened 2017-04-19
encoding/blockchain: guard against large mallocs

jbowens:
Don't trust the element count when decoding VarstrLists. Allocating an
array at that size may exhaust all memory available. I found this
particular instance using fuzz testing. I think there are probably more
instances of this, like `ReadVarstring31`.

This bug seems most relevant to a multi-Core network. A bad actor could
submit a transaction to the generator with a math.MaxInt32 element in a
VarstrList to crash the generator. Because the transaction doesn't
actually encode math.MaxInt32 elements, it doesn't hit max byte request
body limits.

To fix this generally, we could:
 * read incrementally into smaller buffers rather than allocating one
   buffer that is the claimed size.
 * not use the `io.Reader` interface and instead expose the length of
   the unread buffer to decoding routines for validation.
 * implement stricter length limits than the protocol currently permits.

### DIFF
--- a/encoding/blockchain/blockchain.go
+++ b/encoding/blockchain/blockchain.go
@@ -122,7 +122,7 @@ func ReadVarstrList(r io.Reader) ([][]byte, int, error) {
 	if nelts == 0 {
 		return nil, n, nil
 	}
-	result := make([][]byte, 0, nelts)
+	var result [][]byte
 	for ; nelts > 0; nelts-- {
 		s, n2, err := ReadVarstr31(r)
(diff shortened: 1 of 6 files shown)


jbowens:
PTAL

kr:
:eyes:

kr:
LGTM

kr:
> - not use the `io.Reader` interface and instead expose the length of
the unread buffer to decoding routines for validation.

I like the sound of that.

PR #979, opened 2017-04-19
dashboard: coalesce authorization grants by guard data

dominic:
When rendering access control objects, we merge multiple 
authorization grants into a single table row. Policies for
the merged row can be edited inline.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -43,6 +43,8 @@
     "redux": "~3.5.2",
     "redux-form": "~5.3.2",
     "redux-thunk": "~2.1.0",
+    "reselect": "^3.0.0",
+    "sha.js": "^2.4.8",
     "uuid": "~2.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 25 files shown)


jeffomatic:
Code LGTM, but one behavior thing that is weird is that editing a token will change its position in the list. I think it makes sense to order objects in reverse-chronological order, but that should be stable after objects are created.

How does this sound:

- Sort access tokens by token creation date.
- Sort X509 entries by earliest policy grant.

Dealing with the post-edit state is tricky. Some ideas:

- Option 1: After a successful edit, include the edited item, including the policy, in the update message. For X509 certs, we can do a comma-separated concat of the subject, although there's no solution I love here. If it looks good, highlight the edited row.
- Option 2: treat the edit form like a single-item view, and simply reload the edit view when the editing has completed.
- Option 3: edit policies inline by replacing the list of policies with a control that lets you add/remove polices without loading a new "page".

dominic:
@jeffomatic Okay great - let me see which option feels best. I'm inclined towards option 3.

dominic:
For the casual observer, we've switched to inline editing for tokens and certificates:

screen shot 2017-04-20 at 2 46 17 pm

PR #980, opened 2017-04-19
cmd: rename CORED_DATA_DIR to DATA_DIR

tessr:
@jeffomatic asked why we named this environment variable CORED_DATA_DIR instead of simply DATA_DIR. I couldn't think of anything other than I tend to over-specify things.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -31,7 +31,7 @@ import (
 
 // config vars
 var (
-	dataDir = env.String("CORED_DATA_DIR", fileutil.DefaultDir())
+	dataDir = env.String("DATA_DIR", fileutil.DefaultDir())
 	dbURL   = env.String("DATABASE_URL", "postgres:///core?sslmode=disable")
 	coreURL = env.String("CORE_URL", "http://localhost:1999")
 
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -65,7 +65,7 @@ var (
 	rpsToken      = env.Int("RATELIMIT_TOKEN", 0)       // reqs/sec
 	rpsRemoteAddr = env.Int("RATELIMIT_REMOTE_ADDR", 0) // reqs/sec
 	indexTxs      = env.Bool("INDEX_TRANSACTIONS", true)
-	dataDir       = env.String("CORED_DATA_DIR", fileutil.DefaultDir())
+	dataDir       = env.String("DATA_DIR", fileutil.DefaultDir())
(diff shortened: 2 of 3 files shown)


jbowens:
sounds/looks gtm

kr:
I think it should have "chain" or "core" or "cored" in the name.

tessr:
Jeff's point was that none of our other environment variables specify that.

tessr:
(Granted, we have a real menagerie of variables going on here.)

jbowens:
> (Granted, we have a real menagerie of variables going on here.)

🐻 🐦 🐸 :shipit:🐒 🦌 🦍 🐈 🐺

tessr:
giphy

kr:
> Jeff's point was that none of our other environment variables specify that.

When we began our naming scheme, it was back when we were operating everything ourselves, so we knew that the target system was going to run cored and nothing else. But now, we have a variety of different deployment targets, including individual desktop users. It would be a lot more polite not to assume we have the run of their env var name space. It's reasonable to ask someone to put `CORED_DATA_DIR` in their `.profile`; less reasonable to ask them to put `DATA_DIR` in their `.profile`.

Yeah we still have a bunch of unqualified env var names sitting around, so the best we can do is phase them out or just deprecate them.

But especially if we are moving to a world where we store config in files on disk, then this will be the one env var to rule them all, and it would be good to be a good citizen from now on.

BTW, if we are going to 🚲 :house: this variable name, I would like it not to have "DIR". My suggestions:

```
COREDROOT
COREDHOME
COREDDATA
CHAINCOREROOT
CHAINCOREHOME
CHAINCOREDATA
```

(Or, those same names but with underscores, if you must.)

jeffomatic:
> It would be a lot more polite not to assume we have the run of their env var name space

SGTM. I was arguing from a consistency standpoint, but if we want to start a new trend then I'm all for it. Taking this idea to the logical end, it may be good to start deprecating the old envvars in favor of more politely-named variants.

jbowens:
> it may be good to start deprecating the old envvars in favor of more politely-named variants.

Assuming we still want them to be environment variables. We've thrown around the idea of culling our menagerie by introducing file-based configuration. We might want to hold off on introducing renamed environment variables until we put in the time to think about what we want configuration to look like going forward.

tessr:
@jeffomatic I am going to close this, if that's alright with you.

jeffomatic:
@tessr Yep fine, thanks for the heads-up.

PR #984, opened 2017-04-20
sdk/java: support client cert authn

boymanjor:
File paths to the client cert and key can now be passed as `TLSCRT`
and `TLSKEY` respectively. RSA private keys adhering to both PKCS#1
and PKCS#8 are supported. The Java standard library doesn't support
PEM encoded keys so the BouncyCastle Crypto APIs library has been
added for its parsing functionality.

### DIFF
--- a/sdk/java/src/main/java/com/chain/http/Client.java
+++ b/sdk/java/src/main/java/com/chain/http/Client.java
@@ -5,11 +5,19 @@ import com.chain.common.*;
 
 import java.io.*;
 import java.lang.reflect.Type;
+import java.math.BigInteger;
 import java.net.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.security.GeneralSecurityException;
+import java.security.KeyFactory;
 import java.security.KeyStore;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.security.spec.KeySpec;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.RSAPrivateCrtKeySpec;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -25,6 +33,10 @@ import com.squareup.okhttp.Request;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
 
+import sun.misc.BASE64Decoder;
+import sun.security.util.DerInputStream;
+import sun.security.util.DerValue;
+
 import javax.net.ssl.*;
 
 /**
@@ -35,12 +47,18 @@ import javax.net.ssl.*;
  * to an HSM server.
  */
 public class Client {
-
   private AtomicInteger urlIndex;
   private List<URL> urls;
   private String accessToken;
   private OkHttpClient httpClient;
+
   private static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
+  private static final String PKCS1_HEADER = "-----BEGIN RSA PRIVATE KEY-----\n";
+  private static final String PKCS1_FOOTER = "-----END RSA PRIVATE KEY-----\n";
+  private static final String PKCS8_HEADER = "-----BEGIN PRIVATE KEY-----\n";
+  private static final String PKCS8_FOOTER = "-----END PRIVATE KEY-----\n";
+  // Used to create empty, in-memory key stores.
+  private static final char[] DEFAULT_KEYSTORE_PASSWORD = "password".toCharArray();
   private static String version = "dev"; // updated in the static initializer
 
   private static class BuildProperties {
@@ -55,7 +73,7 @@ public class Client {
     }
   }
 
-  public Client(Builder builder) {
+  public Client(Builder builder) throws HTTPException {
     List<URL> urls = new ArrayList<>(builder.urls);
     if (urls.isEmpty()) {
       try {
@@ -74,7 +92,7 @@ public class Client {
   /**
    * Create a new http Client object using the default development host URL.
    */
-  public Client() {
+  public Client() throws ChainException {
     this(new Builder());
   }
 
@@ -83,7 +101,7 @@ public class Client {
    *
    * @param url the URL of the Chain Core or HSM
    */
-  public Client(String url) throws BadURLException {
+  public Client(String url) throws ChainException {
     this(new Builder().setURL(url));
   }
 
@@ -92,7 +110,7 @@ public class Client {
    *
    * @param url the URL of the Chain Core or HSM
    */
-  public Client(URL url) {
+  public Client(URL url) throws HTTPException {
     this(new Builder().setURL(url));
   }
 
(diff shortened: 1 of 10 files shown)


boymanjor:
This depends and is rebased on #977 so not quite ready for AL but I wouldn't mind eyes on the private key decoding @tonychain.

I tested using https://gist.github.com/boymanjor/a3980132d70149918b162564dd6042d2 with both key formats. Used the following to convert from PKCS#1 to PKCS#8: 
```
$ openssl pkcs8 -topk8 -inform PEM -outform PEM -nocrypt -in pkcs1.key -out pkcs8.key
```

tarcieri:
@boymanjor one thing I'd be curious about is if okhttp3 supports PEM client certificates natively (not sure how much work it'd be to upgrade)

I know they added native support for PEM truststores:

https://github.com/square/okhttp/commit/f3b4e8895bada51907987de6d025f2df446edee9

boymanjor:
@tonychain ah, cool. I'll do some digging.

boymanjor:
Update: the above example is actually using the Java standard library to build the trust store. Okhttp seems to just allow you to set a custom SslSocketFactory. We're doing almost the exact same thing in our client. That is encouraging.

tonychain:
Sweet, sounds good then

boymanjor:
PTAL

boymanjor:
PTAL @jeffomatic @jbowens @erykwalder AKA The Java Bean Buddies

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -5,11 +5,19 @@
 
 import java.io.*;
 import java.lang.reflect.Type;
+import java.math.BigInteger;
 import java.net.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.security.GeneralSecurityException;
+import java.security.KeyFactory;
```
jeffomatic: Does `import java.security.*` work? Probably time for that trick.

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -25,6 +33,10 @@
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
 
+import sun.misc.BASE64Decoder;
```
jeffomatic: Is this available as part of the standard Java 7 distribution?

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -633,13 +658,96 @@ public Builder setAccessToken(String accessToken) {
     }
 
     /**
+     * Sets the client's certificate and key for TLS client authentication.
+     * @param certPath filepath to pem encoded X.509 certificate
+     * @param keyPath filepath to pem encoded X.509 private key
+     */
+    public Builder setX509KeyPair(String certPath, String keyPath) throws HTTPException {
```
jeffomatic: How do you feel about taking a bytestream for both of these, rather than doing the file-opening in the client. Or, you can make an overload that does the file opening for the user if they want.

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -633,13 +658,96 @@ public Builder setAccessToken(String accessToken) {
...
+        X509Certificate certificate = (X509Certificate) factory.generateCertificate(pemStream);
+
+        // The PKCS standard used by the private key must be determined.
+        // Currently, Chain Core supports PKCS#1 and PKCS#8.
+        // In both cases, the header and footer ascii characters must be stripped and the remaining
+        // Base64 data decoded to DER encoded bytes. In the case of PKCS#8, the standard library
+        // provides a class representation that accepts these bytes as inputs.
+        // For PKCS#1 we must parse these bytes for specific integer values representing
+        // the ASN.1 RSAPrivateKey: https://tools.ietf.org/html/rfc3447#appendix-A.1.2
+        byte[] keyBytes = Files.readAllBytes(Paths.get(keyPath));
```
jeffomatic: I think I might be in favor of adding a dependency lib that can take a vanilla PEM-encoded cert and just Do The Right Thing as an off-the-shelf operation, rather than have a bunch of bespoke code in the SDK.

sdk/java/src/main/java/com/chain/http/Client.java
boymanjor: We'd have to do it at every package level, i.e. `java.security.*`, `java.security.cert.*`, etc. and if these packages use matching class names there will be conflicts. I'd rather leave it as is.

boymanjor: Actually, both `java.security` and `java.security.cert` have a Certificate class.

boymanjor:
@jeffomatic ready for AL.

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -633,28 +652,89 @@ public Builder setAccessToken(String accessToken) {
...
+     * @param keyPath file path to pem encoded private key
+     */
+    public Builder setX509KeyPair(String certPath, String keyPath) throws HTTPException {
+      try (InputStream certStream =
+              new ByteArrayInputStream(Files.readAllBytes(Paths.get(certPath)));
+          InputStream keyStream =
+              new ByteArrayInputStream(Files.readAllBytes(Paths.get(keyPath)))) {
+        return setX509KeyPair(certStream, keyStream);
+      } catch (IOException ex) {
+        throw new HTTPException("Unable to store X509 cert/key pair", ex);
```
jeffomatic: I think HTTPException is a weird error to throw here. HTTPException is a transport-level error during a request. We may need to make a new error class that represents configuration errors.

Might want to double-check that HTTPException is being correctly used in other sites as well.

boymanjor: Yea, I see what you mean.

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -633,28 +652,89 @@ public Builder setAccessToken(String accessToken) {
     }
 
     /**
+     * Sets the client's certificate and key for TLS client authentication.
+     * @param certStream input stream of pem encoded X.509 certificate
```
jeffomatic: s/pem encoded/PEM-encoded

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -633,28 +652,89 @@ public Builder setAccessToken(String accessToken) {
     }
 
     /**
+     * Sets the client's certificate and key for TLS client authentication.
```
jeffomatic: Might be good to specify the valid encodings and formats of the certs.

sdk/java/src/main/java/com/chain/http/Client.java
```diff
@@ -633,28 +652,89 @@ public Builder setAccessToken(String accessToken) {
...
+
+        this.keyManagers = keyManagerFactory.getKeyManagers();
+        return this;
+      } catch (GeneralSecurityException | IOException ex) {
+        throw new ConfigurationException("Unable to store X.509 cert/key pair", ex);
+      }
+    }
+
+    /**
+     * Sets the client's certificate and key for TLS client authentication.
```
jeffomatic: Ditto here for all Javadoc comments above.

Issue #981, opened 2017-04-19
bug: accessing routes that don't exist yield CH000


jeffomatic:
Accessing non-existent RPC routes yields a server error (CH000, 500 status code), when it should be 404, e.g.

```
app=cored buildtag=? processID=chain-jeff-chain.local-12535-e47e47ee636a3489f2e4 reqid=4c97ce4ccc6c7c2f7c8c at=api.go:272 t=2017-04-19T22:46:21.569480246Z status=500 chaincode=CH000 path=/non-existent-route error="missing policy on this route"
```

tessr:
what should the CHXXX code be?

dominic:
Is `CH404` too on the nose? That would communicate a lot to the user right off the bat.

tessr:
We don't use `CH4XX` for anything (did you mean a 404 http status? In which case, I agree.) This question was more like... should we use `CH006`, which is what Not Found is right now?

jeffomatic:
> should we use CH006, which is what Not Found is right now?

SGTM

tessr:
Resolved by #985.

PR #985, opened 2017-04-20
core: return 404 when hitting a non-existent, policy-less route

tessr:
This also means that a 404 will be returned for routes that exist but don't have defined policies, but I think that's okay. (The alternatives seem heavy-handed.) 

This addresses #981.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -262,7 +262,7 @@ func (a *API) authnHandler(handler http.Handler) http.Handler {
 func (a *API) authzHandler(mux *http.ServeMux, handler http.Handler) http.Handler {
 	auth := authz.NewAuthorizer(a.raftDB, grantPrefix, policyByRoute)
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		// return failure ealy if this path isn't legit
+		// return failure early if this path isn't legit
 		if _, pat := mux.Handler(req); pat != req.URL.Path {
 			errorFormatter.Write(req.Context(), rw, errNotFound)
 			return
--- a/core/api.go
+++ b/core/api.go
@@ -190,7 +190,7 @@ func (a *API) buildHandler() {
 		m.ServeHTTP(w, req)
 	})
 
-	handler := a.authzHandler(latencyHandler)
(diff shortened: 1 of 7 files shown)


tessr:
PTAL

kr:
LGTM

Btw if you are interested in exploring one of those alternatives, we faced a similar problem with metrics. Here id what we did https://github.com/chain/chain/blob/a8ede5e1/core/metrics.go#L32-L33

tessr:
Oh, that seems not-so-terrible. I'll play with that.

tessr:
Alright, this is now more precise about errors--it returns a 404 for non-existent routes, but it continues to return a 500 for a missing policy (although that could also easily be updated).

core/api.go
```diff
@@ -259,9 +259,14 @@ func (a *API) authnHandler(handler http.Handler) http.Handler {
 	})
 }
 
-func (a *API) authzHandler(handler http.Handler) http.Handler {
+func (a *API) authzHandler(mux *http.ServeMux, handler http.Handler) http.Handler {
 	auth := authz.NewAuthorizer(a.raftDB, grantPrefix, policyByRoute)
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		// return failure ealy if this path isn't legit
```
kr: nit: ealy

net/http/authz/authorizer.go
```diff
@@ -41,7 +44,7 @@ func NewAuthorizer(rdb *raft.Service, prefix string, policyMap map[string][]stri
 func (a *Authorizer) Authorize(req *http.Request) error {
 	policies := a.policyByRoute[strings.TrimRight(req.RequestURI, "/")]
 	if policies == nil || len(policies) == 0 {
-		return errors.New("missing policy on this route")
+		return ErrMissingPolicy
```
kr: Do we still need this change?

kr:
LGTM after a couple small things

net/http/authz/authorizer.go
tessr: we don't _need_ it (and I left it partly in case someone asked me to return something other than a 500). will remove!

PR #986, opened 2017-04-20
core: grant internal access to local subject name

kr:
A Chain Core TLS config is used by cored peers to talk
to each other, and by corectl to talk to cored. All
those processes have the same identity, so now we
automatically trust the local cert, with the expectation
that the peer will also be using it. This makes
authz misconfiguation impossible as long as the cert and
key files are loaded successfully.

(Motivation copied from
13528a40544824cffd29875e517bb80e1daeec5b.)

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -3,6 +3,7 @@ package core
 
 import (
 	"context"
+	"crypto/x509/pkix"
 	"expvar"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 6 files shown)


kr:
We are seeing `connection refused` when trying to dial the cored process in the java tests. It doesn't seem like that could be related to this but I'll investigate.

kr:
PTAL

boymanjor:
LGTM

PR #987, opened 2017-04-20
encoding/blockchain: add Reader type

jbowens:
Add a Reader type and adapt all Read functions to take that type instead
of a generic io.Reader. Additionally, guard against allocating too much
memory when reading varstring31s. See #978 for background.

This might allow us to make ReadExtensibleString's interface a little
simpler and more direct too, but I think that should be a followup.

Also, because we're just reslicing the buffer in many of the uses, this
change should help performance. For the existing varint benchmarks,
it helps quite a bit by avoiding the need to acquire and copy into a
separate byteReader:

```
name            old time/op  new time/op  delta
ReadVarint31-4   175ns ± 4%    41ns ± 4%  -76.64%  (p=0.001 n=7+7)
ReadVarint63-4   246ns ± 3%    68ns ± 4%  -72.30%  (p=0.001 n=7+6)
```


jbowens:
PTAL

jbowens:
> You don't have to calculate the value of n, copy will do it for you. Also, it's okay to return io.EOF along with some bytes in the last read. So this method can be condensed a little.

Noice, thanks. Updated

jbowens:
just added some quickcheck tests for the encoding/blockchain package too. could also do that in a separate PR.

jbowens:
PTAL

kr:
LGTM

PR #989, opened 2017-04-20
core/authz: add endpoints used by corectl to the internal policy

boymanjor:
This will allow a `corectl` configured with the same TLS certificate as
its remote `cored` to make authenticated requests. This is a part of
the work to make `corectl` use the `cored` API exclusively.


jbowens:
lgtm

PR #992, opened 2017-04-20
net/http/authz: make x509 guard data de/serialization consistent

boymanjor:
Previously, we were serializing the subj name fields as a map while
deserializing into a nested map keyed with "subject". This was causing
unexpected, unauthorized requests during testing.

This logic will change once we introduce a subj name parser and
finalize the guard data structure but this currently impedes testing.


boymanjor:
PTAL

PR #999, opened 2017-04-20
core: add new policy "public," new guard type "any"

tessr:
We are going to gate the `/dashboard` routes behind the auth handlers shortly. When we do that, we need to allow them to be accessible by any client. 

(Dashboard will still have to make a request to cored to actually populate the pages, and cored may respond with an unauth'd error, in which case dashboard will show the "enter access token page." This PR allows dashboard to show that page, instead of returning a 40X to the browser.)

### DIFF
--- a/core/authz.go
+++ b/core/authz.go
@@ -8,6 +8,7 @@ var policies = []string{
 	"network",
 	"monitoring",
 	"internal",
+	"public",
 }
 
 var policyByRoute = map[string][]string{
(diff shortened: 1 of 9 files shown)


tessr:
PTAL

kr:
:eyes:

core/authz.go
```diff
@@ -61,4 +62,25 @@ var policyByRoute = map[string][]string{
 
 	"/raft/join": {"internal"},
 	"/raft/msg":  {"internal"},
+
+	"/":                                       {"public"},
+	"/dashboard":                              {"public"},
+	"/dashboard/core":                         {"public"},
```
kr: Seeing all these makes me think as a todo perhaps we should allow matches like the way ServeMux works, so we could write `/dashboard/` here and get all of them.

net/http/authz/non_loopback_authz.go
```diff
@@ -0,0 +1,9 @@
+//+build !loopback_auth
+
+package authz
+
+func init() {
+	builtinGrants = append(builtinGrants,
+		&Grant{GuardType: "any", Policy: "public"},
```
kr: I suggest moving this to the definition of builtin grants in authorizer.go (and so take it out of loopback_authz.go too).

core/authz.go
tessr: I had this same thought, but I'm not sure how we'd do it safely--since we want the default to be unauthorized, but making "/" public would make the default public. (I guess we could explicitly use regular expressions?)

net/http/authz/loopback_authz.go
```diff
@@ -9,5 +9,6 @@ func init() {
 		&Grant{GuardType: "localhost", Policy: "monitoring"},
 		&Grant{GuardType: "localhost", Policy: "network"},
 		&Grant{GuardType: "localhost", Policy: "internal"},
+		&Grant{GuardType: "any", Policy: "public"},
```
kr: I think this line is unnecessary now?

core/authz.go
kr: Yeah good point.

net/http/authz/loopback_authz.go
tessr: right, thanks!

PR #1000, opened 2017-04-21
core/api: new endpoint "/mockhsm/create-block-key"

boymanjor:
Previously, creating a block signing key was executed by `corectl`
through a database connection. This endpoint helps facilitate our
goal of forcing `corectl` functionality through our ACLs.


boymanjor:
PTAL

kr:
:eyes:

core/run.go
```diff
@@ -121,7 +121,7 @@ func RateLimit(keyFn func(*http.Request) string, burst, perSecond int) RunOption
 // used for Chain Core Developer Edition to expose the configuration UI
 // in the dashboard. API authentication still applies to an unconfigured
 // Chain Core.
-func RunUnconfigured(ctx context.Context, db pg.DB, raftDB *raft.Service, opts ...RunOption) *API {
+func RunUnconfigured(ctx context.Context, db pg.DB, raftDB *raft.Service, opts []RunOption) *API {
```
kr: I suggest leaving this as `...RunOption` so it matches Run.

cmd/corectl/no_mockhsm.go
```diff
@@ -4,6 +4,6 @@ package main
 
 import "chain/database/sql"
 
-func createBlockKeyPair(db *sql.DB, args []string) {
-	fatalln("error: create-block-keypair disabled in prod build")
+func createBlockKeyPair(*sql.DB, []string) {
+	fatalln("error: create-block-keypair disabled in this build")
```
kr: Now that this is protected by cored, I think we can get rid of corectl's no_mockhsm build tag.

boymanjor:
Updated, PTAL

PR #1005, opened 2017-04-21
crypto/x509/name: new package

kr:
This package parses and formats X.509 Distinguished Name
objects according to RFC 4514.

### DIFF
--- a/crypto/x509/name/doc.go
+++ b/crypto/x509/name/doc.go
@@ -1,0 +1,6 @@
+// Package name serializes and parses X.500 Distinguished Names
+// to and from UTF-8 strings.
+// See RFC 4514 "Lightweight Directory Access Protocol (LDAP):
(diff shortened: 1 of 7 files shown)


kr:
```
PACKAGE DOCUMENTATION

package name
    import "."

    Package name serializes and parses X.500 Distinguished Names to and from
    UTF-8 strings. See RFC 4514 "Lightweight Directory Access Protocol
    (LDAP): String Representation of Distinguished Names".
    https://tools.ietf.org/html/rfc4514

FUNCTIONS

func Format(n pkix.Name) (string, error)
    Format formats n as UTF-8 text.

func Parse(dn string) (name pkix.Name, err error)
    Parse parses dn as an X.509 Distinguished Name. The supported fields are
    those of pkix.Name. Unsupported fields are silently discarded.


```

PR #1006, opened 2017-04-21
net/http/authz: match on routes, mux-style

tessr:
We need to be able to pattern match on routes when looking up policies, so that we don't need to define a route-policy mapping for every asset dashboard attempts to return.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509/pkix"
 	"expvar"
 	"fmt"
-	"log"
 	"net/http"
 	"net/http/pprof"
 	"sync"
@@ -326,10 +325,7 @@ func webAssetsHandler(next http.Handler) http.Handler {
 	}))
 	mux.Handle("/", next)
 
-	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		log.Println("handling req: ", req.RequestURI)
-		mux.ServeHTTP(w, req)
-	})
+	return mux
 }
 
 func (a *API) leaderSignHandler(f func(context.Context, *legacy.Block) ([]byte, error)) func(context.Context, *legacy.Block) ([]byte, error) {
--- a/core/api.go
+++ b/core/api.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509/pkix"
 	"expvar"
 	"fmt"
+	"log"
 	"net/http"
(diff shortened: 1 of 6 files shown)


net/http/authz/authorizer.go
```diff
@@ -155,3 +155,57 @@ func (a *Authorizer) grantsByPolicies(policies []string) ([]*Grant, error) {
...
+		if !pathMatch(pattern, route) {
+			continue
+		}
+		if len(policies) == 0 || len(pattern) > n {
+			n = len(pattern)
+			policies = pol
+			pat = pattern
+		}
+	}
+	if pat == "/" {
```
kr: Should this be `pat == ""`? My reading of this is that it would always reject a request for `/` so the redirect to dashboard wouldn't work.

core/api.go
```diff
@@ -265,11 +267,7 @@ func (a *API) authzHandler(mux *http.ServeMux, handler http.Handler) http.Handle
 	auth := authz.NewAuthorizer(a.raftDB, grantPrefix, policyByRoute)
 	auth.GrantInternal(a.internalSubj)
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		// return failure early if this path isn't legit
-		if _, pat := mux.Handler(req); pat != req.URL.Path {
-			errorFormatter.Write(req.Context(), rw, errNotFound)
-			return
-		}
+		// TODO(tessr): check that this path exists; return early if this path isn't legit
```
tessr: This is kind of a bummer, but getting all the pattern matching right here, too, on all the routes (which includes routes which haven't been added to this mux yet) turned out to be such a pain in the butt.

tessr:
This is ready for another pass 💫

core/api.go
```diff
@@ -319,11 +327,7 @@ func webAssetsHandler(next http.Handler) http.Handler {
 	mux.Handle("/", next)
 
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		if req.URL.Path == "/" {
-			http.Redirect(w, req, "/dashboard/", http.StatusFound)
-			return
-		}
-
+		log.Println("handling req: ", req.RequestURI)
```
kr: Debugging output?

Maybe now webAssetsHandler should end with `return mux`?

kr:
LGTM after one possible tweak

PR #1007, opened 2017-04-21
core: add tests for authz grants/policies

erykwalder:
Adds basic tests for authz grants and policies. Testing is done at the
http level.

### DIFF
--- a/core/authz_test.go
+++ b/core/authz_test.go
@@ -132,7 +132,7 @@ func TestAuthz(t *testing.T) {
 
 	for path, policyMap := range pathTokenMap {
 		for policy, want := range policyMap {
-			got := tryEndpoint(t, server.URL, path, tokens[policy])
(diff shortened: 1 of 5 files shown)


core/authz_test.go
```diff
@@ -0,0 +1,161 @@
...
+		for policy, want := range policyMap {
+			got := tryEndpoint(t, server.URL, path, tokens[policy])
+			if got != want {
+				t.Errorf("auth(%s, %s) = %t want %t", path, policy, got, want)
+			}
+		}
+	}
+}
+
+func tryEndpoint(t testing.TB, baseURL, path string, token *accesstoken.Token) bool {
```
kr: Nit: can we call it tryRPC instead? (I am still fighting my (losing) battle against calling these things "endpoints".)

kr:
LGTM

PR #1011, opened 2017-04-21
net/http/authz: standardize X.509 matching

kr:
Future work: update /create-authorization-grant to verify that the grant data is well-formed.

### DIFF
--- a/net/http/authz/tls.go
+++ b/net/http/authz/tls.go
@@ -1,0 +1,72 @@
+package authz
+
+import (
+	"crypto/x509/pkix"
+	"encoding/json"
+)
+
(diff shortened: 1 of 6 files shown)


kr:
PTAL

net/http/authz/tls.go
```diff
@@ -0,0 +1,72 @@
+package authz
+
+import (
+	"crypto/x509/pkix"
+	"encoding/json"
+)
+
+// Same as crypto/x509/pkix.Name but with JSON tags.
+type pkixDN struct {
+	Country            []string `json:"C"`
```
kr: Oh I should put omitempty on all these.

PR #1012, opened 2017-04-21
core: reject bad fields in x509 guard data

kr:


core/grants.go
```diff
@@ -31,11 +31,19 @@ func (a *API) createGrant(ctx context.Context, x apiGrant) error {
...
+				return errors.WithDetail(httpjson.ErrBadRequest, "bad subject field "+k)
+			}
+		}
 	}
 
 	// NOTE: package json produces consistent serialization output,
 	// effectively an ad hoc canonical form. We rely on this to
 	// grant data for equality.
+	// TODO(kr): x509 subject field names are case-insensitive,
+	// so we should do our equality comparisons accordingly.
```
tessr: Is this todo still relevant?

kr: Yes I think so. Don't we currently use `bytes.Equal` to check if two GrantData fields are equal? Those fields are serialized from the user-supplied json directly, so there's no case-folding in there yet.

tessr: oh, yes, yes we do.

tessr:
LGTM

PR #1013, opened 2017-04-21
protocol/bc/legacy: remove TxData db methods

jbowens:
Remove TxData.Scan and TxData.Value. We never store individual
blockchain transactions in the database any more.


kr:
LGTM

PR #1015, opened 2017-04-21
core/config: migrate access tokens to grants 

tessr:
When a core migrates from 1.1 to 1.2, it will need to turn its access tokens into grants. When we start a core that needs to migrate its config table, we also migrate these access tokens.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -71,7 +71,9 @@ func Load(ctx context.Context, db pg.DB, rDB *raft.Service) (*Config, error) {
 			return nil, errors.Wrap(err)
 		}
 
-		// If we were able to find a config in Postgres, store it in Raft
+		// If we were able to find a config in Postgres, store it in Raft.
+		// This also means that we are running this core with raft for the first time
+		// which means that we will also migrate access tokens.
 		if config != nil {
 			val, err := proto.Marshal(config)
 			if err != nil {
@@ -86,8 +88,8 @@ func Load(ctx context.Context, db pg.DB, rDB *raft.Service) (*Config, error) {
 				// If we got this far but failed to delete from PG, it's really NBD. Just
 				// log the failure and carry on.
 				log.Error(ctx, err, "failed to delete config from postgres")
-
 			}
+			err = migrateAccessTokens(ctx, db, rDB)
 			return config, nil
 		}
 		return nil, nil
@@ -280,4 +282,22 @@ func tryGenerator(ctx context.Context, url, accessToken, blockchainID string) er
 	}
 
 	return nil
+}
+
+// this almost certainly should live in another package
+func migrateAccessTokens(ctx context.Context, db *pg.DB, rDB *raft.Service) error {
+	const q = `SELECT id, type, sort_id, created FROM access_tokens`
+	var tokens []*accesstoken.Token
+	err := pg.ForQueryRows(ctx, cs.DB, q, func(id string, maybeType sql.NullString, sortID string, created time.Time) {
+		t := Token{
+			ID:      id,
+			Created: created,
+			Type:    maybeType.String,
+			sortID:  sortID,
+		}
+		tokens = append(tokens, &t)
+	})
+	if err != nil {
+		return nil, "", errors.Wrap(err)
+	}
 }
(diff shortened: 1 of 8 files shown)


tessr:
PTAL

core/config/config.go
```diff
@@ -86,8 +92,8 @@ func Load(ctx context.Context, db pg.DB, rDB *raft.Service) (*Config, error) {
 				// If we got this far but failed to delete from PG, it's really NBD. Just
 				// log the failure and carry on.
 				log.Error(ctx, err, "failed to delete config from postgres")
-
 			}
+			err = migrateAccessTokens(ctx, db, rDB)
```
jeffomatic: Should we bubble this error up?

kr:
:eyes:

core/config/config.go
```diff
@@ -281,3 +287,51 @@ func tryGenerator(ctx context.Context, url, accessToken, blockchainID string) er
...
+		tokens = append(tokens, t)
+	})
+
+	for _, token := range tokens {
+		data := map[string]interface{}{
+			"id": token.ID,
+		}
+		guardData, err := json.Marshal(data)
+		if err != nil {
+			return errors.Wrap(err)
```
jeffomatic: I suspect this is a `panic(err) // should never get here` situation?

net/http/authz/grant.go
```diff
@@ -1,4 +1,75 @@
...
+
+	"chain/database/raft"
+	"chain/errors"
+)
+
 // Generate code for the Grant and GrantList types.
 //go:generate protoc -I. -I$CHAIN/.. --go_out=. grant.proto
+
+// StoreGrant stores a new grant in the provided raft store
+func StoreGrant(ctx context.Context, raftDB *raft.Service, grant Grant, grantPrefix string) error {
```
jeffomatic: Didn't scrutinize this too closely...is this just a copy/paste? If not I can review more thoroughly.

tessr: It's a copy paste :)

core/config/config.go
tessr: Oh, yes, this was a question I wanted to address. If we fail to migrate the tokens, it means that they can't be used on this core. So what should we do here?

core/config/config.go
```diff
@@ -281,3 +287,51 @@ func tryGenerator(ctx context.Context, url, accessToken, blockchainID string) er
...
+				GuardType: "access_token",
+				GuardData: guardData,
+				Policy:    "client-readwrite",
+				CreatedAt: token.Created.Format(time.RFC3339),
+			}
+		case "network":
+			grant = authz.Grant{
+				GuardType: "access_token",
+				GuardData: guardData,
+				Policy:    "network",
```
kr: Nit: this switch could be a little more concise:

```go
grant := authz.Grant{
	GuardType: "access_token",
	GuardData: guardData,
	CreatedAt: token.Created.Format(time.RFC3339),
}
switch token.Type {
case "client":
	grant.Policy = "client-readwrite"
case "network":
	grant.Policy = "network
}
```

or even

```go
policies := map[string]string{
	"client":  "client-readwrite",
	"network": "network",
}
grant := authz.Grant{
	GuardType: "access_token",
	GuardData: guardData,
	Policy:    policies[token.Type],
	CreatedAt: token.Created.Format(time.RFC3339),
}

File: core/config/config.go (hunk 1 above)
Comment (kr) [inline core/config/config.go]: This error strikes me as both 1) unlikely and 2) serious. So I vote that if it happens we should crash and get the customer to call us to help with the migration. They can continue running the old version in the meantime.

Comment (tessr) [inline core/config/config.go]: Alright, I added a panic here - do you think we need to do anything else? (When you say, "get the customer to call us," you mean that people will call if we crash, not that we should print a sentence here like, "Please call us." Right?)

Comment (kr) [inline core/config/config.go]: Heh, yeah. Just a panic sounds good.

File: core/config/config.go [hunk 5]
```diff
@@ -281,3 +290,45 @@ func tryGenerator(ctx context.Context, url, accessToken, blockchainID string) er
...
+		grant := authz.Grant{
+			GuardType: "access_token",
+			GuardData: guardData,
+			CreatedAt: token.Created.Format(time.RFC3339),
+		}
+		switch token.Type {
+		case "client":
+			grant.Policy = "client-readwrite"
+		case "network":
+			grant.Policy = "netowrk"
```
kr: `netowrk`

kr:
LGTM after typo fix

core/config/config.go
tessr: AHHH

tessr: This prompted me to run all the code again 😅

PR #1016, opened 2017-04-22
docs: update block signer guide with latest corectl instructions

dominic:


docs/core/learn-more/configure-block-signers.md
```diff
@@ -60,17 +60,27 @@ Block signer Chain Core URL with network token: https://foo:25f658b749f154a790c8
...
+corectl config-generator -k <generator-pubkey> <quorum> <signer1-pubkey> <signer1-url-with-network-token>
 ```
 
 ```bash
-corectl config-generator -s 2 cce1791bf3d8bb5e506ec7159bad6a696740712197894336c027dec9fbfb9313 https://foo:25f658b749f154a790c8a3aeb57ea98968f51a991c4771fb072fcbb2fa63b6f7@<signer-host>:<signer-port>
+corectl config-generator -k 45ad1f1617d4c6fb8ae5119c524c6266595f0f551c5210dd9f5892b4b39f011f 2 cce1791bf3d8bb5e506ec7159bad6a696740712197894336c027dec9fbfb9313 https://foo:25f658b749f154a790c8a3aeb57ea98968f51a991c4771fb072fcbb2fa63b6f7@<signer-host>:<signer-port>
 ```
 
-The `-s` flag means this generator is also a block signer. The quorum value, 2, means both signatures are required on every block.
+The `-k` flag means this generator is also a block signer, and will provide its
```
dominic: One q here is can you have a generator that _isn't_ a block signer?

kr: You can!

dominic: Okay great. It was a little ambiguous with https://github.com/chain/chain/blob/main/cmd/corectl/doc.go#L43, which says it will be "a participant (not a generator or a signer)" without `-k`. I'll update that in my other docs branch.

PR #1021, opened 2017-04-24
desktop/windows: override home dir

kr:
As of 338f54eb2dda298fcf3d6d0baa2c1f0d40c088b1, the
default home dir for cored is simply $HOME/.chaincore.
For a windows desktop install, it's better to use
LOCALAPPDATA.
    
This change also moves the Postgres data dir inside the
Chain Core home dir and improves support for older
versions of Windows.

### DIFF
--- a/desktop/windows/ChainMgr/ChainMgr.go
+++ b/desktop/windows/ChainMgr/ChainMgr.go
@@ -17,15 +17,11 @@ import (
 // `GOOS=windows GOARCH=amd64 go build chain/desktop/windows/ChainMgr`
 
 const (
-
 	// The Chain Core executable itself.
-	chainCoreExe = `C:/Program Files (x86)/Chain/cored.exe`
-
-	// Data directory for Postgres to store all of its stuff for a specific db.
-	pgDataDir = `C:/Program Files (x86)/Chain/data`
+	chainCoreExe = `C:\Program Files (x86)\Chain\cored.exe`
 
 	// Path for all the Postgres binaries.
-	pg = `C:/Program Files (x86)/Chain/Postgres/bin/`
+	pg = `C:\Program Files (x86)\Chain\Postgres\bin\`
 
 	// Port this db will listen on. Also the year I started kindergarten.
 	pgPort = "1998"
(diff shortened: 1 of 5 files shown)


kr:
Note, I have not tested this. I will work on that next.

desktop/windows/ChainMgr/ChainMgr.go
```diff
@@ -35,6 +35,9 @@ const (
 	dbName = "core"
 )
 
+// TODO(kr): put everything else in here too, e.g. postgres data dir.
```
tessr: Ha, my one suggestion was going to be that you move the postgres data dir too.

kr:
Tested and updated. PTAL

desktop/windows/ChainMgr/ChainMgr.go
```diff
@@ -35,13 +31,27 @@ const (
 	dbName = "core"
 )
 
+var home = oneOf(
+	os.Getenv("CHAIN_CORE_HOME"),
+	filepath.Join(appData(), `Chain Core`),
+)
+
+var (
+	pgDataDir = filepath.Join(home, "Postgres")
```
kr: Should this be `pgdata`? 🚲 🏠

desktop/windows/ChainMgr/ChainMgr.go
```diff
@@ -17,15 +17,11 @@ import (
...
 	// The Chain Core executable itself.
-	chainCoreExe = `C:/Program Files (x86)/Chain/cored.exe`
-
-	// Data directory for Postgres to store all of its stuff for a specific db.
-	pgDataDir = `C:/Program Files (x86)/Chain/data`
+	chainCoreExe = `C:\Program Files (x86)\Chain\cored.exe`
 
 	// Path for all the Postgres binaries.
-	pg = `C:/Program Files (x86)/Chain/Postgres/bin/`
+	pg = `C:\Program Files (x86)\Chain\Postgres\bin\`
```
tessr: In retrospect, maybe we should build these paths with `filepath.Join`.

desktop/windows/ChainMgr/ChainMgr.go
tessr: Sure.

PR #1022, opened 2017-04-24
protocol: delete spent prevouts

### LINKED ISSUE #1018: protocol: spent outputs not removed from state tree
Kudos to @boymanjor for finding this one.

Here's a test with two different transactions, each with a retirement at output index 1. `tx.OutputID(1)` provides the same hash for both: https://gist.github.com/jbowens/434b093c129ae8407e84be32573477c6

I'm not sure if this is a bug in the protocol/bc/legacy mapping to the transaction graph or a bug in protocol/bc?

jbowens:
Remove spent prevouts from the state tree. Previously, we would
remove the prevouts but then immediately add them back.

Fix #1018

### DIFF
--- a/protocol/bc/tx.go
+++ b/protocol/bc/tx.go
@@ -15,9 +15,9 @@ type Tx struct {
 	InputIDs []Hash // 1:1 correspondence with TxData.Inputs
 
 	// IDs of reachable entries of various kinds
-	NonceIDs       []Hash
-	SpentOutputIDs []Hash
-	OutputIDs      []Hash
+	NonceIDs        []Hash
+	SpentOutputIDs  []Hash
+	ResultOutputIDs []Hash
 }
 
 func (tx *Tx) SigHash(n uint32) (hash Hash) {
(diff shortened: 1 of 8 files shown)


jbowens:
PTAL

bobg:
:eyes:

protocol/bc/tx.go
```diff
@@ -5,7 +5,7 @@ import (
 	"chain/errors"
 )
 
-// TxEntries is a wrapper for the entries-based representation of a
+// Tx is a wrapper for the entries-based representation of a
 // transaction.  When we no longer need the legacy Tx and TxData
```
bobg: Might as well remove the "this will be renamed" part of the comment too.

protocol/state/snapshot_test.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Apply the spend transaction.
+	err = snap.ApplyTx(tx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if snap.Tree.Contains(spentOutputID.Bytes()) {
```
bobg: Did we really not already have a test for this??

bobg:
LGTM with one small suggestion.

PR #1023, opened 2017-04-24
protocol/bc/legacy: avoid erroring in MapTx

### LINKED ISSUE #967: protocol/bc/legacy: MapTx errors cause CH000
Currently we map legacy transactions into txgraph transactions when
deserializing them. `legacy.MapTx` can return an error on nonce-less
issuances (and maybe other conditions?). Submitting or building from
a base invalid transaction w/ a nonce-less issuance errors during
request parsing before reaching validation logic. The error gets
formatted as a CH000 internal server error.

For reproducing, here is a test transaction with a nonce-less issuance:

```
0701020000000101280000a9b2b6c5394888ab5396f583ae484b8459486b14268e2bef1b637440335eb6c180a094a58d1d05696e7075742903deff1d4319d67baa10a6d26c1fea9c3e8d30e33474efee1a610a9bb49d758d000101010103010203010129a9b2b6c5394888ab5396f583ae484b8459486b14268e2bef1b637440335eb6c180a094a58d1d010101066f7574707574000869737375616e6365
```

This is a pretty minor bug: mostly cosmetic, I think? Not sure what the
best way to handle it is. If possible, making MapTx error-free and
performing all of the validation logic during validation would solve it.

jbowens:
Currently we map legacy transactions into txgraph transactions when
deserializing them. Because MapTx was erroring on nonce-less issuances,
submitting an invalid transaction w/ a nonce-less issuance would error
during unmarshalling before reaching validation logic. The error
formatter would format this as an internal server error or a bad request
depending on the endpoint.

Removes the error return value from MapTx altogether.

Fix #967


jbowens:
PTAL. I'd like to make MapTx not return an error, but maybe that should be a followup?

protocol/bc/legacy/map.go
```diff
@@ -143,17 +145,10 @@ func mapTx(tx *TxData) (headerID bc.Hash, hdr *bc.TxHeader, entryMap map[bc.Hash
 
 			var (
 				anchorID    bc.Hash
-				setAnchored func(*bc.Hash)
+				setAnchored = func(*bc.Hash) {}
```
bobg: This feels a little hackish to me. Slight preference for testing `setAnchored != nil` at the point where it's called.

bobg:
Removing the error return from `MapTx` sgtm. Please add a test that calling `MapTx` on a nonceless issuance produces an object that fails validation.

protocol/bc/legacy/map.go
jbowens: It's a pretty common way to avoid nil checks in Go. Ex: [database/sql example](https://github.com/golang/go/blob/release-branch.go1.8/src/database/sql/sql.go#L1728), [crypto/tls example](https://github.com/golang/go/blob/release-branch.go1.8/src/crypto/tls/tls.go#L194)

jbowens:
> Please add a test that calling `MapTx` on a nonceless issuance produces an object that fails validation.

Will add once #1045 lands.

protocol/bc/legacy/map.go
bobg: Oh OK, thanks. #TIL

jbowens:
PTAL

bobg:
LGTM

PR #1026, opened 2017-04-24
protocol/bc/legacy: prohibit trailing garbage

jbowens:
Prohibit trailing garbage when parsing a block or a transaction.


protocol/bc/legacy/block.go
```diff
@@ -50,7 +51,18 @@ func (b *Block) UnmarshalText(text []byte) error {
...
+
+	r := blockchain.NewReader(decoded)
+	err = b.readFrom(r)
+	if err != nil {
+		return err
+	}
+
+	// Disallow trailing garbage.
+	if trailing := r.Remaining(); len(trailing) > 0 {
+		return fmt.Errorf("expected EOF, got %x", trailing)
```
kr: Should we include the actual trailing data here? Or just `trailing garbage (%d bytes)`? (In that case I would suggest a method called Len instead of Remaining.)

PR #1029, opened 2017-04-24
core: authenticate raft endpoints 

tessr:


tessr:
PTAL

core/api.go
```diff
@@ -250,24 +250,36 @@ type page struct {
...
+	authenticator := authn.NewAPI(accessTokens, networkRPCPrefix)
+	authorizer := authz.NewAuthorizer(rDB, grantPrefix, policyByRoute)
+
+	var (
+		x509Cert *x509.Certificate
+		err      error
+	)
+	if tlsConfig != nil {
+		// TODO(kr): set Leaf in TLSConfig and use that here.
+		// TODO(tessr): ^ do I keep this TODO?
```
tessr: (Should I keep this todo?)

kr:
:eyes:

cmd/cored/main.go
```diff
@@ -146,6 +147,22 @@ func main() {
...
+		chainlog.Fatalkv(ctx, chainlog.KeyError, err)
+	}
+	db.SetMaxOpenConns(*maxDBConns)
+	db.SetMaxIdleConns(*maxDBConns)
+
+	err = migrate.Run(db)
+	if err != nil {
+		chainlog.Fatalkv(ctx, chainlog.KeyError, err)
+	}
+	resetIfAllowedAndRequested(db, raftDB)
```
kr: I think we can't do this (reset) until after we're listening for raft messages.

core/api.go
```diff
@@ -250,24 +250,36 @@ type page struct {
 	LastPage bool         `json:"last_page"`
 }
 
-func (a *API) authnHandler(handler http.Handler) http.Handler {
-	auth := authn.NewAPI(a.accessTokens, networkRPCPrefix)
+func AuthHandler(mux *http.ServeMux, handler http.Handler, rDB *raft.Service, accessTokens *accesstoken.CredentialStore, tlsConfig *tls.Config) http.Handler {
```
kr: It looks like mux is unused.

core/api.go
kr: Actually, this might be a good opportunity to be a little more specific with the requirements for this function. Instead of taking a whole *tls.Config, maybe it should just take the pkix.Name to trust `internalDN pkix.Name`. What do you think?

core/api.go
```diff
@@ -250,24 +250,36 @@ type page struct {
...
-}
 
-func (a *API) authzHandler(mux *http.ServeMux, handler http.Handler) http.Handler {
-	auth := authz.NewAuthorizer(a.raftDB, grantPrefix, policyByRoute)
-	auth.GrantInternal(a.internalSubj)
-	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		// TODO(tessr): check that this path exists; return early if this path isn't legit
-		err := auth.Authorize(req)
+		if x509Cert != nil {
+			authorizer.GrantInternal(x509Cert.Subject)
```
kr: It looks like this is going to run on every request. We should do it outside the handler closure.

net/http/authn/authn.go
```diff
@@ -62,7 +62,9 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 
 // checks the request for a valid client cert list.
 // If found, it is added to the request's context.
-func (a *API) certAuthn(req *http.Request) context.Context {
+// This function is exported so it can be used when
+// a full API is not available.
+func CertAuthn(req *http.Request) context.Context {
```
kr: Does this need to be exported?

tessr:
This is ready for another pass, PTAL!

kr:
:eyes:

cmd/cored/main.go
```diff
@@ -138,6 +141,17 @@ func main() {
 	if err != nil {
 		chainlog.Fatalkv(ctx, chainlog.KeyError, err)
 	}
+	var (
+		x509Cert   *x509.Certificate
+		internalDN *pkix.Name
+	)
+	if tlsConfig != nil {
+		x509Cert, err = x509.ParseCertificate(tlsConfig.Certificates[0].Certificate[0])
```
kr: Nit: I think this can use `:=` and then we don't need the var declaration for this variable above.

kr:
LGTM

PR #1030, opened 2017-04-24
core: fix x509 guard data validation

boymanjor:
Previously, the check for bad subject fields was iterating through
the top-level "guard_data" object. We expect these fields to be
wrapped in a nested "subject" map.


boymanjor:
PTAL

kr:
Nit: subject should be `core:` (there is no "grants" package or directory)

PR #1031, opened 2017-04-24
dashboard: support limited subset of x509 subject fields

dominic:


dashboard/src/features/accessControl/actions.js
```diff
@@ -73,14 +74,27 @@ export default {
...
-      guardData: {subject: {}},
+      guardData: {},
     }
 
     for (let index in data.subject) {
       const field = data.subject[index]
-      body.guardData.subject[field.key] = field.value
+
+      if (fieldInfo[field.key].array) {
+        const values = body.guardData[field.key] || []
```
jeffomatic: I would like for the SDK to enforce the array-ness here. It is okay for the dashboard to repeat the work, but it is something the SDK should do in general.

dashboard/src/features/accessControl/constants.js
```diff
@@ -20,3 +20,15 @@ export const policyOptions = [
     hint: 'Access to monitoring-specific endpoints'
   },
 ]
+
+export const subjectFieldOptions = [
```
jeffomatic: All-caps for `value` is a little more consistent to the way it will be used in the rest of the codebase.

dashboard/src/features/accessControl/actions.js
dominic: Agreed; the Dashboard needs to submit a valid format to the SDK. I didn't want to tie the two PRs together.

dashboard/src/features/accessControl/constants.js
dominic: Okay, real talk. When I put it as `CN`, the SDK converted the key to `_c_n`, which is no good. Should our snake-caser ignore strings that are all uppercase?

jeffomatic: Gah gnarly. Let's just stick with lower-case for now, since it will work.

jeffomatic:
According to https://github.com/chain/chain/pull/1030/files we actually *do* need a top-level `subject` key. I will update before landing.

dashboard/src/features/accessControl/constants.js
jeffomatic: Note for posterity: Node SDK has been updated so it doesn't snakeize all-caps strings.

dashboard/src/features/accessControl/actions.js
jeffomatic: Note for posterity: the Node SDK was updated to deal with array attributes correctly.

PR #1032, opened 2017-04-24
docs: add tag update docs

chrisgarvin:


docs/core/examples/java/Assets.java
```diff
@@ -7,6 +7,7 @@
 class Assets {
   public static void main(String[] args) throws Exception {
     Client client = new Client();
+    Map<String, Object> tagUpdate;
```
jeffomatic: @chrisgarvin This declaration should go inline, on line 77 below. Otherwise, it will be unclear what `tagUpdate` is in the snippet itself.

jeffomatic: Ditto for the accounts example

Issue #988, opened 2017-04-20
core/generator: data race in getAndAddBlockSignatures


jbowens:
Running `go test -race` in chain/core/generator reveals a data race in the TestGetAndAddBlockSignatures test.

```
==================
WARNING: DATA RACE
Read at 0x00c4201b4558 by goroutine 29:
  chain/protocol/bc/legacy.mapBlockHeader()
      /Users/jackson/src/chain/protocol/bc/legacy/map.go:296 +0x3ab
  chain/protocol/bc/legacy.(*BlockHeader).Hash()
      /Users/jackson/src/chain/protocol/bc/legacy/block_header.go:67 +0x3c
  chain/core/generator.testSigner.SignBlock()
      /Users/jackson/src/chain/core/generator/generator_test.go:115 +0x52
  chain/core/generator.(*testSigner).SignBlock()
      <autogenerated>:2 +0xc7
  chain/core/generator.getSig()
      /Users/jackson/src/chain/core/generator/block.go:141 +0x80

Previous write at 0x00c4201b4558 by goroutine 27:
  chain/core/generator.(*Generator).getAndAddBlockSignatures()
      /Users/jackson/src/chain/core/generator/block.go:126 +0x115a
  chain/core/generator.TestGetAndAddBlockSignatures()
      /Users/jackson/src/chain/core/generator/generator_test.go:80 +0x5fc
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:657 +0x107

Goroutine 29 (running) created at:
  chain/core/generator.(*Generator).getAndAddBlockSignatures()
      /Users/jackson/src/chain/core/generator/block.go:105 +0x38b
  chain/core/generator.TestGetAndAddBlockSignatures()
      /Users/jackson/src/chain/core/generator/generator_test.go:80 +0x5fc
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:657 +0x107

Goroutine 27 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:697 +0x543
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:882 +0xaa
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:657 +0x107
  testing.runTests()
      /usr/local/go/src/testing/testing.go:888 +0x4e0
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:822 +0x1c3
  main.main()
[...]

PR #1033, opened 2017-04-24
core/generator: marshal block once

jbowens:
When requesting signatures, marshal the block once. Besides avoiding the
extra work of marshalling multiple times, this resolves the data race in 
https://github.com/chain/chain/issues/988#issuecomment-296818634

Tested locally by setting up and using a 2-of-2 network.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -439,11 +439,8 @@ type remoteSigner struct {
 	Key    ed25519.PublicKey
 }
 
-func (s *remoteSigner) SignBlock(ctx context.Context, b *legacy.Block) (signature []byte, err error) {
-	// TODO(kr): We might end up serializing b multiple
-	// times in multiple calls to different remoteSigners.
-	// Maybe optimize that if it makes a difference.
-	err = s.Client.Call(ctx, "/rpc/signer/sign-block", b, &signature)
+func (s *remoteSigner) SignBlock(ctx context.Context, marshalledBlock []byte) (signature []byte, err error) {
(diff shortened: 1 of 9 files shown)


cmd/cored/main.go
```diff
@@ -419,11 +419,8 @@ type remoteSigner struct {
 	Key    ed25519.PublicKey
 }
 
-func (s *remoteSigner) SignBlock(ctx context.Context, b *legacy.Block) (signature []byte, err error) {
-	// TODO(kr): We might end up serializing b multiple
-	// times in multiple calls to different remoteSigners.
-	// Maybe optimize that if it makes a difference.
-	err = s.Client.Call(ctx, "/rpc/signer/sign-block", b, &signature)
+func (s *remoteSigner) SignBlock(ctx context.Context, marshalledBlock []byte) (signature []byte, err error) {
+	err = s.Client.Call(ctx, "/rpc/signer/sign-block", string(marshalledBlock), &signature)
```
kr: This seems like it would change the wire format. Should this be json.RawMessage?

jbowens: Ah, I think this function signature is confusing. Maybe `marshalledBlock` should be a string?

I think this should only change the Go interfaces. Previously, Call would json.Marshal the block, calling Block.MarshalText. The resulting request body would be a json string contianing the hex-encoded bytes:
```
"abcdef12..."
```

Now, SignBlock takes the output of Block.MarshalText directly. I think it can't be a json.RawMessage because it doesn't have the json string quotes yet.

kr: Okay makes sense. I was confused about the old format.

PR #1036, opened 2017-04-25
core/blocksigner: always lock block height

jbowens:
Always lock the block height, even on the generator.
If the generator is a signer, it calls SignBlock directly instead of
ValidateAndSignBlock. Since the generator created the block, it's
safe to skip the validation. Previously, calling SignBlock directly
also skipped locking the block height in the `signed_blocks` table.

This change modifies the generator flow to also lock the block height.
The generator _should_ only generate one block at each height. Locking
the block height can help guarantee we don't violate that.

Also, currently on main there's a bug where anyone with network access
could call the sign-block RPC endpoint and get the generator to sign its
own alternative, valid next block. With this change at least that scenario
would be detected.


jbowens:
PTAL

PR #1037, opened 2017-04-25
core: return grant in create-grant API response

jeffomatic:
This makes the `/create-authorization-grant` RPC consistent in behavior to all other `/create-`-prefixed endpoints.

### DIFF
--- a/core/grants.go
+++ b/core/grants.go
@@ -23,20 +23,20 @@ type apiGrant struct {
 // ErrMissingTokenID is returned when a token does not exist.
 var errMissingTokenID = errors.New("id does not exist")
 
-func (a *API) createGrant(ctx context.Context, x apiGrant) error {
+func (a *API) createGrant(ctx context.Context, x apiGrant) (*authz.Grant, error) {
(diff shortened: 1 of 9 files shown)


jeffomatic:
slight revision: removed temp var in grant.go

core/grants.go
```diff
@@ -47,7 +47,7 @@ func (a *API) createGrant(ctx context.Context, x apiGrant) error {
 	// so we should do our equality comparisons accordingly.
 	guardData, err := json.Marshal(x.GuardData)
 	if err != nil {
-		return errors.Wrap(err)
+		return nil, errors.Wrap(err)
 	}
 
 	g := authz.Grant{
```
tessr: Why not just return `g` if `err == nil` at the end of this, instead of modifying `StoreGrant`?

jeffomatic: I want to return the grant including the `CreatedAt` timestamp, which StoreGrant generates or retrieves.

tessr: Ah, got it 👍

tessr:
LGTM

PR #1039, opened 2017-04-25
sdk/ruby: fix path joining

jeffomatic:
This is a backport of #502, which fixed URL joining for use with non-MockHSM signerd URLs.


jeffomatic:
PTAL

kr:
LGTM

PR #1042, opened 2017-04-25
cmd/corectl: grant policy along with creating token

kr:
This makes the implementation match the instructions.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -212,21 +212,23 @@ func createToken(client *rpc.Client, args []string) {
 	}
 	fmt.Println(tok.Token)
 
-	if len(args) == 2 || *flagNet {
-		req := struct{ Guard_Type, Guard_Data, Policy string }{
-			"access_token",
-			tok.ID,
-			"network", // in case of *flagNet
-		}
-		if len(args) == 2 {
-			req.Policy = args[1]
-		}
-		err = client.Call(context.Background(), "/create-authorization-grant", req, nil)
-		if err != nil {
-			fatalln("rpc error:", err)
-		}
-	} else if *flagNet {
+	req := struct{ Guard_Type, Guard_Data, Policy string }{
+		Guard_Type: "access_token",
+		Guard_Data: tok.ID,
+	}
+	switch {
+	case len(args) == 2:
+		req.Policy = args[1]
+	case *flagNet:
+		req.Policy = "network"
 		fmt.Fprintln(os.Stderr, "warning: the network flag is deprecated")
+	default:
+		req.Policy = "client-readwrite"
+		fmt.Fprintln(os.Stderr, "warning: implicit policy name is deprecated")
+	}
+	err = client.Call(context.Background(), "/create-authorization-grant", req, nil)
+	if err != nil {
+		fatalln("rpc error:", err)
 	}
 }
 
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -190,7 +190,7 @@ func createBlockKeyPair(client *rpc.Client, args []string) {
 }
 
 func createToken(client *rpc.Client, args []string) {
-	const usage = "usage: corectl create-token [-net] [name]"
+	const usage = "usage: corectl create-token [-net] [name] [policy]"
 	var flags flag.FlagSet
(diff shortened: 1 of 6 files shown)


kr:
There are four cases, and here's how I chose to treat them:

1. Just `-net` is given — create a `network` grant (same behavior before).
2. Just `[policy]` is given — create a grant for the named policy (the whole point of this PR).
3. Both are given — error.
4. Neither is given — create *only* the token, with no grant.

I'm not entirely sure that's what we want for case 4. Should it keep the old behavior of adding a client-readwrite grant in that case?

kr:
Anyway, PTAL

dominic:
Can you update the docs in https://github.com/chain/chain/blob/main/docs/core/reference/corectl.md#create-token too?

dominic:
re: case 4, I was thinking about that too. It'd be nice if existing workflows kept working for this version; would it be worth adding a deprecation notice there too?

kr:
> re: case 4, I was thinking about that too. It'd be nice if existing workflows kept working for this version; would it be worth adding a deprecation notice there too?

Sounds good. I'll change the behavior and add a warning.

kr:
Actually, tbh supporting the old behavior isn't costing us anything. Maybe we should *not* print a warning, even for `-net`.

kr:
And maybe tone down the deprecation notice in the usage text?

We can support this old interface indefinitely as far as I'm concerned.

kr:
Updated. PTAL

cmd/corectl/main.go
```diff
@@ -200,25 +200,35 @@ func createToken(client *rpc.Client, args []string) {
...
 	err := client.Call(context.Background(), "/create-access-token", req, &tok)
 	if err != nil {
 		fatalln("rpc error:", err)
 	}
 	fmt.Println(tok.Token)
 
-	if *flagNet {
+	req := struct{ Guard_Type, Guard_Data, Policy string }{
+		Guard_Type: "access_token",
+		Guard_Data: tok.ID,
```
kr: This needs to be json `{"id": id}`.

kr:
Fixed the bug @jeffomatic pointed out. PTAL.

docs/core/reference/corectl.md
```diff
@@ -126,13 +126,23 @@ corectl create-block-keypair
...
-otherwise it will create a client token.
+Generates a new access token with the given name
+and grants it access to the named policy.
 
 ```
-corectl create-token [-net] [name]
+corectl create-token [-net] [name] [policy]
 ```
 
+If no policy is given,
```
tessr: nit: you could probably make these lines longer

cmd/corectl/main.go
```diff
@@ -200,25 +200,39 @@ func createToken(client *rpc.Client, args []string) {
...
 	var tok accesstoken.Token
 	err := client.Call(context.Background(), "/create-access-token", req, &tok)
 	if err != nil {
 		fatalln("rpc error:", err)
 	}
 	fmt.Println(tok.Token)
 
-	if *flagNet {
+	req := struct {
+		Guard_Type string
```
tessr: nit: I'd rather we use json tags here, instead of underscores in field names. (At least I assume that's what the underscores are for?)

cmd/corectl/main.go
```diff
@@ -200,25 +200,39 @@ func createToken(client *rpc.Client, args []string) {
...
-		ID, Type string
-	}{
-		ID:   args[0],
-		Type: map[bool]string{true: "network", false: "client"}[*flagNet],
-	}
+	req := struct{ ID string }{args[0]}
 	var tok accesstoken.Token
 	err := client.Call(context.Background(), "/create-access-token", req, &tok)
 	if err != nil {
 		fatalln("rpc error:", err)
```
tessr: Could you also add a TODO here about atomicity between postgres and raft db? We have the same problem that we've seen before, where a token could get created without a grant.

tessr:
I approved this but wercker says there are some things to fix.

docs/core/reference/corectl.md
kr: I was trying to do http://rhodesmill.org/brandon/2012/one-sentence-per-line/ but I won't insist.

tessr: ehhhhhhhhhhhhh

PR #1043, opened 2017-04-25
cmd/corectl: add commands grant and revoke

kr:


docs/core/reference/corectl.md
```diff
@@ -141,3 +141,45 @@ other configuration will be deleted.
 ```
 corectl reset
 ```
+
+### `grant`
```
dominic: There's a table of contents at the top of the file to add these functions to as well.

docs/core/reference/corectl.md
```diff
@@ -141,3 +141,45 @@ other configuration will be deleted.
...
+
+Command grant grants access to a policy
+for the described credentials.
+
+```
+corectl grant [-cn name] [-ou name] [-t token] [policy]"
+```
+
+Flag -cn grants access to any valid X.509 certificate
```
dominic: It'd be nice to match the flag formatting of previous comments, or update those to match this.

cmd/corectl/main.go
```diff
@@ -298,6 +300,72 @@ func reset(client *rpc.Client, args []string) {
...
+	flagOU := flags.String("ou", "", action+" X.509 `Organizational Unit` access")
+
+	flags.Usage = func() {
+		fmt.Println(usage)
+		flags.PrintDefaults()
+		os.Exit(1)
+	}
+	flags.Parse(args)
+	args = flags.Args()
+	if len(args) < 2 {
```
jeffomatic: I may not be up to speed on how `flags` works, but it looks like we're only using the first item in `args`. This check implies that there is at least one more argument we care about?

cmd/corectl/main.go
```diff
@@ -298,6 +300,72 @@ func reset(client *rpc.Client, args []string) {
...
+	editAuthz(client, args, "grant")
+}
+
+func revoke(client *rpc.Client, args []string) {
+	editAuthz(client, args, "revoke")
+}
+
+type grantReq struct {
+	Policy     string
+	Grant_Type string
```
tessr: Same comment about json tags as in #1042

docs/core/reference/corectl.md
```diff
@@ -141,3 +143,64 @@ other configuration will be deleted.
...
+
+If multiple flags are given,
+they will each take effect individually.
+That is, three separate grants will occur,
+not one grant with three restrictions on it.
+
+Arguments:
+
+ * **policy**: the policy to grant access to.
+See [XXX](#) for a list of policies and their meaning.
```
kr: @dominic is there a good place to link out to here?

dominic: The authorization guide has these listed. `../learn-more/authentication` I think.

cmd/corectl/main.go
```diff
@@ -298,6 +300,72 @@ func reset(client *rpc.Client, args []string) {
...
+		req.GrantType = "x509"
+		req.GrantData = map[string]interface{}{"subject": map[string]string{"CN": *flagCN}}
+		ok = ok && editAuthzEntry(client, action, req)
+	}
+	if *flagOU != "" {
+		req.GrantType = "x509"
+		req.GrantData = map[string]interface{}{"subject": map[string]string{"OU": *flagOU}}
+		ok = ok && editAuthzEntry(client, action, req)
+	}
+	if !ok {
```
tessr: So, it's possible that some of these grants were created, right? Should we change this to reflect the possibility that some grants were successfully created, or do we just assume that people will retry (idempotently)?

tessr: I guess this is related to Jeff's comment (1).

kr: It prints an error message for each failure, is that sufficient?

kr:
> A few general thoughts:

Thanks! Yeah I thought about these things… wrote my thoughts inline.

> 1. Per the current implementation, providing multiple flags will generate multiple grants for categorically distinct guards. I think it might make for a simpler experience to make each call to `corectl grant` to generate exactly one guard at a time.

If you feel strongly about this I'm willing to change it, but I have to admit I don't really understand how that would be any simpler (or more complicated — it just seems different).

It's more convenient to be able to do a few at a time, though I have no idea how common that would be vs, say, granting access to three or four different CNs, which would have to be separate command invocations anyway. So I'd expect this to be only a small marginal increase in convenience if we did stick with it.

> 2. What would you think about adding a `-x509json` flag …

I would be fine with adding that sort of flag but I'd rather wait until someone asks for it. As you say, they can use the SDK to do it meanwhile.

> 3. I think it would be good for the output of the grant/revoke subcommands to print out what they just did …

Be careful not to make too many assumptions about how the command will be run. Not every command invocation is done by a human sitting at a terminal waiting to read the output interactively. As a general rule, commands that produce data should print their results; commands that produce side effects should print nothing. (Of course, error messages go to stderr in either case.) Compare `cat` to `mv`.

If there's uncertainty about the effects of the command, a better fix is to make it easy to inspect the current state of the system with other means. For example, `mv` doesn't print "moved file X to Y" even when the destination is a directory; but `ls` lets you see what it just did if you're unsure.

We already have the dashboard to show existing grants. We could also add a corectl command for it (in a future PR) if necessary.

> 4. Bikeshed warning: despite the verbosity, I prefer `create-grant` and `delete-grant` as subcommands. These terms are consistent with the API calls, consistent with a pre-existing command (`create-token`), and they nicely namespace the commands so it is unambiguous what they're doing.

I strongly prefer the existing names in this PR.

In general (in API design), verbs are better than nouns (when you're talking about side effects), and this is a perfect case where that applies. There are specific verbs that are established terms of art that say exactly what we want. We should use them. That's better than a generic verb coupled with a noun.

Consistency with pre-existing commands is not terribly compelling. The existing commands were not designed, they were thrown together slapdash (mostly by me, mea culpa).

Consistency with the API calls is nice, but I think these names are better enough to outweigh that. (IMO the RPC names should also be verbs but I think it's too late to argue about that.) The words "grant" and "revoke" are perfectly precise and accurate and unambiguous.

cmd/corectl/main.go
kr: Yup, typo. Good catch, thanks.

cmd/corectl/main.go
tessr: Oh, yes, yes it is.

jeffomatic:
> If you feel strongly about this I'm willing to change it, but I have to admit I don't really understand how that would be any simpler (or more complicated — it just seems different).

Right, so two issues here:

The first is the footgun potential of misinterpreting lines such as the following:

```
corectl grant -cn Jeff -ou Chain client-readwrite
```

One understandable misreading of the above is that the command will grant the client-readwrite policy to an X.509 certificate whose subject has both a CN of "Jeff" and an OU of "Chain". The actual behavior would generate two separate grants.

That leads me to the second issue, which is that I'm having a hard time thinking of situations where I would want to use multiple flags in the same call, at least in the way that the implementation currently allows. Even if the user didn't misunderstand the behavior of the flags, it's difficult to imagine the circumstances under which a user would want to grant to a token AND a certificate, or to one cert with a CN AND another with an OU, at least in the same exact execution of the tool.

One thing that makes more sense is allowing the user to grant to multiple guards of the same type, such as:

```
corectl grant -t token1 -t token2 ... client-readwrite
```

But that's not the way the tool works.

Anyway, these two reasons are interrelated: because the legitimate use cases of the multi-flag behavior are kind of weird and rare, it's easy to misunderstand the use of multiple flags as parameterizing the same guard for a single grant. Only allowing a single flag gets rid of any of this ambiguity.

> I would be fine with adding that sort of flag but I'd rather wait until someone asks for it.

SGTM 👍 

> Be careful not to make too many assumptions about how the command will be run. Not every command invocation is done by a human sitting at a terminal waiting to read the output interactively. 

I think most of the use cases for this tool would be interactive. Someone will be at a terminal following along with our online documentation, or we'll be telling somebody what to do over Slack or the phone.

My primary concern here is that the effect of running this command is less legible from the original command than is ideal. It is easy to imagine intending to type `-t` and typing `-cn` instead, because it is late, or you zone out for a second and are staring at the wrong line in the usage instructions. We can give the user an immediate opportunity to validate their original intent by printing out what just happened. In some ways my concern here is related to the multiple-flag thing above--is there a way we can give the user some feedback that they actually did the right thing?

I get the comparisons to `mv`, but the concern about producing data vs side-effects feels somewhat academic. This tool is far less of an established norm than Unix builtins and its outcomes are therefore not as obvious, and besides, the extra output doesn't hurt. Also, for what it's worth, there is some reasonable precedent for printing out the side-effects for tools--think about something like `git fetch`, which gives you some diagnostic information and also shows you what got dragged down from the remote. 

> I strongly prefer the existing names in this PR.

Okay. 👍

tarcieri:
> What would you think about adding a -x509json flag that allows the user to specify the full JSON guard data?

@jeffomatic [cfssl](https://github.com/cloudflare/cfssl) has heavily leveraged JSON as part of their CLI tool as a way to make it easier to build wrappers around it. Personally I think it's a great idea to do something like this, although I'd make it easy to also accept input from STDIN.

kr:
> I think most of the use cases for this tool would be interactive.

This kind of assumption is why so many CLI tools are frustrating to work with. Following the conventions of the ecosystem — behaving like a good Unix citizen — is (ought to be) table stakes.

Even for the CLI (not just the API) we should seek orthogonal, simple, composable primitives. Having found some, we should try to make them as convenient to use as possible without sacrificing too much orthogonality.

I'm trying to consider human factors in the design here. This sort of output (especially if it's not machine-readable) gets in the way more often than it helps, in my experience. It makes the system overall bigger, harder to understand.

I'm also concerned about educating new users. I suggest putting "now check the authorization tables" as an explicit step in the instructions right after "grant access". Not only will that give them feedback about what they did, it'll also give them a more effective mental framework for later when they need to inspect the state of the system and they don't happen to be in the "grant access" workflow.

> it is late, or you zone out for a second and are staring at the wrong line

This sort of thing applies to reading the output just as well as the command you typed. It's not immediately obvious to me that echoing back what you typed will help. It also seems like it would aggravate scroll blindness and make the effects harder to perceive, not easier.

kr:
PTAL

@jeffomatic I updated this to require exactly one guard -- is that plus https://github.com/chain/chain/pull/1043#issuecomment-297190791 sufficient to alleviate your concern about understanding what the command is doing?

New syntax is:

```
:; corectl grant -h
usage: corectl grant [policy] [guard]

Where guard is one of:
  token=[name]   to affect an access token
  CN=[name]      to affect an X.509 Common Name
  OU=[name]      to affect an X.509 Organizational Unit

The type of guard (before the = sign) is case-insensitive.

```

cmd/corectl/main.go
```diff
@@ -298,6 +300,66 @@ func reset(client *rpc.Client, args []string) {
...
+
+func editAuthz(client *rpc.Client, args []string, action string) {
+	usage := "usage: corectl " + action + " [policy] [guard]"
+	var flags flag.FlagSet
+
+	flags.Usage = func() {
+		fmt.Fprintln(os.Stderr, usage)
+		fmt.Fprintln(os.Stderr, `
+Where guard is one of:
+  token=[name]   to affect an access token
```
jeffomatic: Nit: `id` is better for tokens, since this string is referred to as "id" everywhere else.

PR #1045, opened 2017-04-25
protocol/bc/bctest: new package

jbowens:
Move the prottest.NewIssuanceTx function into a separate
protocol/bc/bctest package. Also, add a prottest.Initial function
to get the initial block of a chain so that prottest.NewIssuanceTx is still
easy to use.

I'd like to be able to use NewIssuanceTx within the protocol and
protocol/validation packages but can't as long as it's in
protocol/prottest.


jbowens:
PTAL

bobg:
:eyes:

bobg:
LGTM, nice

PR #1049, opened 2017-04-25
core: fix policy for rpc/submit

tessr:


tessr:
Now, the open question is what the policy should actually be for this endpoint. `crosscore-readonly` seems incorrect (although okay for now). Should it be `crosscore-signblock`? Should we rename `crosscore-signblock` to `crosscore-readwrite`?

tessr:
[PTAL]

kr:
> Now, the open question is what the policy should actually be for this endpoint. `crosscore-readonly` seems incorrect (although okay for now).

Agreed, this seems okay for now, since accessing this RPC shouldn't be any more harmful than accessing say /get-blocks.

> Should it be `crosscore-signblock`? Should we rename `crosscore-signblock` to `crosscore-readwrite`? 

For any given Chain Core, does the set of clients that should have access to submit coincide with the set that should have access to signblock? Definitely not, so they can't share a policy.

I think we have two reasonable long term options:

1. make a new crosscore policy just for submit
2. rename `crosscore-readonly` to better reflect its meaning (`crosscore-harmless`? `crosscore-general`? `crosscore-whatever`?)

I lean toward option 2 as long as we can find a reasonable name, since the set of clients that should get /submit is basically always the same as those who should have /get-blocks.

tessr:
`crosscore-safe`?

jeffomatic:
`crosscore-general` gets my vote. Also, just remembered that we should take care to update docs where-ever "network access tokens" are mentioned.

tessr:
What about just `crosscore`?

tessr:
Well, I will remove anything specific from that TODO, and then maybe I can merge this PR, anyways.

kr:
LGTM

PR #1050, opened 2017-04-25
protocol/validation: use subtests

jbowens:
Use t.Run in validation tests. Most of the validation tests have
convenient descriptions in the test case struct already.

### DIFF
--- a/protocol/validation/validation_test.go
+++ b/protocol/validation/validation_test.go
@@ -1,6 +1,7 @@
 package validation
 
 import (
+	"fmt"
 	"math"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


protocol/validation/validation_test.go
```diff
@@ -336,26 +337,26 @@ func TestTxValidation(t *testing.T) {
...
-		err := checkValid(vs, tx.TxHeader)
-		if rootErr(err) != c.err {
-			t.Errorf("case %d (%s): got error %s, want %s; validationState is:\n%s", i, c.desc, err, c.err, spew.Sdump(vs))
-		}
+			if c.f != nil {
+				c.f()
+			}
+			err := checkValid(vs, tx.TxHeader)
+			if rootErr(err) != c.err {
+				t.Errorf("case %d (%s): got error %s, want %s; validationState is:\n%s", i, c.desc, err, c.err, spew.Sdump(vs))
```
kr: Ooo, now that we're using subtests, I think it would be better still to eliminate `case %d (%s)` from the output here and below.

kr: Another nice thing we can do (if we don't want to lose the numeric index) is to put `t.Logf("case %d", i)` at the top, inside the subtest. It'll only show up for the subtests that failed.

kr:
LGTM

PR #1051, opened 2017-04-25
core: consolidate dashboard authz policy mapping

kr:
The policy map uses the same rules as the HTTP ServeMux
to match request paths, which means we can express what
we want for dashboard more succinctly: all dashboard
files are public.


kr:
Also, given the matching rules, `/` is overly-permissive: it allows everything not explicitly mentioned in this table. We should default to deny for anything that's accidentally left out. For the specific case of allowing a request for exactly `/`, that is already done in a handler entirely outsize authz.

kr:
PTAL

jbowens:
> Also, given the matching rules, / is overly-permissive: it allows everything not explicitly mentioned in this table. We should default to deny for anything that's accidentally left out. For the specific case of allowing a request for exactly /, that is already done in a handler entirely outsize authz.

When I tested curling `/` locally with loopback_auth, I got:
```
{"code":"CH009","message":"Request could not be authenticated","temporary":false}
```
Is there something else we need to do to allow access to `/`?

jeffomatic:
> Is there something else we need to do to allow access to /? /info and /dashboard still work as expected.

I've also noticed similar behavior accessing `/` outside of `loopback_auth`. It should redirect to `/dashboard` without authenticating.

kr:
Ah, we moved the redirect outside of authz, but then it inadvertently got moved back inside again when authz moved out of chain/core in #1029. I'll undo the `/` part of this PR and we can fix the redirect separately.

kr:
PTAL

This is what I get for trying to sneak something "small" in to an unrelated PR.

jbowens:
lgtm

kr:
😖

kr:
rebased

PR #1054, opened 2017-04-26
core: convert info's configured_at property to RFC3339

jeffomatic:
The SDKs expect /info to return the `configured_at` property as an
RFC3339, but recent changes had converted it to millisecond Unixtime.
This PR restores the original format.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -92,10 +92,15 @@ func (a *API) leaderInfo(ctx context.Context) (map[string]interface{}, error) {
 		}
 	}
 
+	var (
+		configuredAtSecs  int64 = int64(a.config.ConfiguredAt / 1000)
+		configuredAtNSecs int64 = int64((a.config.ConfiguredAt % 1000) * 1e6)
+	)
+
 	m := map[string]interface{}{
 		"state":                             a.leader.State().String(),
 		"is_configured":                     true,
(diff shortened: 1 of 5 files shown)


jeffomatic:
PTAL

core/core.go
```diff
@@ -92,10 +92,15 @@ func (a *API) leaderInfo(ctx context.Context) (map[string]interface{}, error) {
...
+	var (
+		configuredAtSecs  int64 = int64(a.config.ConfiguredAt / 1000)
+		configuredAtNSecs int64 = int64((a.config.ConfiguredAt % 1000) * 1e6)
+	)
+
 	m := map[string]interface{}{
 		"state":                             a.leader.State().String(),
 		"is_configured":                     true,
-		"configured_at":                     a.config.ConfiguredAt,
+		"configured_at":                     time.Unix(configuredAtSecs, configuredAtNSecs).UTC(),
```
jbowens: It's valid to pass nsecs > 1 second, so you could do
```
time.Unix(0, a.config.ConfiguratedAt * int64(time.Millisecond))
```

jeffomatic: Ah good call. The original `configuredAt` is a `uint64`, so for theoretical overflow sanity's sake, I'd rather decrease the size of the numbers than increase them.

PR #1055, opened 2017-04-26
sdk/java: add CoreConfig class

jeffomatic:
The CoreConfig class finally gives the Java SDK access to
the config package's RPC calls, including:

- /configure
- /reset
- /info

### DIFF
--- a/sdk/java/src/main/java/com/chain/api/CoreConfig.java
+++ b/sdk/java/src/main/java/com/chain/api/CoreConfig.java
@@ -8,134 +8,134 @@ import java.util.*;
 
 public class CoreConfig {
 
-    public static class Info {
-        public String state;
+  public static class Info {
+    public String state;
 
-        @SerializedName("is_configured")
-        public Boolean isConfigured;
+    @SerializedName("is_configured")
+    public Boolean isConfigured;
 
-        @SerializedName("configured_at")
-        public Date configuredAt;
+    @SerializedName("configured_at")
+    public Date configuredAt;
 
-        @SerializedName("is_signer")
-        public Boolean isSigner;
+    @SerializedName("is_signer")
+    public Boolean isSigner;
 
-        @SerializedName("is_generator")
-        public Boolean isGenerator;
+    @SerializedName("is_generator")
+    public Boolean isGenerator;
 
-        @SerializedName("generator_url")
-        public String generatorUrl;
+    @SerializedName("generator_url")
+    public String generatorUrl;
 
-        @SerializedName("generator_access_token")
-        public String generatorAccessToken;
+    @SerializedName("generator_access_token")
+    public String generatorAccessToken;
 
-        @SerializedName("blockchain_id")
-        public String blockchainId;
+    @SerializedName("blockchain_id")
+    public String blockchainId;
 
-        @SerializedName("block_height")
-        public long blockHeight;
+    @SerializedName("block_height")
+    public long blockHeight;
 
-        @SerializedName("generator_block_height")
-        public long generatorBlockHeight;
+    @SerializedName("generator_block_height")
+    public long generatorBlockHeight;
 
-        @SerializedName("generator_block_height_fetched_at")
-        public Date generatorBlockHeightFetchedAt;
+    @SerializedName("generator_block_height_fetched_at")
+    public Date generatorBlockHeightFetchedAt;
 
-        @SerializedName("network_rpc_version")
-        public int networkRpcVersion;
+    @SerializedName("network_rpc_version")
+    public int networkRpcVersion;
 
-        @SerializedName("core_id")
-        public String coreId;
+    @SerializedName("core_id")
+    public String coreId;
 
-        public String version;
+    public String version;
 
-        @SerializedName("build_commit")
-        public String buildCommit;
+    @SerializedName("build_commit")
+    public String buildCommit;
 
-        @SerializedName("build_date")
-        public String buildDate;
+    @SerializedName("build_date")
+    public String buildDate;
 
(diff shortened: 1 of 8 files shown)


jeffomatic:
PTAL @boymanjor @erykwalder

sdk/java/src/main/java/com/chain/api/CoreConfig.java
```diff
@@ -0,0 +1,141 @@
...
+
+  public static void resetEverything(Client client) throws ChainException {
+    Map<String, Boolean> params = new HashMap<>();
+    params.put("everything", true);
+    client.request("reset", params, SuccessMessage.class);
+  }
+
+  public static class Builder {
+    @SerializedName("is_generator")
+    private Boolean isGenerator;
```
boymanjor: This value will default to null. I'm assuming we actually want this to default to `false`, in which case we should use `boolean`. In general, its better to use `boolean` (unless handling null values is required).

jeffomatic: Hmm, very interesting! TBH I typed `Boolean` with the capital B more or less by accident, but now that you have me thinking, I suspect that capital-B `Boolean` may in fact be a bit better in this case.

> I'm assuming we actually want this to default to false

I suspect this is not accurate. There's no particular reason `false` is better than `true` when there is no value  returned from the server. Defaults are okay as a convenience mechanism for input, but they seem irrelevant when we're talking about output.

What `null` would imply here is "undefined", "no information", or even "there's a bug in the response-generating code and you should not rely on this value". I would actually love to be able to mark some (but not all) of these fields as "required", so that the JSON deserializer would throw an exception if the serialized version doesn't contain certain fields. Unfortunately, I dont think [GSON will be adding that any time soon](https://github.com/google/gson/issues/61).

FWIW, what will happen here if you try to use a `null` Boolean object is a null pointer exception, which I think is ultimately better than arbitrarily assuming falsiness. Admittedly it's the lesser of two evils.

boymanjor: I see your point, but this is on the `Builder`. Really we're talking about what should happen if you don't call `setIsGenerator` before calling `configure`. Leaving it as `Boolean` makes `setIsGenerator` a required call. That's fine, just wanted to be sure that is what we wanted.

jeffomatic: @boymanjor Ah, thanks for the clarification, I thought you were referring to the `Boolean` everywhere, which again I kinda just used by accident. I fully agree with using lower-case `boolean` here. Thanks!

jbowens: +1 to providing reasonable defaults in the `Builder`.

Also, more generally, it looks like we use `int` and `long`, not `Integer` and `Long`. I'd vote for consistency and use `boolean` including across response objects.

sdk/java/src/main/java/com/chain/api/CoreConfig.java
```diff
@@ -0,0 +1,141 @@
...
+    public String state;
+
+    @SerializedName("is_configured")
+    public Boolean isConfigured;
+
+    @SerializedName("configured_at")
+    public Date configuredAt;
+
+    @SerializedName("is_signer")
+    public Boolean isSigner;
```
jbowens: Generally, it looks like we use int and long in the SDK, not Integer and Long. I'd vote for consistency and use primitives across all objects.

jbowens: Plus, it's a little easier on the consumer. The object types are easy to use incorrectly. Autoboxing can cause some really unexpected behavior, like a NullPointerException here:

```
if (config.isSigner == true) {
  // ...
}
```

sdk/java/src/main/java/com/chain/api/CoreConfig.java
jeffomatic: Yeah I think the consistency thing is compelling. I wasn't sure how to weigh that until you pointed out that we also have ints/longs in this exact object. So, why pretend to solve the invalid-response problem just for bools? Until we have a proper answer for doing client-side validation, I should just quit worrying about it. Lower-case 4 lyfe.

sdk/java/src/main/java/com/chain/api/CoreConfig.java
```diff
@@ -0,0 +1,141 @@
...
+    @SerializedName("build_commit")
+    public String buildCommit;
+
+    @SerializedName("build_date")
+    public String buildDate;
+
+    @SerializedName("build_config")
+    public BuildConfig buildConfig;
+
+    public Map<String, Object> health;
```
jbowens: is there any more schema to this that we can provide? it's a little unclear how to use this as-is. If I understand the cored code right, this map would look something like:
```
{
  "errors": {
    "generator": "error generating new block"
  }
}
```

Maybe instead of a Map<String,Object> this should be a separate class? Even if we only expose a method that returns this map, it'll give us the flexibility to introduce more convenience methods down the road.

jbowens: I think this also might be the most useful of the fields exposed here, so I don't feel like it's over-engineering to ensure we can make it easy to access.

jeffomatic: Agreed. Let's land this PR and follow up with something. How confident are you in the stability of the structure of the health object?

jbowens: Might want to ask @kr, I'm not really sure.

PR #1056, opened 2017-04-26
core: remove authz workaround

kr:
This workaround, returning HTTP 401 instead of 403,
turns out not to be ideal, since dashboard has behavior
for 401 we don't want here. We decided to update
dashboard in a followup, but for now we can remove
this so at least we're returning the correct status.


kr:
@jeffomatic I think this is what we discussed in slack. Lemme know if I misunderstood.

cc @tessr

PR #1057, opened 2017-04-26
core: remove req path / from policy public

kr:
The matching rules for policies are the same as for
ServeMux: it treats a pattern ending in / as a prefix.
Adding / to the table not only matches the exact request
path /, but any other path not reflected in the table.
Putting policy public on that pattern therefore allows
any unauthenticated request to any path we may have
inadvertently left out of the table. It is safer to
"fail closed" -- to reject missing paths.

Redirecting / to /dashboard/ entirely outside of authz
lets us avoid adding this pattern to the policy table.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -193,6 +193,7 @@ func main() {
 	mux.Handle("/", &coreHandler)
 
 	var handler http.Handler = mux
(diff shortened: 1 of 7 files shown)


kr:
PTAL

PR #1058, opened 2017-04-26
cmd/cored: avoid spurious 404 responses

### LINKED ISSUE #1053: bug: API returns 404 briefly after reset or configure
For a small period after a `/reset`, the API returns a non-JSON 404 error message for all routes. I suspect this may also be the case for any core that is starting up, but I only noticed this issue when resetting from the dashboard.

The core ideally should not return any HTTP response until the API is fully ready, or should return some kind of temporary availability error (503), formatted as JSON.

kr:
The chain/core API object depends on reading a config
object during startup, and reading the config object
depends on reading from raft storage, which in turn
depends on handling HTTP requests. We previously handled
this by adding the API handler to the mux after setup.
This caused us to respond 404 to any non-raft request
received during the brief startup time.

Now, we install a handler in the mux immediately, but
synchronized with the API handler, so it blocks until
the API handler is ready. This should provide the
correct behavior for all HTTP responses at all times.

In the future, we might investigate lazily loading the
config object to avoid this awkward data dependency
sequence altogether.

Fixes #1053.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -18,6 +18,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/kr/secureheader"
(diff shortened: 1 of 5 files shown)


kr:
PTAL

This turned out to be not so hard to fix after all. I'm not sure it's the ideal long term solution (see comment about lazy-loading the config), but it should get the job done.

PR #1059, opened 2017-04-26
cmd/cored: enforce TLS without build tag plain_http

kr:
With this change, TLS will be required by default, and can be made optional with build tag plain_http. Our intention is to require TLS internally during development (hence why the default build configuration is to require it) and for production, but not to require TLS for desktop builds (Mac OS X and Windows).

This change also unconditionally sets the secureheader field HTTPSRedirect to false. Chain Core never listens on both TLS and non-TLS ports simultaneously, so it is pointless to ever set this field to true. Doing so was a longstanding mistake, which we might as well fix now.

Future work: consider renaming plain_http to plain_http_ok to better reflect that TLS is optional (not prohibited) in that configuration.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -53,7 +53,7 @@ There are four build tags that change the behavior of the resulting binary:
   - `reset`: allows the core database to be reset through the api
   - `loopback_auth`: allows unauthenticated requests on the loopback device
   - `no_mockhsm`: disables the MockHSM provided for development
-  - `plain_http`: allows plain HTTP requests (not yet implemented)
+  - `plain_http`: allows plain HTTP requests
 
 ```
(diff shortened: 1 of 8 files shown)


kr:
See https://github.com/chain/chain/compare/tlsflag...rename-plain-http for the renaming follow up.

kr:
I am not sure we want to start using this routinely inside Chain yet, but it is pretty important to turn it on before shipping 1.2 in a server build. So I think we should land this, and use the build tag for a while if necessary.

But we should also try to follow up soon by setting up TLS for ourselves. We can start with self-signed certs.

kr:
So, PTAL.

PR #1061, opened 2017-04-26
core: rename crosscore-readonly to crosscore

tessr:
The `rpc/submit` endpoint isn't actually read-only, so this is a more accurate name.


tessr:
PTAL. Questions, comments, and shed color suggestions can be left below. 👇

kr:
LGTM

PR #1066, opened 2017-04-26
sdk/node: sanitize X509 guard data

jeffomatic:
This commit enforces consistency of behavior and language for the Node
SDK's auth grant interface. Most significantly, the Node SDK will now
arrayify X509 Subject attributes where necessary.

### DIFF
--- a/sdk/node/src/util.js
+++ b/sdk/node/src/util.js
@@ -1,0 +1,42 @@
+const x509SubjectAttributes = {
+  C: {array: true},
+  O: {array: true},
+  OU: {array: true},
+  L: {array: true},
+  ST: {array: true},
+  STREET: {array: true},
+  POSTALCODE: {array: true},
+  SERIALNUMBER: {array: false},
+  CN: {array: false},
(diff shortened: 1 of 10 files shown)


jeffomatic:
PTAL

sdk/node/src/util.js
```diff
@@ -0,0 +1,42 @@
...
+    const attrib = x509SubjectAttributes[k.toUpperCase()]
+    if (!attrib) {
+      throw new Error(`X509 guard data contains invalid subject attribute: ${k}`)
+    }
+
+    let v = oldSubject[k]
+    if (!attrib.array && Array.isArray(v)) {
+      throw new Error(`X509 guard data contains invalid array for attribute ${k}: ${v.toString()}`)
+    } else if (attrib.array && !Array.isArray(v)) {
+      newSubject[k] = [v]
```
kr: This makes me miss `flatten` from ruby. I think this is the first time I've ever missed something from ruby.

PR #1068, opened 2017-04-26
core: add misc validation of create-grant input

### LINKED ISSUE #1065: /create-authorization-grant validations
- [ ] Return an error if the requested policy is not in the list of possible polices
- [ ] Return an error if the guard type is invalid
- [ ] Return an error if the guard data for access_token guards contains anything other than `id`

jeffomatic:
Fixes #1065


jeffomatic:
PTAL

PR #1069, opened 2017-04-26
core: add authz to internal policy

kr:
This is so corectl can use these RPCs.


tessr:
LGTM. We should probably document somewhere what exactly each policy is for. (Although it makes sense, it's non-obvious that corectl would use internal.)

jeffomatic:
Ditto what @tessr said...code looks fine but I admit I have a hazy understanding of the internal policy and its relationship to the production bootstrapping process.

kr:
The short version is `internal` is for anything that cored or corectl needs to do.

PR #1072, opened 2017-04-27
database/raft: plumb TLS config into raft

kr:
Both package raft and package core used to rely on a TLS
client config in DefaultTransport. We inadvertently
broke raft when core stopped using DefaultTransport and
we stopped configuring it.

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -152,9 +152,25 @@ func main() {
 		internalDN = &x509Cert.Subject
 	}
 
+	// TODO(kr): make core.UseTLS take just an http client
+	// and use this object in it.
+	httpClient := new(http.Client)
+	httpClient.Transport = &http.Transport{
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSClientConfig:       tlsConfig,
+		TLSHandshakeTimeout:   10 * time.Second,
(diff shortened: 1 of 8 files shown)


cmd/cored/main.go
```diff
@@ -152,9 +152,26 @@ func main() {
...
+			DualStack: true,
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+
 	raftDir := filepath.Join(home, "raft") // TODO(kr): better name for this
 	// TODO(tessr): remove tls param once we have tls everywhere
```
tessr: this TODO needs updated/removed

cmd/cored/main.go
```diff
@@ -152,9 +152,26 @@ func main() {
...
+	httpClient := new(http.Client)
+	httpClient.Transport = &http.Transport{
+		TLSClientConfig: tlsConfig,
+
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		MaxIdleConns:          100,
```
tessr: just wondering, are there reasons we've chosen these numbers? I feel like we set timeouts and things other places too (although now I can't find anything specific), I wonder if we should be more deliberate about these things

kr: Yeah we prob should be more deliberate about them at some point. I copied these from https://godoc.org/net/http#DefaultTransport which is what we were using before, to avoid any unintended changes in behavior.

tessr: can you add a comment explaining that?

PR #1073, opened 2017-04-27
core: use Chain.State to retrieve latest bc state

jbowens:
Previously, core/generator and core/fetch had their own local pointers
to the latest block and latest snapshot. We already need to store and
expose blockchain state on the Chain struct for the /sign-block RPC.
To address #519 without too invasive of changes, I think we'll need to
pull the state from the shared Chain struct.

### DIFF
--- a/core/generator/block.go
+++ b/core/generator/block.go
@@ -44,6 +44,7 @@ func (g *Generator) makeBlock(ctx context.Context) (err error) {
 	t0 := time.Now()
 	defer recordSince(t0)
 
+	latestBlock, latestSnapshot := g.chain.State()
 	var b *legacy.Block
 	var s *state.Snapshot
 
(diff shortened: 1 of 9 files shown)


kr:
:eyes:

core/generator/block.go
```diff
@@ -79,13 +80,11 @@ func (g *Generator) makeBlock(ctx context.Context) (err error) {
...
 	}
-
-	// g.commitBlock will update g.latestBlock and g.latestSnapshot.
-	return g.commitBlock(ctx, b, s)
+	return g.commitBlock(ctx, b, s, latestBlock)
 }
 
-func (g *Generator) commitBlock(ctx context.Context, b *legacy.Block, s *state.Snapshot) error {
-	err := g.getAndAddBlockSignatures(ctx, b, g.latestBlock)
+func (g *Generator) commitBlock(ctx context.Context, b *legacy.Block, s *state.Snapshot, latestBlock *legacy.Block) error {
```
kr: Nit, should this be called `prevBlock` in this function?

PR #1075, opened 2017-04-27
database/raft: avoid server cert errors for peers

kr:


database/raft/raft.go
```diff
@@ -147,6 +147,11 @@ type Getter interface {
 func Start(laddr, dir, bootURL string, httpClient *http.Client) (*Service, error) {
```
tessr: I know this was a previous PR, but while we're poking at this, can you add `httpClient` to the doc comment for `Start`?

database/raft/tls.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"chain/errors"
+)
+
+func verifyTLSName(name string, client *http.Client) error {
+	c := clientTLS(client)
+	if c == nil {
+		return nil
+	}
+	x509Cert, err := x509.ParseCertificate(c.Certificates[0].Certificate[0])
```
tessr: Is it possible that this line will cause an index out of bounds? (And if so, is crashing the desired behavior here?)

jbowens:
Is this a raft-specific check or does it apply more generally to cored too?

database/raft/tls.go
kr: If there's a TLS client config at all here, it should definitely have a cert, so I think crashing is fine.

kr:
Yeah it is more general. It applies anywhere we have an address or hostname we are planning to tell other processes to use to reach us. So raft and the leader key. We could add another check for the leader key or we could potentially have the leader stuff piggyback on raft's concept of identity. (Like, use the raft cluster member id as the leader key and also have a way to look up an address from a member id. idk)

kr:
Or maybe we should just move this check out into package main. Not sure.

PR #1076, opened 2017-04-27
generated/dashboard: bundle dashboard

jeffomatic:
This is a checkpoint commit for recent dashboard work. It should help us detect any obvious bugs through internal use.


jeffomatic:
PTAL

PR #1080, opened 2017-04-27
docs: update docs to account for authz updates

jeffomatic:
This commit is meant to be an MVP update to documentation that
addresses authentication and authorization. In particular, it updates
language around "network tokens" to account for our new worfklow around
crosscore and crosscore-signblock policies.


jeffomatic:
PTAL

PR #1082, opened 2017-04-27
database/raft: create the allowed members list

tessr:
When a new cored joins a raft cluster, there must be an associated grant stored in that cluster in order for this new cored to talk to the other members of the cluster. 

We'd like the existence of this grant to be a prerequisite for a new member joining the cluster. Similarly, we'd like it to be impossible to delete a grant that is associated with a cluster member, unless that member is also removed from the cluster. In order to keep track of this, we're creating a list of "allowed members"--that is, addresses associated with nodes that are allowed to become members. 

This commit creates that list, but doesn't hook it up to the rest of the code managing cluster membership.

### DIFF
--- a/database/raft/membership.go
+++ b/database/raft/membership.go
@@ -2,19 +2,18 @@ package raft
 
 import "context"
 
-const potentialMemberPrefix = "potential"
+const allowedMemberPrefix = "/raft/allowed"
 
-// AddPotentialMember adds an address for a potential new member to the
-// list of potential cluster members.
-// An address must be listed as a potential cluster member before the node
+// AddAllowedMember adds an address for a member to the list of allowed cluster members.
+// An address must be listed as a allowed cluster member before the node
 // listening on that address can join the cluster.
-func (sv *Service) AddPotentialMember(ctx context.Context, addr string) error {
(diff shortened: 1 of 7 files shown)


tessr:
PTAL

kr:
:eyes:

database/raft/membership.go
```diff
@@ -0,0 +1,22 @@
+package raft
+
+import "context"
+
+const potentialMemberPrefix = "potential"
```
kr: How do you feel about leading `/` in raft keys? I think currently we sometimes do it and sometimes not.

Separately, (nit) I am not sure "potential" is the right word, because as soon as they join they'll be actual members, not potential. But they should still be in this list.

Thirdly, do you think this set of things should be namespaced along with everything else this package does internally into like a `raft/` (or `/raft/`) prefix? So in this case might end up as `raft/potential/example.com:1234`.

tessr: 1. Ah, yes, I'll add in `/`. 
2. What about `allowed`? 
3. I think that's a good idea--I forgot about it, basically.

tessr:
PTAL

kr:
LGTM

PR #1083, opened 2017-04-27
core: s/networkRPC/crosscoreRPC

jeffomatic:
"Cross-core API" has replaced "Network API" in our lexicon. This commit
is a purely server-side, backward-compatible update. Updates to the SDK
and dashboard will follow.


jeffomatic:
PTAL

kr:
lgtm

PR #1079, opened 2017-04-27
cmd/dumpschema: add new command

jbowens:
Add new command to replace bin/dumpschema now that corectl doesn't
have a migrate subcommand.

PR #1084, opened 2017-04-28
core/generator: only upsert higher pending blocks

jbowens:
When upserting a pending block as the generator, only upsert if the
new block is at a higher height than the previous block. This wraps
up #519 for the generator.

This should land after #1079.

### DIFF
--- a/core/generator/block.go
+++ b/core/generator/block.go
@@ -23,6 +23,8 @@ import (
 // signatures required.
 var errTooFewSigners = errors.New("too few signers")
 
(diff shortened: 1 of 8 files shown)


jbowens:
Oof, on second thought, if the other process successfully commits a block at n + 1, the current process will still have a stale state tree. This won't completely handle two processes both generating blocks as if they were leader. It does prevent the two processes from sending different blocks at the same height to signers.

kr:
:eyes:

kr:
LGTM

PR #1085, opened 2017-04-28
net/http/authz: introduce protected grants

tessr:
Protected grants are grants that shouldn't be edited by users. For example, they are used by nodes in a raft cluster to authenticate one another--removing a grant in that situation would cause raft to fail.


core/grants.go
```diff
@@ -118,13 +124,15 @@ func (a *API) listGrants(ctx context.Context) (map[string]interface{}, error) {
 				return nil, errors.Wrap(err)
 			}
 
-			grant := apiGrant{
-				GuardType: g.GuardType,
-				GuardData: data,
-				Policy:    g.Policy,
-				CreatedAt: g.CreatedAt,
+			if !g.Protected { // is this right?
```
tessr: should we return protected grants here?

kr: I like the idea of showing them so the user can see what the true security boundary surface area looks like. But that might mean more work for dashboard to show them grayed-out or whatever, somehow indicate that they're protected.

@jeffomatic @dominic what do you think? Ultimately I am okay with this either way depending on what the product folks think.

core/grants.go
```diff
@@ -20,8 +20,14 @@ type apiGrant struct {
...
 }
 
-// ErrMissingTokenID is returned when a token does not exist.
-var errMissingTokenID = errors.New("id does not exist")
+var (
+	// errMissingTokenID is returned when a token does not exist.
+	errMissingTokenID = errors.New("id does not exist")
+
+	// errProtectedGrant is returned when a grant is protected and therefore cannot
+	// be directly modified or deleted by the user.
```
kr: Nit: s/modified or//

because grants are immutable anyway.

core/grants.go
```diff
@@ -90,6 +97,7 @@ func (a *API) createGrant(ctx context.Context, x apiGrant) (*apiGrant, error) {
 		GuardData: data,
 		Policy:    g.Policy,
 		CreatedAt: g.CreatedAt,
+		Protected: g.Protected,
```
tessr: should we include this when creating a grant? it will (probably) always be false.

tessr:
Updated the API, PTAL

core/grants.go
kr: Nah, in fact I think it's important to force that field to be false in the RPC, so that it's impossible for the client to create a protected grant.

If they made one, they'd be unable to revoke it. The only way protected grants are a reasonable thing to have is if they're completely managed by cored (both creating and deleting).

tessr: Good point. I'll set it at the top of this function, too, then.

kr: Or just leave out this line?

tessr: This line only affects the return value, not what's stored.

kr: Oops, hah.

kr:
This is a bit of a corner case, but what if someone creates a normal grant before cored tries to create its protected grant for the same thing? What would happen? What should happen?

jeffomatic:
Part of me wants to treat those 2 cases as separate grants, but I have a feeling it may be simpler just to give the existing grant protected status. There is a similar edge case where somebody redundantly makes an API call to create a grant when it already exists in protected form, and I think I'd just want to treat that as a no-op.

tessr:
Alright. I broke a thing, I added a test and fixed a thing. PTAL

core/grants.go
```diff
@@ -205,6 +231,8 @@ func (a *API) deleteGrantsByAccessToken(ctx context.Context, token string) error
 
 			if id, _ := data["id"].(string); id != token {
 				keep = append(keep, g)
+			} else if g.Protected {
+				return errProtectedGrant
```
kr: Should we remove this check?

kr:
LGTM — I added one question but good to land IMO with whatever you decide about that

PR #1088, opened 2017-04-28
sdk/java: add concrete health object

jeffomatic:


jeffomatic:
PTAL @jbowens. With @kr's recent update (#1087), the `health.errors` is a much simpler object, and I think exposing the raw map is now sufficient.

PR #1089, opened 2017-04-28
Readme.md: remove heroku deploy button

jbowens:
Chain Core requires filesystem persistence.


jeffomatic:
I suspect we will want to remove `app.json` as well.

PR #1091, opened 2017-04-28
database/raft: remove debugging log messages

kr:
We want cored log output to be useful for operators. For
our own debugging purposes, we can add and remove
logging as necessary during development, or use other
tools such as memory debuggers and tracing.

### DIFF
--- a/database/raft/raft.go
+++ b/database/raft/raft.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"hash/crc32"
 	"io/ioutil"
+	stdlog "log"
 	"net"
 	"net/http"
 	"os"
@@ -201,6 +202,7 @@ func Start(laddr, dir, bootURL string, httpClient *http.Client) (*Service, error
 		Applied:         sv.state.AppliedIndex(),
 		MaxSizePerMsg:   4096,
 		MaxInflightMsgs: 256,
+		Logger:          &raft.DefaultLogger{Logger: stdlog.New(ioutil.Discard, "", 0)},
 	}
 
 	if walobj != nil {
@@ -248,11 +250,9 @@ func Start(laddr, dir, bootURL string, httpClient *http.Client) (*Service, error
 // think there's no leader. It should only be used when election failures are
 // okay, since it will log errors but won't return them.
 func triggerElection(ctx context.Context, sv *Service) {
-	log.Printf(ctx, "triggering election")
 	err := sv.raftNode.Campaign(context.Background())
 	if err != nil {
-		log.Printf(ctx, "election failed")
-		log.Error(ctx, err)
+		log.Error(ctx, err, "election failed")
 	}
 }
 
@@ -510,7 +510,6 @@ func (sv *Service) get(ctx context.Context, key string) ([]byte, error) {
 			err := sv.Set(ctx, "/dummyWrite", []byte(""))
 			if err != nil {
 				// TODO(tessr): should probably return this error
-				log.Printf(ctx, "dummy write failed with %s", err)
 				return
 			}
 		case <-cancelDummy:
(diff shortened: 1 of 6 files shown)


database/raft/raft.go
```diff
@@ -607,7 +602,7 @@ func (sv *Service) serveJoin(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	log.Printkv(req.Context(), "at", "join-done", "addr", x.Addr, "id", newID)
+	log.Printkv(req.Context(), "at", "serveJoin", "addr", x.Addr, "id", newID)
```
kr: Not sure if we should keep this one. (Or maybe reword it.)

database/raft/raft.go
```diff
@@ -650,17 +645,15 @@ func (sv *Service) join(addr, baseURL string) error {
...
+	if !raft.IsEmptySnap(raftSnap) {
 		err = sv.state.RestoreSnapshot(raftSnap.Data, raftSnap.Metadata.Index)
 		if err != nil {
 			return errors.Wrap(err)
 		}
 		sv.confState = raftSnap.Metadata.ConfState
 		sv.snapIndex = raftSnap.Metadata.Index
-		log.Printkv(ctx, "at", "joined", "appliedindex", raftSnap.Metadata.Index)
 	}
+	log.Printkv(ctx, "at", "joined", "appliedindex", raftSnap.Metadata.Index)
```
kr: Also not sure if we should keep this one.

database/raft/raft.go
```diff
@@ -201,6 +202,7 @@ func Start(laddr, dir, bootURL string, httpClient *http.Client) (*Service, error
 		Applied:         sv.state.AppliedIndex(),
 		MaxSizePerMsg:   4096,
 		MaxInflightMsgs: 256,
+		Logger:          &raft.DefaultLogger{Logger: stdlog.New(ioutil.Discard, "", 0)},
```
tessr: what does this line do?

database/raft/raft.go
```diff
@@ -510,7 +510,6 @@ func (sv *Service) get(ctx context.Context, key string) ([]byte, error) {
 			err := sv.Set(ctx, "/dummyWrite", []byte(""))
 			if err != nil {
 				// TODO(tessr): should probably return this error
```
tessr: could you update this todo to "do something about this error"? 

or... not? we do occasionally get this error and it's totally fine, the node just retries the dummy write.

database/raft/raft.go
tessr: I think one "joined the cluster" message is useful, probably don't need both of them.

kr: Okay, then any opinion on which one we should keep (new member vs existing member)? I will prob remove the serveJoin message since it would happen on an arbitrary existing member. This one on the new member is prob more salient.

database/raft/raft.go
kr: It discards the coreos raft package log output. Its [default logger](https://github.com/chain/chain/blob/50dff7d598be223161a5a56299719cff95cd99ab/vendor/github.com/coreos/etcd/raft/logger.go#L47-L49) goes to stderr.

I didn't see any difference in the output in my tests, but I suppose it's possible I just didn't hit anything that causes that package to print log messages.

database/raft/raft.go
kr: Aha okay, I'll remove the todo (and add a comment saying it's okay to ignore the error).

database/raft/raft.go
tessr: So if we're trying to debug, we can just pass in a different io writer? SGTM

database/raft/raft.go
tessr: SGTM

tessr:
[approved]

PR #1092, opened 2017-04-28
sdk: rename network RPC version to crosscore RPC version

jeffomatic:
SDKs maintain backward compatibility with network RPC version, but this
field is deprecated in favor of the equivalently-valued crosscore RPC
version.

Also includes dashboard updates. The dashboard assumes the
presence of a 1.2 Core.

### DIFF
--- a/dashboard/src/features/testnet/reducers.js
+++ b/dashboard/src/features/testnet/reducers.js
@@ -20,9 +20,9 @@ export const blockchainId = (state = 0, action) => {
   return state
 }
 
-export const rpcVersion = (state = 0, action) => {
+export const crosscoreRpcVersion = (state = 0, action) => {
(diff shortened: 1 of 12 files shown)


dashboard/src/features/testnet/reducers.js
```diff
@@ -20,9 +20,9 @@ export const blockchainId = (state = 0, action) => {
   return state
 }
 
-export const rpcVersion = (state = 0, action) => {
+export const crosscoreRpcVersion = (state = 0, action) => {
   if (action.type == 'TESTNET_CONFIG') {
-    return action.data.network_rpc_version
+    return action.data.crosscore_rpc_version || action.data.network_rpc_version
```
dominic: Did a double-take that this wasn't camelCased, but it's actually an out-of-band fetch request.

PR #1097, opened 2017-04-29
database/raft: add useTLS config option

tessr:
The standard library's `net/http` package modifies the http client's transport field when requests are made. Therefore it's not safe to check if we should use TLS by checking the transport. This PR adds a `useTLS` flag to the raft service; in the future, the raft service may be configured by passing in run options, a la cored.


tessr:
PTAL

kr:
LGTM

PR #1098, opened 2017-04-30
protocol/vmutil: add mechanism for adding jumps in Builder

bobg:
The `Builder` object keeps track of jump-target addresses and fills them in when `Build` is called.

### DIFF
--- a/core/txbuilder/constraint.go
+++ b/core/txbuilder/constraint.go
@@ -41,7 +41,8 @@ func (t timeConstraint) code() []byte {
 		}
 		builder.AddOp(vm.OP_MAXTIME).AddInt64(int64(t.maxTimeMS)).AddOp(vm.OP_LESSTHANOREQUAL)
 	}
-	return builder.Program
+	prog, _ := builder.Build()
+	return prog
 }
 
 // outpointConstraint requires the outputID (and therefore, the outpoint) being spent to equal the
@@ -53,7 +54,8 @@ func (o outputIDConstraint) code() []byte {
 	builder.AddData(bc.Hash(o).Bytes())
 	builder.AddOp(vm.OP_OUTPUTID)
 	builder.AddOp(vm.OP_EQUAL)
-	return builder.Program
+	prog, _ := builder.Build()
+	return prog
 }
 
 // refdataConstraint requires the refdatahash of the transaction (if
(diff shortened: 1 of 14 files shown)


bobg:
PTAL

protocol/vmutil/builder.go
```diff
@@ -1,31 +1,111 @@
...
+// number. The actual program location of the target does not need to
+// be known yet, as long as SetJumpTarget is called before Build.
+func (b *Builder) AddJumpIf(target int) *Builder {
+	return b.addJump(vm.OP_JUMPIF, target)
+}
+
+func (b *Builder) addJump(op vm.Op, target int) *Builder {
+	b.AddOp(op)
+	var addrBytes [4]byte
+	if addr, ok := b.jumpAddr[target]; ok {
```
erykwalder: We can reduce code paths by removing the conditional here and always doing the else block.

core/txbuilder/constraint.go
```diff
@@ -41,7 +41,8 @@ func (t timeConstraint) code() []byte {
 		}
 		builder.AddOp(vm.OP_MAXTIME).AddInt64(int64(t.maxTimeMS)).AddOp(vm.OP_LESSTHANOREQUAL)
 	}
-	return builder.Program
+	prog, _ := builder.Build()
```
erykwalder: This change seems to encourage ignoring errors for programs known to be "safe". I have a slight nag in my head that now that the signature has changed, the implementation can change, and outside expectations won't be met. I can't think of any reasonable solutions aside from an additional type, which seems worse.

All this to say, I don't think this concern is blocking, especially since the previous interface exposed was only 4 methods. Just wanted to share this feeling.

protocol/vmutil/builder_test.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"testing"
+
+	"chain/protocol/vm"
+)
+
+func TestAddInt64(t *testing.T) {
```
erykwalder: Fine to leave it in, but is there a deeper reason for including this test with this batch?

protocol/vmutil/builder.go
bobg: Well-spotted, thanks.

core/txbuilder/constraint.go
bobg: Does it help that I added a promise to the contract of `Builder.Build` saying that errors are impossible if you never call `AddJump` or `AddJumpIf`?

protocol/vmutil/builder_test.go
bobg: No, and I agree, it's cleaner without it; removed.

erykwalder:
LGTM

kr:
Do you mind holding off on this until 1.3?

bobg:
rebased

PR #1103, opened 2017-05-01
all: use "localhost auth" instead of "loopback auth"

jeffomatic:
"Localhost" is more legible than "loopback". This builds on the
direction already started by the "localhost" authorization guard.

The loopback_auth build tag will be updated in the future, when there is
less QA churn.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -49,15 +49,15 @@ var (
 	// This is the default, Chain development configuration.
 	// These options can be updated with build tags.
 	BuildConfig = struct {
-		LoopbackAuth bool `json:"is_loopback_auth"`
-		MockHSM      bool `json:"is_mockhsm"`
-		Reset        bool `json:"is_reset"`
-		PlainHTTP    bool `json:"is_plain_http"`
+		LocalhostAuth bool `json:"is_localhost_auth"`
+		MockHSM       bool `json:"is_mockhsm"`
+		Reset         bool `json:"is_reset"`
+		PlainHTTP     bool `json:"is_plain_http"`
 	}{
-		LoopbackAuth: false,
-		MockHSM:      true,
-		Reset:        false,
-		PlainHTTP:    false,
+		LocalhostAuth: false,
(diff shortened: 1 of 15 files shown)


kr:
:eyes:

tessr:
can you update the Makefile in `database/raft` too?

core/config/config.go
```diff
@@ -49,15 +49,15 @@ var (
 	// This is the default, Chain development configuration.
 	// These options can be updated with build tags.
 	BuildConfig = struct {
-		LoopbackAuth bool `json:"is_loopback_auth"`
-		MockHSM      bool `json:"is_mockhsm"`
-		Reset        bool `json:"is_reset"`
-		PlainHTTP    bool `json:"is_plain_http"`
+		LocalhostAuth bool `json:"is_localhost_auth"`
```
kr: Do you mind not changing this variable name? Please change as little as possible for now. 🙏

jeffomatic:
@tessr The build tag itself isn't changing in this commit. Was there something else you were referring to?

tessr:
@jeffomatic Ahh, sorry, no. I can't read 😳

core/config/config.go
jeffomatic: Are you sure? There are only three spots where this variable is mentioned. I think it will be less confusing if we can make the language as consistent as we can. For example, here are two of the references:

screen shot 2017-05-01 at 2 06 21 pm

screen shot 2017-05-01 at 2 06 11 pm

If I keep the existing variable name `LoopbackAuth`, then someone who is unfamiliar with this may think the discrepancy is a mistake.

kr: Yes, I would rather not touch those places if possible, and keep this change limited to where the output happens.

jeffomatic: Done, PTAL

kr:
LGTM

Thanks for https://github.com/chain/chain/pull/1103#discussion_r114200557, it was significantly easier to review this way. :)

PR #1104, opened 2017-05-01
database/raft: match behavior of crypto/tls

kr:
We need to verify just the hostname, not including the
colon and port number suffix.


tessr:
LGTM

PR #1105, opened 2017-05-01
database/raft: use correct HTTP client

kr:
DefaultClient (which uses the system cert pool) will
refuse to validate the peer's certificate.


tessr:
LGTM 

(\*silent screaming\*)

PR #1106, opened 2017-05-01
dashboard: support protected grant flag

dominic:
Persist individual grants in the redux store instead of just policy names
to allow access to the `protected` flag.

Protected grants are marked as `checked` and `disabled` when editing,
indicating to users that they cannot be removed.

### DIFF
--- a/dashboard/src/features/accessControl/components/EditPolicies.jsx
+++ b/dashboard/src/features/accessControl/components/EditPolicies.jsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { BaseNew, CheckboxField } from 'features/shared/components'
-import { policyOptions } from 'features/accessControl/constants'
+import { policyOptions, hasProtectedGrant } from 'features/accessControl/constants'
 import { reduxForm } from 'redux-form'
 import actions from 'features/accessControl/actions'
 import styles from './EditPolicies.scss'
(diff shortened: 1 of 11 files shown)


dominic:
screen shot 2017-05-01 at 3 09 33 pm
screen shot 2017-05-01 at 3 09 49 pm

dashboard/src/features/accessControl/selectors.js
```diff
@@ -2,17 +2,20 @@ import { createSelector } from 'reselect'
...
-  policies => policies.map(
-    policy => policyOptions.find(
-      elem => elem.value == policy
-    ).label
-  )
-)
+  item => item.grants,
+  grants => grants.map(
+    grant => {
+      let isProtected = grant.protected
```
jeffomatic: Super nit: if we're adding `const`, might as well make these two vars `const` as well?

dashboard/src/features/accessControl/components/EditPolicies.jsx
```diff
@@ -15,10 +15,14 @@ class EditPolicies extends React.Component {
     return(
       <div className={styles.main}>
         {policyOptions.map(option => {
+          const grant = this.props.item.grants.find(g => g.policy == option.value)
```
jeffomatic: I believe the backend implementation treats `{policy: "foo", protected: false}` and `{policy: "foo", protected: true}` as separate grants. While this is an unlikely edge case, I think it's a good idea to try to account for it here.

One possibility is to have the edit checkboxes correspond to the unprotected versions of the grants, and to list any protected grants that may appear under the guard as non-interactive "footnotes" in the editing interface.

dominic: Based on https://github.com/chain/chain/pull/1085#issuecomment-298082163 I expected that one couldn't create a non-protected version of a protected grant, nor would two ever co-exist. 

I'd like to hold off on that change until some protected grants have landed? It's really hard to test these right now and get a sense of what the right experience is.

PR #1108, opened 2017-05-01
database/raft: update debugging Makefile for current tool interfaces

kr:


database/raft/Makefile
```diff
@@ -8,21 +8,24 @@ build:
...
 	DATABASE_URL=postgres:///core0?sslmode=disable \
 	cored
 
 run1:
-	LISTEN=127.0.0.1:1998 \
+	LISTEN=localhost:1998 \
 	CHAIN_CORE_HOME=$(HOME)/.chaincore1 \
-	BOOTURL=http://127.0.0.1:1999/ \
+	ROOT_CA_CERTS=$(HOME)/.chaincore1/root_ca_certs \
+	BOOTURL=https://localhost:1999/ \
```
tessr: do we need https even though we're building with `plain_http`?

database/raft/Makefile
```diff
@@ -8,21 +8,24 @@ build:
...
-	LISTEN=127.0.0.1:1999 \
+	LISTEN=localhost:1999 \
 	CHAIN_CORE_HOME=$(HOME)/.chaincore0 \
+	ROOT_CA_CERTS=$(HOME)/.chaincore0/root_ca_certs \
 	DATABASE_URL=postgres:///core0?sslmode=disable \
 	cored
 
 run1:
-	LISTEN=127.0.0.1:1998 \
+	LISTEN=localhost:1998 \
```
tessr: is this changing based on how folks are expected to generate their certs?

database/raft/Makefile
kr: Oh I guess not. Will remove.

kr: Actually, I would rather remove plain_http from the build tags here if that's okay with you.

database/raft/Makefile
kr: Yes. Using the hostname should work either way.

kr:
> Can I suggest two separate sets of commands, for TLS and non-TLS chain core?

Yeah that's a good idea.

Another thought would be to try to stick to the default set of build tags (i.e. none) in the Makefile, and let us individually tweak this as necessary to test alternative configurations.

But I'm gonna withdraw this PR for now. We don't need to worry about it before we open the tree for 1.3 development.

kr:
I think this PR is stale by now. Closing.

PR #1109, opened 2017-05-01
database/raft: fix WAL snapshot after joining a cluster

kr:
We need to save the initial snapshot to the WAL
immediately when joining a cluster. Failing to do this
means subsequent entries are not saved until the next
snapshot, causing a crash on restart.

In the vendored raft package, on startup, the read
snapshot serves as the base index for the slice of
entries. The slice of entries is initially empty, and
entries are appended as they're read. Entries with an
index greater than the currently read entry are sliced
off.

The crash happens there. If we fail to save the initial
snapshot, then the base index will be 0 and the first
entry record has a nonzero index, causing an
out-of-bounds slice operation.

### DIFF
--- a/database/raft/raft.go
+++ b/database/raft/raft.go
@@ -182,13 +182,27 @@ func Start(laddr, dir, bootURL string, httpClient *http.Client, useTLS bool) (*S
 		return nil, err
 	}
 
-	if walobj != nil {
+	recover := walobj != nil
+	if recover {
 		sv.id, err = readID(sv.dir)
 		if err != nil {
 			return nil, errors.Wrap(err)
 		}
 	} else if bootURL != "" {
-		err = sv.join(laddr, bootURL) // sets sv.id and state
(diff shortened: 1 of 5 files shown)


database/raft/raft.go
```diff
@@ -208,23 +222,10 @@ func Start(laddr, dir, bootURL string, httpClient *http.Client, useTLS bool) (*S
 		Logger:          &raft.DefaultLogger{Logger: stdlog.New(ioutil.Discard, "", 0)},
 	}
 
-	if walobj != nil {
+	if recover {
 		sv.raftNode = raft.RestartNode(c)
 		triggerElection(ctx, sv)
```
tessr: so this is no longer causing a panic? Did you figure out what was going on?

kr: I updated the patch to preserve the existing behavior, so we do this when recovering from an existing WAL but not when joining.

PR #1115, opened 2017-05-01
cmd/corectl: new command to add allowed addresses

tessr:
... for raft. 

Also includes a fix to the raft package: previously, raft did not anticipate a failure partway through a node attempting to join the cluster, and so it created the WAL directory and the node ID in an order that did not accommodate retrying after a partial failure.


tessr:
PTAL

(puts on my 🚲 🏠  helmet :construction_worker_woman: )

kr:
LGTM

PR #1116, opened 2017-05-02
core: add self to raft member list

tessr:


core/run.go
```diff
@@ -131,6 +131,11 @@ func RunUnconfigured(ctx context.Context, db pg.DB, raftDB *raft.Service, opts .
 	for _, opt := range opts {
 		opt(a)
 	}
+	err := a.addAllowedMember(ctx, struct{ Addr string }{routableAddress})
```
kr: I think we need `if raftDB.IsNew()` in here somewhere (and in Run).

kr: Oh, huh, I guess actually it doesn't hurt to do it no matter what. So perhaps we should remove the new IsNew method?

PR #1117, opened 2017-05-02
cmd/corectl: add wait subcommand

jeffomatic:
This is a generic utility for blocking until cored is available.
A typical usecase would be as follows:

```
cored &
corectl wait
corectl config-generator
corectl create-token foobar client-readwrite
```

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -470,10 +470,8 @@ func wait(client *rpc.Client, args []string) {
 			break
 		}
 
-		if statusErr, ok := errors.Root(err).(rpc.ErrStatusCode); ok {
-			if statusErr.StatusCode/100 != 5 {
-				break
-			}
+		if statusErr, ok := errors.Root(err).(rpc.ErrStatusCode); ok && statusErr.StatusCode/100 != 5 {
+			break
 		}
 
 		time.Sleep(500 * time.Millisecond)
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -22,6 +22,7 @@ import (
 	"chain/core/rpc"
 	"chain/crypto/ed25519"
 	"chain/env"
+	"chain/errors"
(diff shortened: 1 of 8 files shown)


dominic:
Would it be too wild for every command to individually call `wait` before proceeding? Maybe we don't need to expose this as a public command.

kr:
> Would it be too wild for every command to individually call `wait` before proceeding?

Personally I would prob find that more confusing than helpful, a little too magic.

Like if my cored's just not running at all and I do `corectl foo`, I would much rather see an immediate "connection refused" message than it just hanging there for a long time.

cmd/corectl/main.go
```diff
@@ -437,3 +439,24 @@ func splitAfter2(s, sep string) (a, b string) {
...
+		fatalln("error: wait takes no args")
+	}
+
+	for {
+		err := client.Call(context.Background(), "/info", nil, nil)
+		if err == nil {
+			break
+		}
+
+		if statusErr, ok := errors.Root(err).(rpc.ErrStatusCode); ok {
```
kr: Nit: this could say `ok && statusErr.StatusCode/100 != 5` but nbd.

kr:
LGTM

jeffomatic:
Hmm, my hypothesis is that `wait` is for unattended/automated workflows, whereas it might be useful to be given an error message if you're trying to create an access token and you never started Chain Core in the first place.

@kr maybe you can confirm, but I think we've gotten rid of any grey area where `cored` can be running but still not available (due to raft bootstrapping etc.). So if you start cored and then manually run a corectl command, it should pretty much do the right thing.

kr:
> we've gotten rid of any grey area where `cored` can be running but still not available 

That's right. Once cored is listening (which happens almost immediately on startup, before anything else) it should definitely serve correct responses (which may block for a half a second before responding during startup).

PR #1119, opened 2017-05-02
bin/wait-and-configure: remove file

jeffomatic:
This command can be replaced with the following one-liner:

    sh -c "corectl wait; corectl config-generator"


jeffomatic:
PTAL

kr:
LGTM

PR #1120, opened 2017-05-02
docs: add allow-address to corectl docs

tessr:


tessr:
PTAL

chrisgarvin:
@tessr could you add the `allow-address` link to the Commands section up at line 53? Also looks like the `wait` link is not correct in that list at the moment

tessr:
@chrisgarvin done, PTAnotherL! ⛄️

PR #1121, opened 2017-05-02
net/http/authn: return 401 when loopback_auth is off

kr:
This is a temporary workaround until we can more
carefully plan how to treat authn vs authz in various
clients such as Dashboard. Currently, we want to return
a 401 to indicate that the user needs to supply some
credentials, but we want to avoid returning that error
for requests to known-public resources (/dashboard/).

In the future, we should avoid ever returning 401,
internally treating authn as a set of orthogonal
credentials, which may legitimately be empty. (An empty
set of credentials is still useful when accessing public
resources.) Clients should then detect the "needs to
prompt for credentials" situation by requesting access
to a known lowest-common-denominator resource (most
likely /info) and looking for a 403 response to that
request.

### DIFF
--- a/net/http/authn/authn.go
+++ b/net/http/authn/authn.go
@@ -58,8 +58,12 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 		ctx = newContextWithLocalhost(ctx)
 	}
 
-	// Temporary workaround. See loopbackOn comment above.
-	if want := loopbackOn || strings.HasPrefix(req.URL.Path, "/dashboard/"); want && local {
+	// Temporary workaround. Dashboard is always ok.
+	// See loopbackOn comment above.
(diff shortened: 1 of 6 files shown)


net/http/authn/authn.go
```diff
@@ -51,9 +58,14 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 		ctx = newContextWithLocalhost(ctx)
 	}
 
+	// Temporary workaround. See loopbackOn comment above.
+	if want := loopbackOn || strings.HasPrefix(req.URL.Path, "/dashboard/"); want && local {
```
dominic: Should this be just `"/dashboard" to catch the URL without the trailing slash?

kr: Yes. Updated.

kr:
PTAnotherL

This wasn't doing the right thing for non-localhost requests. I reworked the logic (and added the check for just `/dashboard` with no trailing slash).

tessr:
LGTM

PR #1123, opened 2017-05-02
cmd/cored: use correct HTTP client

kr:
When connecting to signerd we need to use the trusted
root cert pool that was configured, so we can't use the
DefaultTransport in package net/http.


kr:
PTAL

jbowens:
lgtm

PR #1124, opened 2017-05-02
cmd/vet: add defaulthttp check

jbowens:
Add check to cmd/vet to check for use of the http.DefaultClient and
its package-level convenience functions.

### DIFF
--- a/cmd/vet/defaulthttp.go
+++ b/cmd/vet/defaulthttp.go
@@ -80,11 +80,15 @@ func checkHttpCall(f *File, node ast.Node) {
 	if !ok {
 		return
 	}
-	if obj.Pkg() == nil || obj.Pkg().Scope().Lookup(obj.Name()) != obj {
+	if !isPackageLevel(obj) {
 		return
 	}
 	if obj.Pkg().Path() != "net/http" || obj.Name() != "DefaultClient" {
 		return
 	}
 	f.Badf(call.Pos(), "use of net/http.DefaultClient.%s", fun.Name())
+}
+
+func isPackageLevel(obj types.Object) bool {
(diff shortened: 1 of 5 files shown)


cmd/vet/defaulthttp.go
```diff
@@ -0,0 +1,94 @@
...
+		return
+	}
+	obj, ok := f.pkg.uses[clientSel.Sel]
+	if !ok {
+		return
+	}
+	if !isPackageLevel(obj) {
+		return
+	}
+	if obj.Pkg().Path() != "net/http" || obj.Name() != "DefaultClient" {
```
kr: I think this would not find cases where we've assigned DefaultClient to another variable, right?

```go
x := http.DefaultClient
x.Post(...)
```

That is probably fine to start, but there's one way in which this pattern is likely to come up: when we're using DefaultClient as a default value for an HTTP client, like in https://github.com/chain/chain/blob/6d40c894/core/rpc/rpc.go#L126-L130:

```go
client := c.Client
if client == nil {
	client = http.DefaultClient
}
resp, err := client.Do(req.WithContext(ctx))
```

One way to thoroughly check for this would be to use https://godoc.org/golang.org/x/tools/go/pointer

```go
config := &pointer.Config{
	// ...
}
config.AddQuery(/* http.DefaultClient */)
config.AddQuery(/* obj */)
result, err := pointer.Analyze(config)
objPtr := result.Queries[/* obj */]
defPtr := result.Queries[/* http.DefaultClient */]
if objPtr.MayAlias(defPtr) {
	fail()
}
```

However, I think doing the analysis separately like this for every call site would be too slow, and even if we did one big batch analysis it still might be too slow.

kr:
LGTM (but maybe add a TODO about the pointer analysis)

PR #1125, opened 2017-05-02
core: provide even # of arguments to Printkv

jbowens:
Printkv expects an even number of arguments.


kr:
> Should we fail CI builds if our cmd/vet tool picks up something?

Heh, that was the intention all along, but maybe it stopped wercking at some point.

PR #1126, opened 2017-05-02
cmd/corectl: expose cored error messages

### LINKED ISSUE #1078: corectl does not provide useful error messages
Running an invalid corectl command, such as `corectl create-token foobar invalidpolicy` yields a generic error message (`400 Bad Request`). The error message and detail contained in the body of the response should be relayed to user.

jeffomatic:
This commit allows corectl to report formatted error data (chain code,
message, and detail) to the command-line user.

Fixes #1078

### DIFF
--- a/core/rpc/rpc.go
+++ b/core/rpc/rpc.go
@@ -140,6 +140,8 @@ func (c *Client) CallRaw(ctx context.Context, path string, request interface{}) 
 	}
 
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		defer resp.Body.Close()
+
 		resErr := ErrStatusCode{
 			URL:        cleanedURLString(u),
 			StatusCode: resp.StatusCode,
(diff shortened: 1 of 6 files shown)


jeffomatic:
Example output:

```
$ corectl create-token tok-1 foooobar
tok-1:58dd068d30385a0e4afce07242a5214c8d2568421953a591e9daf9affde133bd
Auth grant error:
RPC error: CH003 Invalid request body
Detail: invalid policy: foooobar
$ corectl create-token tok-1 foooobar
RPC error: CH302 Access token id is already in use
Detail: id "tok-1" already in use
```

jeffomatic:
PTAL

jeffomatic:
This will help debugging some of the corectl flows described in our (possibly stale) documentation.

kr:
:eyes:

core/rpc/rpc.go
```diff
@@ -138,11 +140,20 @@ func (c *Client) CallRaw(ctx context.Context, path string, request interface{})
...
 			StatusCode: resp.StatusCode,
 		}
+
+		// Attach formatted error message, if available
+		var errData httperror.Response
+		err := json.NewDecoder(resp.Body).Decode(&errData)
+		if err == nil && errData.ChainCode != "" {
+			resErr.ErrorData = &errData
+		}
+		resp.Body.Close()
```
kr: This is unlikely to matter but can you put this call back at the top of the if body in a defer? That way an inadvertent panic in between  there and here won't leak a connection.

kr:
LGTM with a small suggestion

PR #1127, opened 2017-05-02
all: fix use of %s format string with bc.Hash

jbowens:
These are picked up by go vet. The pointer has a String
method, the value does not. Unfortunately, this makes
non-addressable bc.Hashes a little difficult to format.

I opted for calling the value-receiver method `Bytes` and
using the `%x` format specifier here. Another option is to
keep the `HexString` method and prefer `HexString` to
`String` so that we can use it when we have a non-pointer
value. Thoughts?


erykwalder:
LGTM

PR #1129, opened 2017-05-02
docker/ci: Pin rust crate versions

tonychain:
Pin the versions of all crates to the ones originally installed when rust was
added to the docker image.

This makes the build reproducible so future versions of crates won't break the
docker image build (like they're doing now)


tonychain:
I built a new Docker image with this: `chaindev/ci:20170502`

The wercker.yml has been updated to use it.

bobg:
LGTM

PR #1130, opened 2017-05-02
docker/de: fix startup.sh token generation

jeffomatic:
Now that corectl does not talk to the data store directly, we need the Docker startup script to wait for cored to become available before generating an access token.

Example output:

    Chain Core is online!
    Listening on: http://localhost:1999
    Autogenerating acccess token with client-readwrite privileges...
    Copy the whole line below:
    client:965aca3b62a068bf57af2302c621359237116ee6683f17fd5ca09d86645886ed


jeffomatic:
PTAL

PR #1131, opened 2017-05-02
protocol/state: fix duplicate nonce bug

### LINKED ISSUE #1128: protocol/state: duplicate, overlapping nonces possible
The tx max and min times should not be used for nonce validation, but rather the time range associated with the nonce itself because the nonce time range can be greater than and encompass the tx time range.

Additionally, just comparing the max times is not sufficient. The nonces should not be able to overlap, so the max time of the nonce in the state should be less than the min time of the nonce in the transaction.

It may be worth not even doing the time comparison, but instead checking presence and relying on block timestamp based expiration.

Relevant Code:
https://github.com/chain/chain/blob/23f405c000f934ffe264781d8337a69916ab4028/protocol/state/snapshot.go#L74-L81

erykwalder:
Changes the snapshot to store and check the max time of the nonce,
rather than the tx maxtime. Also enforces the nonce uniqueness with
the block timestamp.

Fixes #1128

### DIFF
--- a/generated/rev/RevId.java
+++ b/generated/rev/RevId.java
@@ -1,4 +1,4 @@
 
 public final class RevId {
-	public final String Id = "main/rev3090";
+	public final String Id = "main/rev3091";
 }
--- a/generated/rev/revid.go
+++ b/generated/rev/revid.go
@@ -1,3 +1,3 @@
 package rev
 
-const ID string = "main/rev3090"
+const ID string = "main/rev3091"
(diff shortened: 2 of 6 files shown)


bobg:
I'd like to resolve https://github.com/chain/chain/pull/728#discussion_r114406260 before landing this.

erykwalder:
Agreed

bobg:
I guess https://github.com/chain/chain/issues/1128#issuecomment-298768833 is a resolution of https://github.com/chain/chain/pull/728#discussion_r114406260. (Right, @danrobinson?)

LGTM

PR #1138, opened 2017-05-02
sdk/java: add javadoc to CoreConfig class

dominic:


sdk/java/src/main/java/com/chain/api/CoreConfig.java
```diff
@@ -11,67 +11,135 @@
...
     public String state;
 
+    /**
+     * Whether the core has been configured.
+     */
     @SerializedName("is_configured")
     public boolean isConfigured;
 
+    /**
+     * RFC3339 timestamp reflecting when the core was configured.
```
jeffomatic: This is deserialized to a `Date`, so it's not RFC3339 in Java. Ditto for other instances of "RFC3339" in the comments.

sdk/java/src/main/java/com/chain/api/CoreConfig.java
```diff
@@ -11,67 +11,135 @@
...
     public String blockchainId;
 
+    /**
+     * Height of the blockchain in the local core.
+     */
     @SerializedName("block_height")
     public long blockHeight;
 
+    /**
+     * Height of the blockchain in the generator
```
jeffomatic: period

sdk/java/src/main/java/com/chain/api/CoreConfig.java
```diff
@@ -11,67 +11,135 @@
...
     @SerializedName("build_config")
     public BuildConfig buildConfig;
 
+    /**
+     * Blockchain error information.
+     */
     public Health health;
 
+    /**
+     *
```
jeffomatic: Might as well make something up here.

sdk/java/src/main/java/com/chain/api/CoreConfig.java
```diff
@@ -94,53 +162,103 @@
     }
   }
 
+  /**
+   * Get info on specified Chain Core.
```
jeffomatic: I think we use the third-person present in Java comments, so "Gets" instead of "Get". Ditto elsewhere.

sdk/java/src/main/java/com/chain/api/CoreConfig.java
dominic: Oops. Meant to remove this. We don't document this key anywhere right now so I want to leave it blank.

PR #1139, opened 2017-05-02
net/http/authn: make client certs more optional

kr:
The std library package crypto/tls provides several
options for ClientAuthType; see
https://godoc.org/crypto/tls#ClientAuthType. None of
these do what we want, which is to request a client
cert, and validate it if given, but to treat an invalid
cert the same as a missing client cert: the connection
should be permitted but the cert should not be used in
authz.

We can fix this by using RequestClientCert and calling
Verify ourselves in the authn package. This also
requires plumbing to give authn access to the cert pool.

### DIFF
--- a/net/http/authn/authn.go
+++ b/net/http/authn/authn.go
@@ -2,6 +2,7 @@ package authn
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/hex"
 	"net"
 	"net/http"
@@ -24,6 +25,7 @@ var loopbackOn = false
 type API struct {
 	tokens             *accesstoken.CredentialStore
 	crosscoreRPCPrefix string
+	rootCAs            *x509.CertPool
 
 	tokenMu  sync.Mutex // protects the following
 	tokenMap map[string]tokenResult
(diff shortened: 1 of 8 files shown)


net/http/authn/authn.go
```diff
@@ -78,9 +81,28 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
...
+			CurrentTime:   time.Now(),
+			Intermediates: x509.NewCertPool(),
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		}
+		for _, cert := range certs[1:] {
+			opts.Intermediates.AddCert(cert)
+		}
+		_, err := certs[0].Verify(opts)
+		if err != nil {
+			// errors.New("tls: failed to verify client's certificate: " + err.Error())
```
kr: I'll replace that with an explanatory comment.

net/http/authn/authn.go
```diff
@@ -78,9 +81,28 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 
 // checks the request for a valid client cert list.
 // If found, it is added to the request's context.
```
tessr: it might be nice to make this function comment a little more comprehensive now

kr:
PTAL

PR #1149, opened 2017-05-03
desktop/windows: fix README for 1.2

tessr:
This clarifies that the README is mostly intended to be used for building the installer from source. It also fixes the paths used in the step-by-step build instructions and adds the necessary build tags to the compilation step.


tessr:
PTAL

tessr:
Actually, !PTAL. I want to fix one more thing.

tessr:
Okay now PTAL

kr:
LGTM

PR #1151, opened 2017-05-03
docs: add cored and supporting documentation

dominic:


docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+* **SPLUNKADDR**
+* **LOGFILE**
+* **LOGSIZE**
+* **LOGCOUNT**
+* **LOG_QUERIES**
+* **MAXDBCONNS**
+* **RATELIMIT_TOKEN**
+* **RATELIMIT_REMOTE_ADDR**
+* **INDEX_TRANSACTIONS**
+* **BOOTURL**
```
kr: Setting BOOTURL causes the cored process to join an existing Chain Core cluster if it's not already a member. (it it is already a member of a cluster, BIOTURL has no effect.)

This should be the url of any cored process already in the cluster, or it can also be a load balancer that forwards requests to any node.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults to `:1999`
+* **DATABASE_URL**: URL of the Postgres database, defaults to `postgres:///core?sslmode=disable`
+* **CHAIN_CORE_HOME**: Path to Chain Core data directory on local file system, defaults to `$HOME/.chaincore`
+* **ROOT_CA_CERTS**
+* **SPLUNKADDR**
+* **LOGFILE**
+* **LOGSIZE**
+* **LOGCOUNT**
+* **LOG_QUERIES**
+* **MAXDBCONNS**
```
kr: this limits the number of simultaneous connections to postgres. it is a good idea to set it to something larger than the default in production, something like 100 or 200. the default is 10.

but we should reevaluate the default. we should also make new measurements about what the recommended setting would be for production.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+* **RATELIMIT_TOKEN**
```
kr: rate limit token limits the number of requests per second for each access token. a client attempting to send more than that will be served 4xx responses.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+* **RATELIMIT_REMOTE_ADDR**
```
kr: this is like ratelimit token but the limit applies to each remote ip  address. it is okay to set both of these in which case both limits will apply

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults to `:1999`
+* **DATABASE_URL**: URL of the Postgres database, defaults to `postgres:///core?sslmode=disable`
+* **CHAIN_CORE_HOME**: Path to Chain Core data directory on local file system, defaults to `$HOME/.chaincore`
+* **ROOT_CA_CERTS**
+* **SPLUNKADDR**
+* **LOGFILE**
+* **LOGSIZE**
+* **LOGCOUNT**
+* **LOG_QUERIES**
```
kr: log queries should be used only for debugging, and not in production. it prints every sql query executed. this env var may be removed in the future. do not use it.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+* **-version**: Print version and build information
+
+## Environment Variables
+
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults to `:1999`
+* **DATABASE_URL**: URL of the Postgres database, defaults to `postgres:///core?sslmode=disable`
+* **CHAIN_CORE_HOME**: Path to Chain Core data directory on local file system, defaults to `$HOME/.chaincore`
+* **ROOT_CA_CERTS**
+* **SPLUNKADDR**
+* **LOGFILE**
```
kr: logfile sets the location of the base file for chain core log output. if unset, the log will print to stderr. if set to file x, cored will write to x and rotate x to x.1, x.2, etc

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+* **LOGSIZE**
```
kr: logsize limits the size of the base log file

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+* **LOGCOUNT**
```
kr: logcount limits the number of rotated log files to keep. the default is 9

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+
+* **-version**: Print version and build information
+
+## Environment Variables
+
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults to `:1999`
+* **DATABASE_URL**: URL of the Postgres database, defaults to `postgres:///core?sslmode=disable`
+* **CHAIN_CORE_HOME**: Path to Chain Core data directory on local file system, defaults to `$HOME/.chaincore`
+* **ROOT_CA_CERTS**
+* **SPLUNKADDR**
```
kr: splunkaddr is the address of a splunk  log sink server to send logs to. we should probably not document this because i don't think we want to suppprt it

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+## Flags
+
+* **-version**: Print version and build information
+
+## Environment Variables
+
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults to `:1999`
+* **DATABASE_URL**: URL of the Postgres database, defaults to `postgres:///core?sslmode=disable`
+* **CHAIN_CORE_HOME**: Path to Chain Core data directory on local file system, defaults to `$HOME/.chaincore`
+* **ROOT_CA_CERTS**
```
kr: root ca certs is the filepath of a file containing a set of pem-encoded concatenated root ca certificates to trust. if unset, cored will *not* use the system roots, it will trust no CA certs. 

it is also okay to put leaf certs in this file to "pin" them, to trust exactly those certs as valid.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,54 @@
...
+go build -tags 'loopback_auth plain_http reset' chain/cmd/cored
+```
+
+## Flags
+
+* **-version**: Print version and build information
+
+## Environment Variables
+
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults to `:1999`
```
kr: in a multi process configuration, this must be a reachable, routable address.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,99 @@
...
+The following variables control the location of Chain Core's data stores,
+and its routable address.
+
+* **CHAIN_CORE_HOME**: Path to Chain Core data directory on local file system,
+defaults to `$HOME/.chaincore`.
+
+* **DATABASE_URL**: URL of the Postgres database, defaults
+to `postgres:///core?sslmode=disable`.
+
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults
```
kr: Oh this is just `LISTEN`, not `LISTEN_ADDR`.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,99 @@
...
+* **LOGSIZE**: Size limit of the base log file in bytes, defaults to 5MB.
+
+* **LOGCOUNT**: Number of rotated log files to keep, defaults to 9.
+
+* **MAXDBCONNS**: Maximum number of simultaneous connections to Postgres from
+Chain Core, defaults to 10.
+
+* **RATELIMIT_TOKEN**: Maximum number of requests-per-second
+allowed with an individual access token. Requests made beyond
+the limit will receive a `4xx` response.
```
kr: Sorry I was on my phone before and didn't bother looking up the actual response code. It's HTTP 429, CH007.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,99 @@
...
+
+    Can be stacked with **RATELIMIT_REMOTE_ADDR**.
+
+* **RATELIMIT_REMOTE_ADDR**: Maximum number of requests-per-second
+allowed fro a remote IP address. Requests made beyond
+the limit will receive a `4xx` response.
```
kr: Same code for this one, 429, CH007.

docs/core/reference/cored.md
```diff
@@ -0,0 +1,99 @@
...
+
+    Can be stacked with **RATELIMIT_TOKEN**.
+
+* **BOOTURL**: Setting value causes the `cored` process to join an existing
```
kr: Setting *this* value?

docs/core/reference/cored.md
```diff
@@ -0,0 +1,99 @@
...
+Chain Core cluster if it's not already a member. If it is already a member
+of a cluster, this has no effect.
+
+    This should be the url of any `cored` process already in the cluster,
```
kr: Nit: "URL" (caps)

docs/core/reference/cored.md
```diff
@@ -0,0 +1,99 @@
...
+* **LISTEN_ADDR**: Address the Chain Core server will listen on, defaults
+to `:1999`. In a multi process configuration, this must be a
+reachable, routable address.
+
+### Extended Functionality
+
+The following variables allow for more fine-grained operation control of
+Chain Core, such as log rotation, rate limits, and clustering.
+
+* **ROOT_CA_CERTS**: Path to file containing a set of pem-encoded concatenated
```
kr: Nit: "PEM" (caps) (my bad for being sloppy in my previous feedback, sorry)

docs/core/reference/cored.md
```diff
@@ -0,0 +1,99 @@
...
+root CA certificates to trust. If unset, `cored` trust no CA certs. See the
```
kr: Nit: "trusts" or "will trust"

docs/core/reference/cored.md
```diff
@@ -0,0 +1,96 @@
+# cored Command
+
+Command `cored` provides the Chain Core daemon and API server.
+
+## Installation
+
+`cored` is included with all desktop installations of Chain Core
+Developer Edition, and is run by default when starting Mac, Windows and Docker.
+
+### From source
```
kr: Do we really want to describe building from source here? Is this document for users of cored or for developers of cored?

I feel like it might be better to keep this file focused on users' needs and leave out all mention of how to build cored. (Building is something only we have to worry about. That's internal documentation.)

docs/core/reference/cored.md
```diff
@@ -0,0 +1,95 @@
...
+Command `cored` provides the Chain Core daemon and API server.
+
+`cored` is included with all official releases of Chain Core
+Developer Edition, and is run by default when starting Mac, Windows and Docker.
+
+If you plan to make changes to the internals of Chain Core, you can find
+build instructions in our [GitHub readme](https://github.com/chain/chain/blob/main/Readme.md#building-from-source);
+if you want to _use_ chain core, please download one of our official releases.
+
+### Build tags
```
jeffomatic: I think the build tags are non-sequitor now that we've moved compile instructions away from this doc?

docs/core/reference/cored.md
```diff
@@ -0,0 +1,95 @@
...
+* **LISTEN**: Address the Chain Core server will listen on, defaults
+to `:1999`. In a multi process configuration, this must be a
+reachable, routable address.
+
+### Extended Functionality
+
+The following variables allow for more fine-grained operation control of
+Chain Core, such as log rotation, rate limits, and clustering.
+
+* **ROOT_CA_CERTS**: Path to file containing a set of PEM-encoded concatenated
```
jeffomatic: I changed the section titles from "XYZ Auth" to "XYZ authentication". So the hashlinks will need to get updated too. Sorry 😬

dominic: nbd thx!

docs/core/reference/cored.md
dominic: Sorta, but they are printed by Chain Core at startup, and in `cored -version`. I realize it's duplication from the README, but I think it's important for users to understand what the core is telling them without having to reference GitHub.

dominic: Should we call them something else here? "Feature Flags"?

docs/core/reference/cored.md
```diff
@@ -2,14 +2,14 @@
...
+`cored` is included with all official releases of Chain Core
 Developer Edition, and is run by default when starting Mac, Windows and Docker.
 
-### From source
+If you plan to make changes to the internals of Chain Core, you can find
+build instructions in our [GitHub readme](https://github.com/chain/chain/blob/main/Readme.md#building-from-source);
+if you want to _use_ chain core, please download one of our official releases.
 
-_The instructions in this section require having the Go programming environment installed and the PATH variable correctly configured. See the [Chain Core Readme file](https://github.com/chain/chain/blob/main/Readme.md#building-from-source) for details._
+### Build tags
```
kr: So if we're going to refer people to the readme to build it, would it be possible to omit any mention of the build tags in this doc?

I would slightly prefer to describe the behavior differences between developer edition and the server build in terms of user-visible things, e.g. "Chain Core Developer Edition has some features designed for convenience during development and testing: it has a reset button to delete all data, it permits unauthenticated requests over the loopback device (localhost), it has a mock HSM, and it allows unencrypted (non-TLS) HTTP requests. These features are unsafe for use in production and are disabled in Chain Core Enterprise Edition."

(I don't know if "Chain Core Enterprise Edition" is exactly the right name for that build, but you get the idea.)

docs/core/reference/cored.md
jeffomatic: What is shown at startup or `-v` are the weird inversions of the build tags, so it shows `mockhsm` and not `no_mockhsm`. There's one level of indirection there. My 2c would be to punt on presenting these in the doc, but if you anticipate a ton of confusion, we should reflect what is actually printed out.

PR #1152, opened 2017-05-03
cmd/cored: rename plain_http to http_ok

kr:


kr:
Consensus: we are going to go with `http_ok`.

tessr:
🍦

kr:
PTAL

PR #1159, opened 2017-05-03
cmd/cored: rename plain_http to http_ok (fixup)

kr:
The previous commit missed some places.


kr:
PTAL

dominic:
@kr Added Dashboard support

PR #1160, opened 2017-05-03
cmd/cored: turn reset on by default

kr:
The build tag is now "no_reset" to turn it off.


kr:
Do I need to update a test.

PR #1169, opened 2017-05-04
core: don't create self-grant for raft member list if not using TLS

tessr:
This fixes a bug where a single internal grant was being created with an empty certificate for cores that were being run without a TLS config.

### DIFF
--- a/core/membership.go
+++ b/core/membership.go
@@ -26,25 +26,28 @@ func (a *API) addAllowedMember(ctx context.Context, x struct{ Addr string }) err
 		return errors.Wrap(err)
 	}
 
-	data := map[string]interface{}{
-		"subject": map[string]string{
-			"CN": hostname,
-		},
-	}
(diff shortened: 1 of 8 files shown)


tessr:
PTAL @kr especially; I'm not sure if adding yet another field to the API object is in the best taste.

kr:
:eyes:

core/membership.go
```diff
@@ -26,25 +26,28 @@ func (a *API) addAllowedMember(ctx context.Context, x struct{ Addr string }) err
...
 
-	data := map[string]interface{}{
-		"subject": map[string]string{
-			"CN": hostname,
-		},
-	}
-
-	guardData, err := json.Marshal(data)
-	if err != nil {
+	if a.useTLS {
```
kr: also maybe consider `if !a.useTLS { return nil }` here?

PR #1171, opened 2017-05-04
all: tick version to 1.2 release candidate 2

jeffomatic:


jeffomatic:
Will land this once we're through the last round of QA and bugfixes.

PR #1172, opened 2017-05-04
core/coreunsafe: delete grants when resetting a core

tessr:


tessr:
PTAL before jeff finishes eating lunch

core/coreunsafe/reset.go
```diff
@@ -74,6 +75,16 @@ func ResetEverything(ctx context.Context, db pg.DB, rDB *raft.Service) error {
...
 
+	// Delete all grants in raft storage
+	for _, p := range core.Policies {
+		err = rDB.Delete(ctx, core.GrantPrefix+p)
+		if err != nil {
+			return errors.Wrapf(err, "could not delete grants for policy %s from RaftDB", p)
+		}
+	}
+
+	// TODO(tessr): remove this core from the allowed members list in raft, too, potentially
```
kr: Maybe we should just delete the whole allowed members list? If reset is only available on single-node clusters it prob doesn't matter but maybe we should delete it just to be consistent?

tessr: We don't really have a good way to delete the whole list right now, but I'll update this TODO to be clearer.

PR #1174, opened 2017-05-05
database/sqlutil: log queries in a driver (new package)

kr:
Package sqlutil provides a SQL driver that logs every
query. This approach is a bit easier than our previous
approach of wrapping all standard-library SQL functions.
The old approach is now unnecessary, and package
chain/database/sql is now deleted.


database/sqlutil/logdriver.go
```diff
@@ -0,0 +1,87 @@
...
+	queryer, ok := lc.Conn.(driver.Queryer)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+	logQuery(context.Background(), query, args)
+	return queryer.Query(query, args)
+}
+
+// TODO(kr): implement context variants
+// (but don't bother until lib/pq does first.
```
kr: typo

PR #1180, opened 2017-05-11
database/pg/pgtest: make useful outside of package testing

kr:
This package doesn't use nearly all of the methods of
type testing.TB. By defining a new interface containing
just what we need here, we let the caller supply their own
object. In particular, *log.Logger satisfies Fataler.


kr:
PTAL

PR #1182, opened 2017-05-11
sdk/java: relax visibility of Transaction.Builder properties

jeffomatic:
Moving visibility from private to protected allows derived classes to
access the builder internals for custom client-side behavior.


erykwalder:
LGTM

PR #1184, opened 2017-05-12
cmd/report: report test results to S3

kr:
The new command "report" executes its arguments, then uploads the result to S3 and adds a summary of the result to a history table for a convenient web-accessible overview.

Future work: announce in Slack when a test fails.

### DIFF
--- a/cmd/report/doc.go
+++ b/cmd/report/doc.go
@@ -1,0 +1,20 @@
+/*
+
+Command report executes its argument as a child process
+and reports the results to S3.
+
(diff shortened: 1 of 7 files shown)


kr:
PTAL

This depends on a newer version of the AWS Go SDK, so it won't build just yet.

Also, I'm not wild about `report` as a command name. 🚲 🏠  encouraged.

kr:
Rebased; PTAL.

jbowens:
lgtm

PR #1189, opened 2017-05-13
cmd/cored: set feature flags consistently

kr:
This affects two features: mockhsm and reset (with
corresponding build tags no_mockhsm and no_reset).
The rule is: feature flags in chain/core/config always
default to false; build tags in chain/cmd/cored must set
them to true when the feature is enabled. This can be
either because the build tag is present, e.g. http_ok,
or absent, e.g. no_reset.

This also fixes a bug. Previously, building with
no_reset would incorrectly report reset=true in the
cored version output. A file in cored was explicitly
setting it to true when the feature was enabled, and the
default value hard coded in package config was also
true, but nothing ever set it to false when the feature
was disabled.

### DIFF
--- a/cmd/cored/mockhsm.go
+++ b/cmd/cored/mockhsm.go
@@ -5,9 +5,14 @@ package main
 import (
 	"chain/core"
 	"chain/core/blocksigner"
+	"chain/core/config"
 	"chain/core/mockhsm"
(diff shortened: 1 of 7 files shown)


kr:
PTAL

bobg:
LGTM

PR #1192, opened 2017-05-15
database/sinkdb: new package

jbowens:
Split database/raft into two packages, one handling raft-specific logic,
the other exposing a consistent kv store interface.

database/raft manages the raft.Node and exposes http endpoints for
communication between raft nodes.

database/sinkdb implements a consistent kv store on top of
database/raft. For now, committed kv data is stored in a simple protocol
buffer persisted to the filesystem.


core/config/config.go
```diff
@@ -57,53 +55,47 @@ var (
...
-// it will try Postgres next. If it finds a config in Postgres but not in raft
-// storage, the config will be added to raft storage.
-func Load(ctx context.Context, db pg.DB, rDB *raft.Service) (*Config, error) {
-	data, err := rDB.Get(ctx, "/core/config")
+// It will first try to load the config from sinkdb; if that fails,
+// it will try Postgres next. If it finds a config in Postgres but not in sinkdb
+// storage, the config will be added to sinkdb.
+func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
+	c := new(Config)
+	found, err := sdb.Get(ctx, "/core/config", c)
```
tessr: ✨  pretty!

database/raft/raft.go
```diff
@@ -83,14 +81,28 @@ type Service struct {
...
-	state     *state.State
+	state     State
 	confState raftpb.ConfState
 	done      bool
 
 	// Current log position, accessed only from runUpdates goroutine
 	snapIndex uint64
 }
 
+type State interface {
```
tessr: why is this an interface? is there more than one implementation of this type?

jbowens: to circumvent a bidirectional dependency. I was hoping that database/raft would only be raft-specific logic and completely ignorant of its use in implementing a kv store, but it needs a way to access and modify the raft state. This interface keeps key-value logic out of the raft package but still lets it interact with the key-value state.

database/raft/raft.go
```diff
@@ -442,31 +426,29 @@ func (sv *Service) allocNodeID(ctx context.Context) (uint64, error) {
...
 		if isTimeout(err) {
 			continue
 		}
-		return resp, err
+		return err
 	}
 }
 
-func (sv *Service) get(ctx context.Context, key string) ([]byte, error) {
+func (sv *Service) linearizeRead(ctx context.Context) error {
```
tessr: I know this is wordier, but could we call this `linearizableRead` instead? `linearizeRead` sounds like we are taking a read and linearizing it, whatever that means.

tessr: Oh, wait, I suppose that's actually what we're doing here. _nevermind_

database/sinkdb/sinkdb.go
```diff
@@ -0,0 +1,132 @@
...
+	buf := db.state.get(key)
+	if len(buf) == 0 {
+		return false, err
+	}
+	return true, proto.Unmarshal(buf, v)
+}
+
+// GetInconsistent performs a non-linearizable read of the provided key.
+// The value may be stale. The read value is unmarshalled into v.
+func (db *DB) GetInconsistent(key string, v proto.Message) (found bool, err error) {
```
tessr: 🚲  I prefer the word "stale" to "inconsistent" but 🚲

database/sinkdb/sinkdb.go
```diff
@@ -0,0 +1,132 @@
...
+		Type:  sinkpb.Op_SET,
+		Key:   key,
+		Value: encodedValue,
+	})
+	return op
+}
+
+// Exec executes the provided operations. If all of the provided conditionals
+// are met, all of the provided effects are applied atomically.
+func (db *DB) Exec(ctx context.Context, ops ...Op) error {
```
tessr: Does this function get us enough atomicity to do things like migrate multiple tables from pg? (e.g. right now we migrate config and access tokens separately, but that should be atomic)

database/sinkdb/sinkdb.go
```diff
@@ -0,0 +1,132 @@
...
+	buf := db.state.get(key) // read directly from state
+	if len(buf) == 0 {
+		return false, err
+	}
+	return true, proto.Unmarshal(buf, v)
+}
+
+// AddAllowedMember configures sinkdb to allow the provided address
+// to participate in Raft.
+func (db *DB) AddAllowedMember(ctx context.Context, addr string) error {
```
tessr: Is there a reason that `isAllowedMember` is on the state and `AddAllowedMember` is not? Is it just because `AddAllowedMember` needs to be exported?

database/sinkdb/state.go
```diff
@@ -132,84 +133,52 @@ func (s *State) Apply(data []byte, index uint64) (satisfied bool, err error) {
...
+			Type:  sinkpb.Op_SET,
 			Key:   nextNodeID,
 			Value: proto.EncodeVarint(oldID + 1),
 		}},
 	})
+	return instruction
+}
 
-	return b
+func (s *state) EmptyWrite() (instruction []byte) {
```
tessr: this is great

net/http/authz/authorizer.go
```diff
@@ -20,16 +18,16 @@ var ErrNotAuthorized = errors.New("not authorized")
 var builtinGrants = []*Grant{{GuardType: "any", Policy: "public"}}
 
 type Authorizer struct {
-	raftDB      *raft.Service
-	raftPrefix  string
+	sdb         *sinkdb.DB
+	kvPrefix    string
```
tessr: What about `keyPrefix`?

database/raft/raft.go
tessr: Ah, got it. Clever.

tessr:
Just a few 🚲 -type comments, but this is starting to look really good!

database/sinkdb/sinkdb.go
jbowens: Oo, I think so. We should be able to pass in all the Set operations and a single `sinkdb.IfNotExists("/core/config")`.

database/sinkdb/sinkdb.go
jbowens: Yeah, it needs to be exported and it needs to access raft. The `state` struct doesn't have access to the `*raft.Service` struct because the `state` struct is created and passed in to `raft.Start`

database/raft/raft.go
```diff
@@ -442,31 +426,29 @@ func (sv *Service) allocNodeID(ctx context.Context) (uint64, error) {
...
-// Set won't have changed the value again, but it is
+// RequestRead requests a linearizable read. Upon successful return,
+// the caller is guaranteed that a read from the Service's state will
+// be linearizable; this is, if a Set happens before a Get, the Get
+// will observe the effects of the Set. (There is still no guarantee
+// an intervening Set won't have changed the value again, but it is
 // guaranteed not to read stale data.)
-// This can be slow; for faster but possibly stale reads, see Stale.
-func (sv *Service) Get(ctx context.Context, key string) ([]byte, error) {
+func (sv *Service) RequestRead(ctx context.Context) error {
```
kr: Nit: if you're happy with this name then cool, but if not then consider `WaitRead`.

database/sinkdb/sinkdb.go
jbowens: 🚲  I co-opted the terminology from cockroach's consistency levels: https://github.com/cockroachdb/cockroach/issues/343#issuecomment-75370498

I get the point Spencer's making since in my cases, the value will not actually be stale, but I'm not sure if "inconsistent" is better. ¯\\\_(ツ)\_/¯

database/raft/raft.go
```diff
@@ -442,31 +426,29 @@ func (sv *Service) allocNodeID(ctx context.Context) (uint64, error) {
...
 
-// Get gets a value from the key-value store.
-// It is linearizable; that is, if a
-// Set happens before a Get,
-// the Get will observe the effects of the Set.
-// (There is still no guarantee an intervening
-// Set won't have changed the value again, but it is
+// RequestRead requests a linearizable read. Upon successful return,
+// the caller is guaranteed that a read from the Service's state will
+// be linearizable; this is, if a Set happens before a Get, the Get
```
kr: We can be a lot more concrete with this description now that this method has such a good interface.

"RequestRead requests a linearizable read. Upon successful return, subsequent reads will observe all writes that happened before the call to RequestRead. (There is still …)"

(Or possibly "WaitRead waits for…" if you go with that name.)

database/raft/raft.go
kr: It could also just be a lowercase version of the public method, so `requestRead`.

database/sinkdb/sinkdb.go
```diff
@@ -0,0 +1,132 @@
+// Package sinkdb provides a strongly consistent key-value store.
+package sinkdb
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/golang/protobuf/proto"
+
+	"chain/database/raft"
```
kr: Maybe in a followup, I think the right import path for package raft is now `chain/net/raft`.

database/sinkdb/sinkdb.go
tessr: 🚲  🚲  🚲  whatever you want to do is fine, ultimately 🚲  🚲  🚲  🚲  🚲  


🚲

database/sinkdb/sinkdb.go
```diff
@@ -0,0 +1,132 @@
...
+func IfNotExists(key string) (op Op) {
+	op.conds = append(op.conds, &sinkpb.Cond{
+		Type: sinkpb.Cond_NOT_KEY_EXISTS,
+		Key:  key,
+	})
+	return op
+}
+
+// Delete encodes a delete operation for key.
+func Delete(key string) (op Op) {
```
kr: Nit: these return signatures should prob just be `Op` (rather than `(op Op)` which is stutter and doesn't add any useful info).

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

Yeah, even if the body of the method is a little uglier (like one extra `var op Op` line or whatev) the better interface outweighs that imo.

database/sinkdb/sinkdb.go
```diff
@@ -0,0 +1,132 @@
...
+	return db, nil
+}
+
+// DB provides access to an opened kv store.
+type DB struct {
+	state *state
+	raft  *raft.Service
+}
+
+type Op struct {
```
kr: Maybe think about putting this type and its constructors in a separate file op.go. I have a feeling the list will grow in the future.

database/sinkdb/sinkdb.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Get performs a linearizable read of the provided key. The
+// read value is unmarshalled into v.
+func (db *DB) Get(ctx context.Context, key string, v proto.Message) (found bool, err error) {
+	err = db.raft.RequestRead(ctx)
+	if err != nil {
+		return false, err
+	}
+	buf := db.state.get(key)
+	if len(buf) == 0 {
```
kr: Are we sure this is a safe way to check for "found"? A zero-value protobuf can serialize to nil, which we'll happily insert into the state map (and which will correctly unmarshal back to the zero-value struct). In many cases it prob won't matter, but I think it would be good if the `found` value returned by this func exactly corresponds to the condition that's tested by `IfNotExists`.

We were already doing it this way before, so maybe we don't need to change now. But could we at least add a todo?

I think the "right way" would be for `get` to return a bool and have `Get` just look at the bool, not the value, before unmarshaling.

core/grants.go
```diff
@@ -159,24 +154,15 @@ func (a *API) deleteGrant(ctx context.Context, x apiGrant) error {
...
 		GuardData: guardData,
 		Protected: x.Protected, // should always be false
 	}
 
-	grantList := new(authz.GrantList)
-	err = proto.Unmarshal(data, grantList)
-	if err != nil {
+	var grantList authz.GrantList
+	found, err := a.sdb.Get(ctx, GrantPrefix+x.Policy, &grantList)
+	if err != nil || !found {
```
kr: When found is false, this will return `errors.Wrap(nil)`, which is `nil`, which is the right thing to do, but maybe deserves a comment?

core/grants.go
```diff
@@ -114,19 +113,15 @@ func (a *API) listGrants(ctx context.Context) (map[string]interface{}, error) {
...
 		// perhaps could denormalize the data in storage to speed this up,
 		// but for now assume a small number of grants
-		data, err := a.raftDB.Get(ctx, GrantPrefix+p)
+		var grantList authz.GrantList
+		ok, err := a.sdb.Get(ctx, GrantPrefix+p, &grantList)
 		if err != nil {
 			return nil, errors.Wrap(err)
 		}
-		if data == nil {
+		if !ok {
```
kr: We could also delete this whole if stmt right? (b/c grantList.Grants will be empty)

core/coreunsafe/reset.go
```diff
@@ -63,23 +63,23 @@ func ResetBlockchain(ctx context.Context, db pg.DB, rDB *raft.Service) error {
...
+	err := ResetBlockchain(ctx, db, sdb)
 	if err != nil {
 		return errors.Wrap(err)
 	}
 
-	// Delete all grants in raft storage
+	// Delete all grants in sinkdb
 	for _, p := range core.Policies {
-		err = rDB.Delete(ctx, core.GrantPrefix+p)
+		err = sdb.Exec(ctx, sinkdb.Delete(core.GrantPrefix+p))
```
kr: Todo: all these deletes in a big batch.

database/sinkdb/sinkdb.go
jbowens: yeah, agreed it's wonky. i'll TODO it for now

database/sinkdb/sinkdb.go
tessr: Awesome! Do you want to add a TODO to the migration function?

kr:
LGTM too.

PR #1193, opened 2017-05-15
vendor/github.com/lsegal/gucumber: update

tessr:
Updates `github.com/lsegal/gucumber` to `71608e2f6e76fd4da5b09a376aeec7a5c0b5edbc`. This is a dependency used by the aws packages.


kr:
I think it may no longer be a dependency used by the aws packages! Their docs claim they vendor all their dependencies, and there is a vendor directory inside aws-sdk-go, and it doesn't contain gucumber :shruggie:

tessr:
Ah, does this mean we can delete all the packages we've vendored just for AWS?

kr:
> Ah, does this mean we can delete all the packages we've vendored just for AWS?

Yeah, pretty sure.

tessr:
Awesome! I had made a list of them for _just this very purpose_

PR #1195, opened 2017-05-16
cmd/integtest: run integration tests (new command)

kr:
This new command provides support for integration test
packages. It sets up common resources neede by most or
all tests, such as running a postgres cluster and
providing a working directory.

### DIFF
--- a/cmd/integtest/main.go
+++ b/cmd/integtest/main.go
@@ -90,7 +90,7 @@ func main() {
 	err = cmd.Wait()
 	if err != nil {
 		log.Printf("%s: %v", base, err)
-		panic("cmd failed")
+		os.Exit(1)
 	}
 }
 
--- a/cmd/integtest/main.go
+++ b/cmd/integtest/main.go
@@ -28,10 +28,12 @@ var (
 	gobin    = dir + "/bin"
 	pgdir    = dir + "/pg"
 	pgrun    = dir + "/pgrun" // for socket file
+	pglog    = dir + "/pglog" // for log file
 )
 
 var (
 	flagT = flag.Duration("t", 15*time.Minute, "abort the test after the given duration")
+	flagL = flag.Duration("l", 0, "pg log_min_duration_statement (-1 to disable)")
 )
 
 func main() {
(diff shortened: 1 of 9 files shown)


kr:
PTAL

cmd/integtest/main.go
```diff
@@ -0,0 +1,197 @@
...
+		"sockdir": pgrun,
+	}))
+	must(ioutil.WriteFile(pgdir+"/postgresql.conf", buf.Bytes(), 0600))
+
+	buf.Reset()
+	must(hbaTemplate.Execute(&buf, nil))
+	must(ioutil.WriteFile(pgdir+"/pg_hba.conf", buf.Bytes(), 0600))
+
+	fmt.Println("postgres", "-D", pgdir)
+	cmd := command(ctx, "postgres", "-D", pgdir)
```
jbowens: Do we always print the command before executing it? maybe `command` should print it to avoid accidental drift between the printlns and the executed commands?

cmd/integtest/pg.go
```diff
@@ -0,0 +1,57 @@
...
+
+# replication
+# logging
+datestyle = 'iso, mdy'
+timezone = 'UTC'
+lc_messages = 'en_US.UTF-8'     # locale for system error message
+lc_monetary = 'en_US.UTF-8'     # locale for monetary formatting
+lc_numeric = 'en_US.UTF-8'      # locale for number formatting
+lc_time = 'en_US.UTF-8'       # locale for time formatting
+
```
jbowens: would there be any value collecting postgres query logs for debugging?
https://github.com/chain/chain/blob/main/cmd/benchcore/main.go#L773-L777

jbowens:
lgtm

cmd/integtest/pg.go
kr: Good call, I will put it in a todo. (I'm thinking the best way to do this is to tar up the whole directory and put that in in S3, so all the files are available including the pg log.)

kr: Oh actually it would be easy enough to at least emit the logs now, even if we don't upload the whole thing to S3. That would at least be useful for local debugging.

kr:
PTAL

jbowens:
lgtm

PR #1197, opened 2017-05-16
database/sinkdb/sinkdbtest: new package

jbowens:
Add package with test utilities for creating test sinkdb instances.
This will also prevent tests in separate packages from interacting.

### DIFF
--- a/database/sinkdb/sinkdbtest/sinkdbtest.go
+++ b/database/sinkdb/sinkdbtest/sinkdbtest.go
@@ -4,26 +4,43 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 
 	"chain/database/sinkdb"
 )
 
+const dataDirectoryPrefix = `chain-syncdbtest`
+
 // NewDB creates a new sinkdb instance with a random temporary
 // storage directory.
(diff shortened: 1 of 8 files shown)


kr:
It could do what the pgtest package does to GC old databases: name them with the datetime of creation and delete old ones on startup. Note that the finalizer in that package is only meant to close connections to avoid hitting the postgres connection limit. Since sinkdb has no connection limit we prob don't need that here.

jbowens:
Updated to gc old directories using the directory modified time. That should work for our purposes, right?

kr:
> directory modified time

Oh good idea. Yeah that seems reasonable. :eyes:

database/sinkdb/sinkdbtest/sinkdbtest.go
```diff
@@ -0,0 +1,46 @@
...
+
+func gcDataDirectories() {
+	tempDir := os.TempDir()
+	cutoff := time.Now().Add(-time.Hour * 24)
+	dirents, _ := ioutil.ReadDir(tempDir)
+	for _, dirent := range dirents {
+		if !strings.HasPrefix(dirent.Name(), dataDirectoryPrefix) {
+			continue
+		}
+		if dirent.ModTime().After(cutoff) {
```
kr: Before?

kr: Oops my bad.

PR #1198, opened 2017-05-16
net/raft: move from database/raft

jbowens:
Move chain/database/raft to chain/net/raft. The raft package now only
implements networking between raft nodes.


kr:
LGTM

PR #1199, opened 2017-05-16
database/sinkdb: fix key existence check

jbowens:
Propagate a flag indicating the existence of the key instead of
interpreting an empty byte slice as a nonexistent key.

### DIFF
--- a/database/sinkdb/sinkdb.go
+++ b/database/sinkdb/sinkdb.go
@@ -54,20 +54,14 @@ func (db *DB) Get(ctx context.Context, key string, v proto.Message) (found bool,
 		return false, err
 	}
 	buf, found := db.state.get(key)
-	if !found {
-		return false, err
-	}
-	return true, proto.Unmarshal(buf, v)
+	return found, proto.Unmarshal(buf, v)
 }
 
 // GetStale performs a non-linearizable read of the provided key.
 // The value may be stale. The read value is unmarshalled into v.
 func (db *DB) GetStale(key string, v proto.Message) (found bool, err error) {
 	buf, found := db.state.get(key) // read directly from state
-	if !found {
-		return false, err
-	}
-	return true, proto.Unmarshal(buf, v)
+	return found, proto.Unmarshal(buf, v)
 }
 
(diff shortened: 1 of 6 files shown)


database/sinkdb/sinkdb.go
```diff
@@ -53,9 +53,8 @@ func (db *DB) Get(ctx context.Context, key string, v proto.Message) (found bool,
 	if err != nil {
 		return false, err
 	}
-	buf := db.state.get(key)
-	// TODO(jackson): propagate real key existence bool
-	if len(buf) == 0 {
+	buf, found := db.state.get(key)
+	if !found {
```
kr: I think it would be better to call Unmarshal unconditionally:

```go
buf, found := db.state.Get(key)
return found, proto.Unmarshal(buf, v)
```

The reason is that sometimes the caller doesn't care if the key doesn't exist — they might want to ignore the found flag and see a zero value as if that had been explicitly stored. Separately, sometimes the caller might want to reuse a protobuf struct to avoid allocating. If both of these happen at the same time, the right thing to do is to `v.Reset()` so the caller gets a zero value out. Unmarshal does this; it calls Reset before unmarshaling, even if the byte slice is empty.

database/sinkdb/sinkdb.go
```diff
@@ -64,9 +63,8 @@ func (db *DB) Get(ctx context.Context, key string, v proto.Message) (found bool,
 // GetStale performs a non-linearizable read of the provided key.
 // The value may be stale. The read value is unmarshalled into v.
 func (db *DB) GetStale(key string, v proto.Message) (found bool, err error) {
-	buf := db.state.get(key) // read directly from state
-	// TODO(jackson): propagate real key existence bool
-	if len(buf) == 0 {
+	buf, found := db.state.get(key) // read directly from state
+	if !found {
```
kr: Same thing here as above.

database/sinkdb/sinkdb.go
jbowens: Ah, very good point. 👍

kr:
LGTM

PR #1200, opened 2017-05-16
core/coreunsafe: reset sinkdb atomically

jbowens:
Refactor ResetBlockchain and ResetEverything to delete sinkdb's
data atomically.


kr:
LGTM

PR #1201, opened 2017-05-16
database/sinkdb: turn AddAllowedMember into an Op constructor

kr:
This lets it be combined with other Ops in a single
atomic batch.


kr:
PTAL

jbowens:
lgtm

PR #1204, opened 2017-05-17
database/sinkdb: expose ErrConflict

jbowens:
Remove raft.ErrUnsatisifed, replacing it with sinkdb.ErrConflict. By
default map sinkdb.ErrConflict to a 409 status code. When we introduce
code paths where sinkdb.ErrConflict is result of a temporary conflict in a
read-modify-write cycle, we can add another error mapped to a temporary
error.

Also, remove the error return value from raft.State.Apply. The raft
package would panic if it was non-nil. Now, the sinkdb state will panic
in those error conditions.

### DIFF
--- a/core/errors.go
+++ b/core/errors.go
@@ -16,6 +16,7 @@ import (
 	"chain/core/txbuilder"
 	"chain/core/txfeed"
 	"chain/database/pg"
+	"chain/database/sinkdb"
 	"chain/errors"
 	"chain/net/http/authz"
 	"chain/net/http/httperror"
@@ -28,6 +29,8 @@ func isTemporary(info httperror.Info, err error) bool {
 	case "CH000": // internal server error
 		return true
 	case "CH001": // request timed out
+		return true
+	case "CH012": // conflict
 		return true
 	case "CH761": // outputs currently reserved
 		return true
@@ -63,6 +66,7 @@ var errorFormatter = httperror.Formatter{
 		errNotAuthenticated:        {401, "CH009", "Request could not be authenticated"},
 		txbuilder.ErrMissingFields: {400, "CH010", "One or more fields are missing"},
 		authz.ErrNotAuthorized:     {403, "CH011", "Request is unauthorized"},
+		sinkdb.ErrConflict:         {409, "CH012", "Database transaction conflict; retry"},
 		asset.ErrDuplicateAlias:    {400, "CH050", "Alias already exists"},
 		account.ErrDuplicateAlias:  {400, "CH050", "Alias already exists"},
 		txfeed.ErrDuplicateAlias:   {400, "CH050", "Alias already exists"},
--- a/core/errors.go
+++ b/core/errors.go
@@ -30,8 +30,6 @@ func isTemporary(info httperror.Info, err error) bool {
 		return true
 	case "CH001": // request timed out
 		return true
-	case "CH012": // conflict
-		return true
 	case "CH761": // outputs currently reserved
 		return true
 	case "CH706": // 1 or more action errors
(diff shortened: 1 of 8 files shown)


core/errors.go
```diff
@@ -29,6 +30,8 @@ func isTemporary(info httperror.Info, err error) bool {
 		return true
 	case "CH001": // request timed out
 		return true
+	case "CH012": // conflict
+		return true
```
kr: How would this be used by clients? Generally I would expect that issuing the same request again would result in another conflict. If the client wants to retry their semantic operation, they'd have to do another read-modify-write, they can't just retry the same write. So I would think it should generally be permanent.

Or is the idea here that the entire read-modify-write cycle is happening inside of the HTTP handler? Then yeah the client's error would be temporary? But still I feel like in that situation, we should handle it in the server if possible. The default should be if ErrConflict is unhandled it should bubble up as a permanent error to the client (e.g. for /configure I think that would be the right thing if we weren't already explicitly checking for "already configured").

net/raft/raft.go
```diff
@@ -381,55 +379,56 @@ func runTicks(rn raft.Node) {
 
 // Exec proposes the provided instruction and waits for it to be
 // satisfied.
-func (sv *Service) Exec(ctx context.Context, instruction []byte) error {
+func (sv *Service) Exec(ctx context.Context, instruction []byte) (satisfied bool, err error) {
 	prop := proposal{Wctx: randID(), Instruction: instruction}
 	data, err := json.Marshal(prop)
 	if err != nil {
-		return errors.Wrap(err)
+		return satisfied, errors.Wrap(err)
```
kr: Nit, these could all say `false` instead. I think that'd be easier to interpret since they will always actually be false.

core/errors.go
jbowens: Yeah, the idea was latter. We could try to handle it in the server, but to-date we've generally pushed that retry logic into the clients to avoid multiplying our retries by retrying at different points in the request cycle. This error seems akin to a Postgres serialization conflict error. Does lib/pq automatically retry that for us?

jbowens: Maybe we need two errors, the default of which is a permanent error. If the handler is doing a read-modify-write cycle, it can re-map the permanent error into a temporary error so that the client will retry.

kr: Makes sense.

Yeah eventually we'll prob need to distinguish between "your operation conflicted" and "an internal operation conflicted and you should retry".

Currently we're only sending the latter. In fact, should this case be a 5xx error?

jbowens: Currently the only conditionals we use are IfNotExists when storing a config. That seems like a "Your operation conflicted" error, so I updated this to make this error permanent. When we add an "an internal operation conflicted and you should retry" error case, we can add the temporary error.

This has me thinking though, the IfNotExists conditional makes /configure not idempotent. Maybe we should be using conditional that allows the operation if the key doesn't exist or its value is equal to x. Can we compare two protobufs using proto.Equal without knowing the schema? If not, gogo/protobuf has a [stable_marshaler option](https://github.com/gogo/protobuf/pull/156).

jbowens: I think we need something like the c++ library's [MessageDifferencer](https://github.com/google/protobuf/blob/master/src/google/protobuf/util/message_differencer.h)

jbowens: I guess that wouldn't be sufficient for /configure, because the resulting Config proto isn't deterministic based on the request. The /configure handler will populate with a random core ID, the current time, etc.

kr: Yeah huh. Okay then.

kr:
LGTM

PR #1207, opened 2017-05-17
remove outdated table of contents in blockchain programs paper

danrobinson:


jeffomatic:
@danrobinson Is this update on the `ivy` branch? If so, I will just replay my diff-patch process with the tip of `ivy` and force-push over my other PR. That will be cleaner than doing this as a second-order PR into another PR.

danrobinson:
It is not... I'd be happy to cherry-pick it onto there but I'm worried about messing up their build process. Could I just do it as a PR into the main 1.2 docs branch?

This can also just wait until all the builds are done.

jeffomatic:
Okay, if it's not urgent, why don't we hang tight and sort it out when there's a bit less churn.

PR #1208, opened 2017-05-17
database/sinkdb: protect state with a mutex

jbowens:
Before the sinkdb refactor, all accesses to the state were protected
by raft.Service's stateMu. After the refactor, sinkdb directly accesses
the state without acquiring raft.Service's mutex, allowing data races.

This adds a mutex to protect the state. The raft package's stateMu is
still necessary for protecting raft-specific fields and for the
condition variable. There are a few places where we can likely avoid
acquiring stateMu's mutex and instead rely on State's internal locking.

### DIFF
--- a/database/sinkdb/state.go
+++ b/database/sinkdb/state.go
@@ -2,6 +2,7 @@ package sinkdb
 
 import (
 	"bytes"
+	"sync"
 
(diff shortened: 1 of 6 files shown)


kr:
:eyes:

kr:
LGTM

PR #1209, opened 2017-05-17
docs: rename /data to /blockchain in the table of contents

oleganza:


kr:
Is there a reason for this change? Personally I would rather not make these sorts of changes unless the old name is very bad. (IMO "data" or "data model" is fine.)

oleganza:
The file was already changed from `data.md` to `blockchain.md` some time ago. This only fixes 404 issue. The title is "Blockchain Specification" as of now. 

@dan WDYT?

danrobinson:
LGTM

PR #1210, opened 2017-05-18
net/raft: add explicit cluster initialization

jbowens:
Move cluster initialization out into two explicit methods: Init to
initialize a new cluster and Join to join an existing cluster.

For now this only changes the interface of the net/raft package. In a
followup we can introduce init and join endpoints.


jbowens:
PTAL

jbowens:
unexported `Initialized` and moved the init/join calls to main

kr:
LGTM

PR #1211, opened 2017-05-18
core/config: fix mockhsm log statement

jbowens:
Log the block pub key in hex.

When configuring with the blockhsm, previously cored would log the
raw bytes of the block-signing key and add an unnecessary "\n"
to the end:

```
app=cored buildtag=? processID=chain-mba.local-26010 reqid=5ab5bcc6660c33989b9f at=mockhsm.go:21 t=2017-05-18T17:15:14.019931843Z message="Generated new block-signing key J\xd4\x00\x0f\xbcR*\xf84(\x1c\xa0q\xb1\xe3\x16\xd1;\u0382R\xb2\xdbuS\x15\x0f4\xc3\x00\x852\n"
```

### DIFF
--- a/core/config/mockhsm.go
+++ b/core/config/mockhsm.go
@@ -18,9 +18,9 @@ func getOrCreateDevKey(ctx context.Context, db pg.DB, c *Config) (blockPub ed255
 		return nil, err
 	}
 	if created {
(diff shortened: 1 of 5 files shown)


kr:
LGTM

PR #1215, opened 2017-05-18
core: respond with Bad Request on bad address

jbowens:
In calls to the /add-allowed-member RPC endpoint, return a 400 Bad
Request error response when the provided address is invalid.

```
jackson@mba ~ $ corectl allow-address http://:1999
RPC error: CH161 Address is invalid
Detail: too many colons in address
```


tessr:
lgtm

PR #1217, opened 2017-05-18
cmd/chainvet: rename from cmd/vet

### LINKED ISSUE #1173: rename chain/cmd/vet to chainvet
The name `vet` is confusing with the Go tool `vet` (run as `go vet`).

jbowens:
Fix #1173.


cmd/chainvet/doc.go
```diff
@@ -5,26 +5,26 @@
...
+chainvet examines Go source code and reports suspicious constructs, such as
+Printf calls whose arguments do not align with the format string. chainvet
+uses heuristics that do not guarantee all reports are genuine problems,
+but it can find errors not caught by the compilers.
 
 It can be invoked three ways:
 
 By package, from the go tool:
-	go vet package/path/name
+	chainvet package/path/name
```
kr: Ooo I bet this way doesn't work at all, because `vet` relies on the go tool to find all the package files for it. Our chainvet is a fork of `vet` so I think it only supports the second two forms.

And there's no way to teach `go` how to invoke chainvet in this way. Even if we dropped our binary in the tools dir, that would let us run it as `go tool chainvet` but that's no better than just running `chainvet` directly.

PR #1218, opened 2017-05-19
net/http/authz: refactor grant storage

kr:
This change primarily introduces a new type, *authz.Store, to hold the logic for reading and writing Grant objects to persistent storage (specifically, to sinkdb). It also defines a new interface, Loader, to load a set of grants for a given set of policies. Type *Store satisfies Loader.

NewAuthorizer now accepts a Loader rather than a concrete database type. This lets callers supply a *Store object directly, interpose a wrapper to augment the results (this is what package core does), or provide a completely separate implementation (such as, hypothetically, a static in-memory-only grant set for unit tests).

We also take this opportunity to remove the special cases and build tag from package authz and put them where they belong:
- logic for adding the required `internal` X.509 grant is now in chain/core
- logic for adding localhost grants is now in chain/cmd/cored

Future work: also move the delete logic into package authz (and possibly un-export EqualGrant).


kr:
PTAL

tessr:
👀

tessr:
(LGTM)

PR #1220, opened 2017-05-19
core: do not forward request credentials

jbowens:
Do not forward incoming request credentials. Instead, rely on the
internal credentials for authenticating internal RPCs.

Also, always use https when forwarding requests. High-availability isn't
available in Developer Edition, and we expect all internal RPCs to be
authenticated via TLS.


jbowens:
What's the recommended way to setup tls certs to test this locally?

kr:
I've been using TLS on my laptop for a while now. It is mostly pretty seamless except for the big ugly browser warning for the self-signed cored cert.

Maybe we can get one or two of us (who weren't involved in developing our TLS support) to give it a test drive and see how it feels.

PR #1223, opened 2017-05-22
protocol: add missing check for txheader min/maxtime vs. block timestamp

bobg:
Also try to detect an expired transaction during `/submit-transaction`.

### DIFF
--- a/core/txbuilder/finalize.go
+++ b/core/txbuilder/finalize.go
@@ -45,6 +45,9 @@ func FinalizeTx(ctx context.Context, c *protocol.Chain, s Submitter, tx *legacy.
 	if err != nil {
 		return errors.Wrap(err, "tx rejected")
 	}
+	if tx.Tx.MaxTimeMs < c.TimestampMS() {
+		return errors.Wrap(ErrRejected, "tx expired")
+	}
 
 	err = s.Submit(ctx, tx)
 	return errors.Wrap(err)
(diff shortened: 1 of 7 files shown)


bobg:
PTAL

core/txbuilder/finalize.go
```diff
@@ -45,6 +45,9 @@ func FinalizeTx(ctx context.Context, c *protocol.Chain, s Submitter, tx *legacy.
 	if err != nil {
 		return errors.Wrap(err, "tx rejected")
 	}
+	if tx.Tx.MaxTimeMs < c.TimestampMS() {
+		return errors.Wrap(ErrRejected, "tx expired")
+	}
```
jbowens: I think this check already happens up in the submit handler: https://github.com/chain/chain/blob/main/core/transact.go#L273-L275

kr:
👀

protocol/block.go
```diff
@@ -88,6 +88,17 @@ func (c *Chain) GenerateBlock(ctx context.Context, prev *legacy.Block, snapshot
...
 		}
 
+		// Filter out transactions that are not yet valid, or no longer
+		// valid, per the block's timestamp.
+		if tx.Tx.MinTimeMs > b.TimestampMS {
+			// TODO(bobg): log this?
+			continue
+		}
+		if tx.Tx.MaxTimeMs < b.TimestampMS {
+			// TODO(bobg): log this?
```
kr: My vote would be nah, don't log it, since we surface this to the client more directly. (Either by the "expired" message in this PR or by a timeout on /submit if it's too far in the future.)

core/txbuilder/finalize.go
kr: oops my bad, forgot we already do that when i suggested this

PR #1225, opened 2017-05-22
docs: clarify backward compatibility rules for SDKs

jeffomatic:
This commit refines our versioning language such that
backward compatibility guarantees only hold between server packages and
client packages of the previous minor version. In particular, this
is intended to eliminate any interpretation that holds that new SDKs
should be backward-compatible with application code written for a
previous SDK version.

The goal of this change is to allow us to pursue a more aggressive
release roadmap, and to eliminate versioning kludges in the SDK. It
allows us to introduce new usage patterns (potentially breaking changes)
in consecutive minor versions of the SDK without eliminating a smooth
upgrade path for application code.

### DIFF
--- a/docs/core/reference/versioning.md
+++ b/docs/core/reference/versioning.md
@@ -60,7 +60,17 @@ Software packages are mutually compatible if they share the same major and minor
 
 #### Deprecations and breaking changes
 
-Deprecations and breaking changes can occur when the minor version changes. If a feature is deprecated, it will remain functional for at least one minor version before being removed.
+Deprecations and breaking changes can occur when the minor version changes. To the greatest extent possible, breaking changes in server packages will be preceded by a deprecation announcement and at least one minor version cycle of continued support.
+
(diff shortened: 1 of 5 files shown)


docs/core/reference/versioning.md
```diff
@@ -60,7 +60,17 @@ Software packages are mutually compatible if they share the same major and minor
 
 #### Deprecations and breaking changes
 
-Deprecations and breaking changes can occur when the minor version changes. If a feature is deprecated, it will remain functional for at least one minor version before being removed.
+Deprecations and breaking changes can occur when the minor version changes. To the greatest extent possible, breaking changes in server packages will be preceded by a deprecation announcement and at least one minor version cycle of continued support.
```
kr: I would remove "to the greatest extent possible". We should keep a strong promise for client-server compatibility, not hedge our language to give us wiggle room in the future.

jeffomatic: SGTM

PR #1226, opened 2017-05-22
protocol: add missing check for txheader min/maxtime vs. block timestamp

bobg:
Backported from https://github.com/chain/chain/pull/1223 and https://github.com/chain/chain/pull/1224

### DIFF
--- a/core/txbuilder/finalize.go
+++ b/core/txbuilder/finalize.go
@@ -45,6 +45,9 @@ func FinalizeTx(ctx context.Context, c *protocol.Chain, s Submitter, tx *legacy.
 	if err != nil {
 		return errors.Wrap(err, "tx rejected")
 	}
+	if tx.Tx.MaxTimeMs < c.TimestampMS() {
+		return errors.Wrap(ErrRejected, "tx expired")
+	}
 
 	err = s.Submit(ctx, tx)
 	return errors.Wrap(err)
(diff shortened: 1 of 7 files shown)


core/txbuilder/finalize.go
```diff
@@ -45,6 +45,9 @@ func FinalizeTx(ctx context.Context, c *protocol.Chain, s Submitter, tx *legacy.
 	if err != nil {
 		return errors.Wrap(err, "tx rejected")
 	}
+	if tx.Tx.MaxTimeMs > 0 && tx.Tx.MaxTimeMs < c.TimestampMS() {
```
kr: Is this part absolutely necessary? I'd prefer to keep change to point releases as small and conservative as possible.

protocol/block.go
```diff
@@ -88,6 +88,17 @@ func (c *Chain) GenerateBlock(ctx context.Context, prev *legacy.Block, snapshot
 			continue
 		}
 
+		// Filter out transactions that are not yet valid, or no longer
+		// valid, per the block's timestamp.
+		if tx.Tx.MinTimeMs > 0 && tx.Tx.MinTimeMs > b.TimestampMS {
+			// TODO(bobg): log this?
```
kr: I don't think these are necessary. Same comment as in the PR on main.

core/txbuilder/finalize.go
bobg: Ha, was just having this discussion with @jeffomatic in Slack. Your opinion tipped the balance; removed.

kr:
LGTM

PR #1229, opened 2017-05-22
cmd/cored: do not automatically setup raft cluster

jbowens:
Remove the BOOTURL environment variable from cored. If there is no
existing raft cluster configuration, start cored but return an error
from all RPCs indicating that the process's cluster needs to be
initialized.

The operator can call the /init-cluster or /join-cluster endpoints to
start a new Chain Core cluster or join an existing cluster.

Introduce a new 'init_cluster' build tag to automatically initialize
a new cluster if one doesn't exist in cmd/cored. Use this new build tag
when building Developer Edition.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -24,6 +24,7 @@ import (
 	"chain/errors"
 	"chain/log"
 	"chain/net/http/authz"
+	"chain/net/raft"
 	"chain/protocol"
 	"chain/protocol/bc"
 	"chain/protocol/state"
@@ -51,6 +52,7 @@ var (
 		MockHSM       bool `json:"is_mockhsm"`
 		Reset         bool `json:"is_reset"`
 		HTTPOk        bool `json:"is_http_ok"`
+		ClusterInit   bool `json:"is_cluster_init"`
 	}
 )
 
@@ -61,10 +63,11 @@ var (
 func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	c := new(Config)
 	found, err := sdb.Get(ctx, "/core/config", c)
-	if err != nil {
+	if errors.Root(err) == raft.ErrUninitialized {
+		return nil, nil
+	} else if err != nil {
 		return nil, errors.Wrap(err)
-	}
-	if found {
+	} else if found {
 		return c, nil
 	}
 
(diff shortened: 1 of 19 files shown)


jbowens:
PTAL

A few questions:
* What should the init and join rpc endpoints be named?
* What should the auto-init build tag be named?
* Is it okay that for a non-Developer Edition Chain Core, dashboard will load, but you won't be able to configure until you `corectl init`?:

screen shot 2017-05-22 at 3 50 11 pm

kr:
> * What should the init and join rpc endpoints be named?
> * What should the auto-init build tag be named?

The names you chose seem fine to me.

> * Should we preserve BOOTURL for now as a deprecated environment variable?

I think we should not. That env var only matters while setting up a new cluster. Existing deployments will not be affected.

docker/ci/bin/setup-core
```diff
@@ -23,5 +23,8 @@ PATH=$(go env GOPATH)/bin:$PATH:$CHAIN/bin
 go install -tags 'http_ok localhost_auth' chain/cmd/{cored,corectl}
 initlog=`mktemp`
 cored 2>&1 | tee $initlog &
+
+corectl wait
+corectl init
```
kr: Should we use the `init_cluster` build tag here instead?

core/core.go
```diff
@@ -152,6 +154,37 @@ func (a *API) configure(ctx context.Context, x *config.Config) error {
...
+	// add this process's address as an allowed member
+	err = a.addAllowedMember(ctx, struct{ Addr string }{a.addr})
+	return err
+}
+
+func (a *API) joinCluster(ctx context.Context, x struct {
+	BootAddress string `json:"boot_address"`
+}) error {
+	// validate the format of the boot address
+	_, _, err := net.SplitHostPort(x.BootAddress)
```
kr: We could actually try making a connection here. :stare:
Just a thought.

core/config/config.go
```diff
@@ -51,6 +52,7 @@ var (
 		MockHSM       bool `json:"is_mockhsm"`
 		Reset         bool `json:"is_reset"`
 		HTTPOk        bool `json:"is_http_ok"`
+		ClusterInit   bool `json:"is_cluster_init"`
```
kr: `init_cluster` or `cluster_init`? (It's different in build-cored-release.)

core/errors.go
```diff
@@ -92,6 +92,8 @@ var errorFormatter = httperror.Formatter{
 		errInvalidAddr:                 {400, "CH161", "Address is invalid"},
 		raft.ErrAddressNotAllowed:      {400, "CH162", "Address is not allowed"},
 		raft.ErrUninitialized:          {400, "CH163", "No cluster configured"},
```
kr: I should've caught this before, but I think we should not use the word "configure" to describe this state, since we already use it to mean something else that's pretty specific. Can we change this to say "Cluster not initialized" or "Uninitialized cluster" or something like that?

jbowens:
> Where is the build constraint?

Oops, good catch

core/config/config.go
```diff
@@ -61,10 +63,11 @@ var (
 func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	c := new(Config)
 	found, err := sdb.Get(ctx, "/core/config", c)
-	if err != nil {
+	if errors.Root(err) == raft.ErrUninitialized {
```
kr: I think maybe it might be better to handle this in cored? Like

```go
conf, err := config.Load(ctx, db, sdb)
if err != nil && errors.Root(err) != raft.ErrUninitialized {
}
```

and this function would stay as it was? (or maybe just have ErrUninitialized appear in its godoc)

kr:
LGTM (there's one minor thought there but it shouldn't block this)

PR #1231, opened 2017-05-22
net/raft: only use HTTPS

jbowens:
Developer Edition does not support multi-process Chain Cores. In EE,
processes are required to use TLS for authentication internally, so we
can remove the useTLS parameter to raft.Start.

Related to #1220.


kr:
\o/

PR #1232, opened 2017-05-23
core: propagate raft errors

### LINKED ISSUE #1219: core: expose raft errors on Core's health object
The net/raft.Service type has an Err method to expose serious errors that prevent raft from making progress. Currently, there's no way for the operator to access this value. Seems like we should propagate it onto the health object returned by /info.

jbowens:
Propagate raft errors into the /info health object.

Fix #1219.

### DIFF
--- a/core/health.go
+++ b/core/health.go
@@ -24,6 +24,11 @@ func (a *API) health() (x struct {
 	Errors map[string]string `json:"errors"`
 }) {
 	x.Errors = make(map[string]string)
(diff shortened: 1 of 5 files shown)


kr:
What do you think about adding some lines in core/health.go instead:

```go
if err := a.sdb.RaftService().Err(); err != nil {
	x.Errors["raft"] = err.Error()
}
```

PR #1233, opened 2017-05-23
core: remove useTLS flag on API struct

jbowens:
The useTLS flag is unnecessary now that Chain Core always uses TLS for
internal RPCs.


kr:
LGTM

PR #1236, opened 2017-05-24
cmd/integtest: check out repos into working dir

kr:
Don't inherit the Unix user's GOPATH and CHAIN
variables. Instead, maintain a dedicated Go workspace
with a checkout of the repos in $HOME/integration.


kr:
PTAL

PR #1237, opened 2017-05-24
net/http/authz: remove unnecessary return param

kr:
The caller doesn't need a copy of the existing grant;
it already has all that data at hand.

### DIFF
--- a/core/access_tokens.go
+++ b/core/access_tokens.go
@@ -52,7 +52,7 @@ func (a *API) createAccessToken(ctx context.Context, x struct{ ID, Type string }
 		// We've already returned if x.Type wasn't specified, so this must be a bad type.
 		return nil, accesstoken.ErrBadType
 	}
-	_, err = a.grants.Save(ctx, grant)
+	err = a.grants.Save(ctx, grant)
 	if err != nil {
 		return nil, errors.Wrap(err)
 	}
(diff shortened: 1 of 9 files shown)


kr:
PTAL

jeffomatic:
If the grant already exists, the `CreatedAt` timestamp of the original creation will be returned by `Store.Save`. AFAIK the dashboard uses this field for presentation purposes. I haven't thought through the full consequences of this change, but it seems like this does affect the semantics at least a little bit.

kr:
Dangit, true.

kr:
One thing we might want to get in the habit of in the future is returning just a "success" status for operations like this. If the client needs a copy of the object, it can look it up.

kr:
Weird, `grep -r create-authorization-grant dashboard` returns nothing. Can you point me to where dashboard does this operation?

kr:
Ah nm, I see it as `authorizationGrants.create`.

jeffomatic:
What's funny is I have the distinct impression my [original design](https://gist.github.com/jeffomatic/45b95fb1f4f725a4373fb4ec51f12d5b) just returned an OK message; I'm not sure how/when the full object response got introduced. More generally, I might be on the side where it's case-by-case whether the object should be returned; e.g. when creating accounts and assets, it's useful to get back the thing that got created, since there are some auto-generated fields like `id` that are useful.

For what it's worth, it seems that the dashboard is doing anything useful with `CreatedAt` field as it appears in the `create-authorization-grant` response. The dashboard obeys the pattern you described and uses the field as returned by the `list-authorization-grants` API call. That being said, it would be nice for the API as it exists to stay consistent, at least until we decide to change it formally.

kr:
Yeah agreed. Luckily I think it will be pretty easy in this case to keep it compatible.

kr:
PTAL

PR #1238, opened 2017-05-25
database/sinkdb: document Op

kr:


kr:
PTAL

database/sinkdb/op.go
```diff
@@ -6,6 +6,18 @@ import (
...
 )
 
+// Op represents a change to the data store.
+// Each Op starts with conditions, boolean predicates over
+// existing stored data.
+// If all conditions return true, the Op is said to be satisfied.
+// It then results in zero or more effects,
+// mutations to apply to the data.
+// If there is more than one way to satisfy the conditions,
+// the effects may differ depending on how they are satisfied.
```
jbowens: This sentence confused me:
> If there is more than one way to satisfy the conditions, the effects may differ depending on how they are satisfied.

Should it be "If there is more than one way to apply the effects, the resulting state may differ depending on how they are applied."?

kr: Haha, this sentence was a little too aspirational. I'll just delete it for now.

kr:
PTAL

jbowens:
lgtm

Issue #1239, opened 2017-05-25
\u0000 in reference data causes database error


danrobinson:
Submitting a transaction with reference data that contains the sequence `\u0000` causes `cored` to enter an infinite loop printing the following error:

```app=cored buildtag=? processID=chain-DANIELs-MacBook-Pro.local-54096-2c346f0425176f71e225 at=pin.go:216 t=2017-05-24T23:47:40.341422183Z error="pin \"tx\" callback: inserting annotated_txs to db: pq: unsupported Unicode escape sequence"
/Users/danielrobinson/Projects/go/src/chain/core/query/index.go:137 - chain/core/query.(*Indexer).insertAnnotatedTxs
/Users/danielrobinson/Projects/go/src/chain/core/query/index.go:69 - chain/core/query.(*Indexer).IndexTransactions
/Users/danielrobinson/Projects/go/src/chain/core/query/index.go:55 - chain/core/query.(*Indexer).IndexTransactions-fm
/Users/danielrobinson/Projects/go/src/chain/core/pin/pin.go:214 - chain/core/pin.(*pin).processBlock
/usr/local/Cellar/go/1.8/libexec/src/runtime/asm_amd64.s:2197 - runtime.goexit
```

jbowens:
Fixed in #1240.

PR #1240, opened 2017-05-25
core/query: validate reference data

jbowens:
Ignore invalid json in reference data when indexing transactions, and skip
reference data that contains the null byte escape sequence. This
fixes two separate but related issues:

  1. A user could use Chain Core to create a transaction with the
  `\u0000` escape sequence inside a string within the reference data.
  Chain Core would allow the transaction, because `\u0000` is a valid
  escape sequence in a JSON string, but indexing the transaction
  would fail (#1239) due to Postgres jsonb limitations:
  https://www.postgresql.org/message-id/E1YHHV8-00032A-Em@gemulon.postgresql.org

  2. A non-Chain Core client could create transactions with any sort of
  invalid json, and Chain Core would fail to index them. These
  transactions would be impossible to create with Chain Core, but
  another but other Chain Protocol clients or a malicious user could
  create them.

### DIFF
--- a/core/asset/block.go
+++ b/core/asset/block.go
@@ -31,7 +31,10 @@ type Saver interface {
 func Annotated(a *Asset) (*query.AnnotatedAsset, error) {
 	jsonTags := json.RawMessage(`{}`)
 	jsonDefinition := json.RawMessage(`{}`)
-	if len(a.RawDefinition()) > 0 {
+
+	// a.RawDefinition is the asset definition as it appears on the
+	// blockchain, so it's untrusted and may not be valid json.
+	if pg.IsValidJSONB(a.RawDefinition()) {
 		jsonDefinition = json.RawMessage(a.RawDefinition())
 	}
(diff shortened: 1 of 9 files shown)


jbowens:
!PTAL, still investigating #1239

jbowens:
PTAL

kr:
LGTM, maybe the commit message needs to be tweaked after the latest changes?

jbowens:
I just realized the json.Unmarshal won't error on invalid UTF-8 within a quoted JSON string:
> When unmarshaling quoted strings, invalid UTF-8 or invalid UTF-16 surrogate pairs are not treated as an error. Instead, they are replaced by the Unicode replacement character U+FFFD.

Postgres _will_ error if we try to insert JSON with invalid UTF-8 characters, so I added an additional check to ensure that the reference data is valid UTF-8.

kr:
LGTM again :)

PR #1241, opened 2017-05-25
net/http/authz: return sinkdb Op instead of executing

kr:
This will allow the caller to combine the returned Op
with other Ops into an atomic batch.

### DIFF
--- a/database/sinkdb/op.go
+++ b/database/sinkdb/op.go
@@ -69,3 +69,11 @@ func AddAllowedMember(addr string) Op {
 		}},
 	}
 }
+
+// Error returns an Op representing an error condition.
+// Exec will return err, and have no effect,
+// when the returned Op is executed.
+// If err is nil, Error returns the zero Op.
+func Error(err error) Op {
+	return Op{err: err}
+}
--- a/net/http/authz/store.go
+++ b/net/http/authz/store.go
@@ -91,7 +91,7 @@ func (s *Store) Delete(ctx context.Context, policy string, delete func(*Grant) b
 		}
 	}
 
-	// We didn't match any grants, don't need to do an update. Return success
+	// We didn't match any grants, don't need to do an update. Return no-op.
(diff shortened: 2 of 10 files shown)


kr:
PTAL

net/http/authz/store.go
```diff
@@ -94,16 +93,12 @@ func (s *Store) Delete(ctx context.Context, policy string, delete func(*Grant) b
 
 	// We didn't match any grants, don't need to do an update. Return success
```
tessr: Probably should update this comment to "return no-op"

database/sinkdb/op.go
```diff
@@ -69,3 +69,11 @@ func AddAllowedMember(addr string) Op {
 		}},
 	}
 }
+
+// Error returns an Op representing an error condition.
+// Exec will return err, and have no effect,
+// when the returned Op is executed.
+// If err is nil, Error returns the zero Op.
```
tessr: I am a leeeettle bit worried that this is non-obvious but maybe it's part of a pattern I'm not familiar with.

tessr: Oh, I guess it's similar to Errors.Wrap returning nil if err is nil

net/http/authz/store.go
```diff
@@ -94,16 +93,12 @@ func (s *Store) Delete(ctx context.Context, policy string, delete func(*Grant) b
...
 	}
 
-	gList := &GrantList{Grants: keep}
-	err = s.sdb.Exec(ctx, sinkdb.Set(key, gList))
-	if err != nil {
-		return errors.Wrap(err)
-	}
-
-	return nil
+	// TODO(tessr): Make this safe for concurrent updates. Will likely require a
```
tessr: does this todo belong here?

kr: Yes. What if the read-modify-write of Delete interleaves with the read-modify-write of another Delete or a Save?

PR #1242, opened 2017-05-25
net/raft: campaign immediately in Init

jbowens:
Immediately campaign for leader when initializing a new cluster. This
should speed up tests. We'll no longer need to wait electionTick * tickDur
before having a usable raft node.

If we convert to raft.RawNode at some point, we could avoid the
less-than-ideal synchronous Ready channel wait. Because StartNode
documents that it appends ConfChangeAddNode entries for each peer
to the initial log, I think we are guaranteed that the channel wait will
not block indefinitely. We might not be guaranteed that it'll actually
contain all the initial config changes, but in that case Campaign will
fail silently and the node will campaign after electionTick ticks.

See coreos/etcd#7985.

### DIFF
--- a/net/raft/raft.go
+++ b/net/raft/raft.go
@@ -240,17 +240,16 @@ func Start(laddr, dir string, httpClient *http.Client, state State) (*Service, e
 
 	// Start the algorithm. It is okay to not lock startMu since
 	// sv hasn't escaped yet.
-	sv.startLocked(id, raftNode, walobj)
+	sv.id = id
+	sv.raftNode = raftNode
+	sv.startLocked(walobj)
 
 	return sv, nil
 }
 
 // startLocked begins the raft algorithm. It requires sv.startMu
 // to already be locked.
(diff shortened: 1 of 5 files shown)


kr:
:eyes:

net/raft/raft.go
```diff
@@ -304,7 +304,23 @@ func (sv *Service) Init() error {
 
 	peers := []raft.Peer{{ID: firstNodeID, Context: []byte(sv.laddr)}}
 	raftNode := raft.StartNode(sv.config(firstNodeID), peers)
-	sv.startLocked(firstNodeID, raftNode, walobj)
+
+	sv.id = firstNodeID
+	sv.raftNode = raftNode
+
+	// StartNode appends to the initial log a ConfChangeAddNode entry for
+	// each peer. We can't campaign until these entries are applied,
```
kr: Might add a couple words saying that "each peer" here is actually just the current node. (It doesn't have to be, you can add as many peers as you want to start with. But the way we use the package it's always just one.)

PR #1243, opened 2017-05-25
docs: move versioned docs to main

jeffomatic:
This PR moves historical versions of our guide/reference documentation into the main branch. Some of the tooling had to be adjusted so that SDK docs (which remain versioned using historical branches) are generated and uploaded correctly.

This makes the local file structure match the online structure more closely, and allows us to skip using backport PRs to move documentation updates from `main` into a historical branch. It also opens the door for documentation that lives outside the versioning system.

After this PR is merged, we should make sure to fold the Ivy Playground documentation into `docs/1.2` before docs are uploaded again.


jeffomatic:
Most of the volume in this PR is in file copies--the ultimate change is not nearly as large as the file count would suggest. There is a clear narrative in the commit log of this branch.

cmd/docgenerate/main.go
```diff
@@ -0,0 +1,175 @@
+// Command docgenerate generates static documentation for Chain Core and the
+// Chain Core SDKs.
+//
+// Usage:
+//
+//     docgenerate chainPath outputDir
```
kr: Could/should this use `$CHAIN` instead of an arg?

kr:
This generally looks pretty nice to me.

> Most of the volume in this PR is in file copies

Hah yeah looks like most of it is from Nitti SVG files?

cmd/docgenerate/main.go
jeffomatic: I went back and forth on that myself. I think after a conversation we had earlier in the week about someone struggling with `build-cored-release`, it occurred to me that it was nice if a tool didn't have `$CHAIN` as a dependency if it didn't absolutely have to.

At any rate, the `upload-docs` script (which is probably how `docgenerate` is invoked 95% of the time) feeds `$CHAIN` as a parameter here, so I don't think we'll even have to worry about the extra verbosity in practice.

jeffomatic:
> Hah yeah looks like most of it is from Nitti SVG files?

Yeah. Those are duplicated at the moment, but a future PR can start to make the different version dirs pull assets from a common location, which I think is one nice benefit of this change.

kr:
LGTM

PR #1244, opened 2017-05-26
database/sinkdb: introduce All combinator

kr:
This will allow clients to combine conditions and
effects in constructors for composite operations, with
an interface convenient for other packages to use in a
call to Exec.

Consider a hypothetical create function:

    func create(key, value) sinkdb.Op {
        return sinkdb.All(
            sinkdb.IfNotExists(key),
            sinkdb.Set(key, value),
        )
    }

Create could return []Op, but Op is easier to
combine with further operations:

    db.Exec(ctx,
        create(key, value),
        sinkdb.Delete(key2),
    )

### DIFF
--- a/database/sinkdb/op.go
+++ b/database/sinkdb/op.go
@@ -22,6 +22,22 @@ type Op struct {
 	effects []*sinkpb.Op
 }
 
+// All encodes the atomic application of all its arguments.
+//
+// The returned Op is satisfied if all arguments would be satisfied.
+// Its effects (if satisfied) are the effects of the arguments, in order.
+func All(op ...Op) Op {
+	var outer Op
+	for _, inner := range op {
+		if inner.err != nil {
+			return inner
+		}
+		outer.conds = append(outer.conds, inner.conds...)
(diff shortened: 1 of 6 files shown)


kr:
PTAL

kr:
I was considering `And` as the name for this, but I think it's inappropriate because these aren't boolean conditions or even pure functions, they have side effects.

Also considering `AllOf` because maybe it looks nicer? 🚲 🏠 

(That may only make sense alongside a potential future combinator `One` vs `OneOf`.)

jbowens:
lgtm

I like `All` or maybe `Combine`

database/sinkdb/op.go
```diff
@@ -22,6 +22,22 @@ type Op struct {
 	effects []*sinkpb.Op
 }
 
+// All encodes the atomic application of all its arguments.
+//
+// The returned Op is satisfied if all arguments would be satisfied.
+// Its effects (if satisfied) are the effects of the arguments, in order.
```
kr: I'll remove the "in order" now that #1247 is landed.

kr:
rebased

PR #1245, opened 2017-05-26
core/asset: fix inserts for escaped strings

erykwalder:
Passing the definition as a string can cause unexpected problems with
escaped strings, so we pass it as a byte array instead.


jbowens:
Does this fix the issue because Postgres doesn't try to interpret it as a Postgres `text` type?

erykwalder:
Pretty sure, yeah.

PR #1246, opened 2017-05-26
docker/ci: fix java installation

jeffomatic:
Downloads of JDK 7 are now behind a registration wall. This commit uses
a copy that is cached to the chain-qa S3 bucket.

### DIFF
--- a/docker/ci/install/install-java
+++ b/docker/ci/install/install-java
@@ -2,16 +2,14 @@
 
 set -e
 
-curl -L0 -o jdk-7u79-linux-x64.tar.gz -H \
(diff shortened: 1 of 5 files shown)


jeffomatic:
PTAL

kr:
Oh actually, would it be possible to include a hash check a la https://github.com/chain/chain/blob/8c1ae162c0f5d915f332f4badd44336d03b2ad95/docker/ci/install/install-codesafe#L12-L14?

jeffomatic:
> Oh actually, would it be possible to include a hash check

Done!

PR #1247, opened 2017-05-26
database/sinkdb: disallow multiple writes to a key in a batch

kr:
It is extremely error-prone to write the same key more
than once in a batch. So we disallow it.


kr:
PTAL

jbowens:
lgtm

PR #1250, opened 2017-05-29
sdk/ruby: refactor test suite

dominic:
Break monolithic integration test into multiple stateless test cases.


dominic:
To do:

- [ ] Pay to receiver tests
- [ ] Transaction feed tests
- [ ] Remove `integration_spec.rb`

PR #1253, opened 2017-05-30
core/query: annotate witness arguments

danrobinson:
This PR allows witness arguments to be viewed in the SDK. This is necessary for certain applications of Ivy for which a client needs to be able to inspect the values used to satisfy a contract.


danrobinson:
PTAL

bobg:
LGTM. Wercker is failing on cored because it found an `xxx` in the Ivy compiler, which is safe to ignore.

PR #1254, opened 2017-05-30
database/sinkdb: return the version of a read key

kr:
We replace the "found" bool with a Version object that
contains more information, and update the call sites to
use the new interface.

In the future, we can define a conditional Op, IfMatch,
that checks the key's version.


kr:
PTAL

jbowens:
lgtm

PR #1255, opened 2017-05-30
net/raft: crash process on unrecoverable errors

jbowens:
cored cannot operate without an active raft service. Chain Core
leadership is maintained Postgres, so crashing the process ensures
that the leader will abdicate if its raft service fails.

Since runUpdates runs forever, there is now no way for a raft.Service
to be marked as 'done'. If we want to add a way to gracefully shutdown
a raft service in the future, we can add back the donec channel and
related logic.

### DIFF
--- a/net/raft/raft.go
+++ b/net/raft/raft.go
@@ -562,23 +562,19 @@ func (sv *Service) waitRead(ctx context.Context) error {
 
 	select {
 	case idx := <-req.index:
-		ok := sv.wait(idx)
-		if !ok {
-			return errors.New("raft shutdown")
-		}
+		sv.wait(idx)
 	case <-ctx.Done():
 		return ctx.Err()
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


kr:
:eyes:

net/raft/raft.go
```diff
@@ -613,7 +575,7 @@ func (sv *Service) waitRead(ctx context.Context) error {
 func (sv *Service) wait(index uint64) bool {
 	sv.stateMu.Lock()
 	defer sv.stateMu.Unlock()
-	for !sv.done && sv.state.AppliedIndex() < index {
+	for sv.state.AppliedIndex() < index {
 		sv.stateCond.Wait()
 	}
 	return sv.state.AppliedIndex() >= index //if false killed b/c of done signal
```
kr: this method no longer needs a return value

PR #1256, opened 2017-05-30
database/sinkdb: introduce IfMatch condition

kr:
This lets clients do safe conditional writes.

Also update net/http/authz to use the new feature.

### DIFF
--- a/database/sinkdb/op.go
+++ b/database/sinkdb/op.go
@@ -49,6 +49,22 @@ func IfNotExists(key string) Op {
 	}
 }
 
+// IfMatch encodes a conditional to make an instruction
+// successful only if the stored version matches v.
+// It checks the key returned by method Key on v.
+//
+// If v.Exists() is false,
+// IfMatch(v) is equivalent to IfNotExists(v.Key()).
+func IfMatch(v Version) Op {
+	return Op{
+		conds: []*sinkpb.Cond{{
+			Type:  sinkpb.Cond_INDEX_EQUAL,
+			Key:   v.key,
+			Index: v.n,
+		}},
+	}
+}
+
 // Delete encodes a delete operation for key.
 func Delete(key string) Op {
 	return Op{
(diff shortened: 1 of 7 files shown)


kr:
PTAL

database/sinkdb/op.go
```diff
@@ -49,6 +49,22 @@ func IfNotExists(key string) Op {
 	}
 }
 
+// IfMatch encodes a conditional to make an instruction
+// successful only if the stored version matches v.
+// It checks the key returned by method Key on v.
```
kr: This is awkward wording.

kr:
```
IfMatch encodes a conditional to make an instruction
successful only if v's key's stored version matches v.
```

kr:
```
IfMatch encodes a conditional to make an instruction
successful only if the version stored in v's key matches v.
```

database/sinkdb/op.go
```diff
@@ -49,6 +49,22 @@ func IfNotExists(key string) Op {
...
+//
+// If v.Exists() is false,
+// IfMatch(v) is equivalent to IfNotExists(v.Key()).
+func IfMatch(v Version) Op {
```
jbowens: 🚲 🏠  any thoughts on `IfUnmodified`?

kr: `IfUnmodified` is fine by me.

kr: or `IfNotModified` to match `IfNotExists`?

jbowens:
lgtm, tho `IfMatch` feels a little ambiguous

database/sinkdb/op.go
kr: We could hypothetically also have `IfExists` and `IfModified` though those ones seem much less useful.

kr: (I started with IfMatch because that's what HTTP calls the equivalent header field, but that's not a very good reason.)

PR #1257, opened 2017-05-31
sdk/node: create account specific test

chrisgarvin:


sdk/node/test/accounts.js
```diff
@@ -0,0 +1,246 @@
...
+  describe('Promise style', () => {
+
+    before('set up API objects', () => {
+
+      // Key creation and signer setup
+      return expect(Promise.all([
+        client.mockHsm.keys.create({alias: aliceAlias}),
+        client.mockHsm.keys.create({alias: bobAlias}),
+        client.mockHsm.keys.create()
+      ])).to.be.fulfilled
```
jeffomatic: How necessary is the `.to.be.fulfilled` here? If we remove it, and one of the above function calls fails (maybe a duplicate alias), will the `before` statement still fail?

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,246 @@
...
+
+    describe('Batch account creation', () => {
+      let batchResponse = {}
+
+      before(() => client.accounts.createBatch([
+        {alias: `carol-${uuid.v4()}`, rootXpubs: [otherKey.xpub], quorum: 1}, // success
+        {alias: 'david'}, // failure
+        {alias: `eve-${uuid.v4()}`, rootXpubs: [otherKey.xpub], quorum: 1}, // success
+      ])
+      .then(resp => {batchResponse = resp}))
```
jeffomatic: Maybe there's a way to fix the indentation here so that `.then` does not appear to chain off of the `before`?

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,246 @@
...
+            })
+          ).to.be.fulfilled
+        })
+        .then(page => {
+          assert.deepEqual(page.items[0].tags, {x: 1})
+        })
+      })
+
+      it('fails to update account tags', () => {
+        return Promise.resolve()
```
jeffomatic: I think the `it` callback just needs to `return expect(...)`, everything else might just be redundant.

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,246 @@
...
+      })
+    })
+
+    describe('Single account creation', () => {
+
+      it('creates two accounts', () => {
+        return expect(Promise.all([
+          client.accounts.create({alias: aliceAlias, rootXpubs: [aliceKey.xpub], quorum: 1}),
+          client.accounts.create({alias: bobAlias, rootXpubs: [bobKey.xpub], quorum: 1})
+        ])).to.be.fulfilled
```
dominic: I think we can actually remove all of the `expected(...).to.be.fulfilled` from our entire suite. The main purpose of this was to have our single giant test print out a line number when at the Promise that failed instead of the end of the test.

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,246 @@
...
+        signer.addKey(bobKey, client.mockHsm.signerConnection)
+      })
+    })
+
+    describe('Single account creation', () => {
+
+      it('creates two accounts', () => {
+        return expect(Promise.all([
+          client.accounts.create({alias: aliceAlias, rootXpubs: [aliceKey.xpub], quorum: 1}),
+          client.accounts.create({alias: bobAlias, rootXpubs: [bobKey.xpub], quorum: 1})
```
dominic: This is probably a holdover from the old test, but we only really need to create one account here. We can get rid of `bobAlias` and the associated account here.

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,246 @@
...
+      return expect(Promise.all([
+        client.mockHsm.keys.create({alias: aliceAlias}),
+        client.mockHsm.keys.create({alias: bobAlias}),
+        client.mockHsm.keys.create()
+      ])).to.be.fulfilled
+
+      .then(keys => {
+        aliceKey = keys[0]
+        bobKey = keys[1]
+        otherKey = keys[2]
```
dominic: Could we get away with just one mock HSM key for these tests? 3 seems excessive.

sdk/node/test/accounts.js
chrisgarvin: yes, you're correct. the `before` statement will fail if we do that.. think we can rid of it then

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,246 @@
...
+    })
+
+    describe('Single account tags update', () => {
+
+      it('successfully updates account tags', () => {
+        return Promise.resolve()
+
+        .then(() => {
+          return expect(
+            client.accounts.create({
```
dominic: We could probably create two accounts in a before block here and remove that from single and batch account tag test cases.

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,177 @@
...
+  describe('Promise style', () => {
+
+    before('set up API objects', () => {
+
+      // Key and account creation
+      return client.mockHsm.keys.create()
+      .then(key => {
+        mockHsmKey = key
+
+        client.accounts.createBatch([
```
jeffomatic: I think you need a `return` here, or else the promise chain is unlikely to fail properly.

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,177 @@
...
+
+describe('Accounts test', () => {
+
+  describe('Promise style', () => {
+
+    before('set up API objects', () => {
+
+      // Key and account creation
+      return client.mockHsm.keys.create()
+      .then(key => {
```
jeffomatic: Nit: might be good to indent one level here.

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,177 @@
...
+
+    describe('Single account creation', () => {
+
+      it('account creation successful', () => {
+        return client.accounts.create({alias: `alice-${uuid.v4()}`, rootXpubs: [mockHsmKey.xpub], quorum: 1})
+      })
+
+      it('account creation rejected due to missing key fields', () => {
+        return expect(client.accounts.create({alias: 'david'}))
+        .to.be.rejectedWith('CH202')
```
jeffomatic: indent

sdk/node/test/accounts.js
```diff
@@ -0,0 +1,177 @@
...
+      })
+    })
+  })
+
+  // These just test that the callback is engaged correctly. Behavior is
+  // tested in the promises test.
+  describe('Callback style', () => {
+
+    it('Single account creation', (done) => {
+      client.accounts.create(
```
jeffomatic: If we are just trying to prove that the callback is being invoked at all, this is one alternative:

```
client.accounts.create(
  {}, // intentionally blank,
  done
)
```

Like I said in chat, I'm not sure which one I prefer...what do you think?

sdk/node/test/accounts.js
chrisgarvin: instead of just returning right there, I think it may be better to go this route:
```
    before('set up API objects', () => {

      // Key and account creation
      return client.mockHsm.keys.create()
        .then(key => { mockHsmKey = key })
        .then(() => {
          return client.accounts.create({alias: xAccountAlias, rootXpubs: [mockHsmKey.xpub], quorum: 1, tags: {x: 0}})
        })
        .then(() => {
          return client.accounts.create({alias: yAccountAlias, rootXpubs: [mockHsmKey.xpub], quorum: 1, tags: {y: 0}})
        })
    })
```
`createBatch` won't fail if for instance i don't provide a key for one of the accounts, which would then trigger future tests failing for not very clear reasons

jeffomatic: LGTM

PR #1258, opened 2017-05-31
net/raft: respect context deadline in WaitRead

jbowens:
In WaitRead we ignore context.DeadlineExceeded errors returned from
waitRead because we use context.WithDeadline for timing out
individual retries. It's also possible that the parent context has
expired, so return the parent context's error if that happens.


kr:
LGTM

PR #1262, opened 2017-05-31
core: exec self in /join-cluster

jbowens:
If the joined cluster is already configured, the process will still be
running as an unconfigured Core process. Exec self to restart and
reload the config, if any. In the long term, configuration should be
incremental and not require a process restart to pick up changes.


kr:
LGTM

PR #1263, opened 2017-05-31
net/raft: add default WaitRead timeout

jbowens:
If no context deadline is set, use a 30 second timeout in WaitRead.
Otherwise, WaitRead may block forever.


jbowens:
PTAL

kr:
LGTM

PR #1265, opened 2017-06-01
sdk/node: create asset specific test

chrisgarvin:


sdk/node/test/assets.js
```diff
@@ -0,0 +1,159 @@
...
+  })
+
+  // These just test that the callback is engaged correctly. Behavior is
+  // tested in the promises test.
+  describe('Callback style', () => {
+
+    it('Single asset creation', (done) => {
+      client.assets.create(
+        {}, // intentionally blank
+        () => done()
```
jeffomatic: You can just pass `done` here. No need to wrap it in a lambda.

PR #1268, opened 2017-06-01
docs/1.2: remove corectl init, corectl join

jbowens:
These commands and their documentation were added in
d20cecb8ca646d547b9031f9db9180753245115d, in the 1.3 development branch.
When the documentation was reorganized, it seems that it was cut from
main and the documentation for these commands snuck in.

### DIFF
--- a/docs/1.2/core/reference/cored.md
+++ b/docs/1.2/core/reference/cored.md
@@ -83,6 +83,13 @@ the limit will receive an HTTP 429 response.
 
     Can be stacked with **RATELIMIT_TOKEN**.
 
+* **BOOTURL**: Setting this value causes the `cored` process to join an
+existing Chain Core cluster if it's not already a member. If it is already
+a member of a cluster, this has no effect.
+
+    This should be the URL of any `cored` process already in the cluster,
+    or a load balancer that forwards requests to any node.
+
 ## Mutual TLS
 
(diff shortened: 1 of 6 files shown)


jeffomatic:
Oops, sorry. Any documentation updates on `main` since the 1.2 release and prior to a8bd8fe76d1a074add322ea0d00e8c234c4cbbae are likely to have been lost.

jeffomatic:
Looking at the logs, I think only d20cecb8ca646d547b9031f9db9180753245115d was affected by this. Sorry again!

jbowens:
No worries!

jeffomatic:
There was also [a change](https://github.com/chain/chain/commit/d20cecb8ca646d547b9031f9db9180753245115d#diff-83f23f7fc2c80a2458881c3974ab00c5) in d20cecb8ca646d547b9031f9db9180753245115d that affected the `cored` reference as well. You may want to add that in here as well.

jbowens:
Whoop, good catch

docs/1.2/core/reference/cored.md
```diff
@@ -83,6 +83,13 @@ the limit will receive an HTTP 429 response.
 
     Can be stacked with **RATELIMIT_TOKEN**.
 
+* **BOOTURL**: Setting this value causes the `cored` process to join an
```
jeffomatic: Hmm was I wrong? It looks like the original commit _removed_ this paragraph?

jbowens: In Core 1.2, to join an existing a cluster you provided this environment variable. In d20cecb we removed this environment variable, replacing it with the `corectl join` command.

jbowens: We [decided not to keep the environment variable](https://github.com/chain/chain/pull/1229#issuecomment-303261089) for backwards-compatibility because it's only used when setting up a new cluster.

Issue #1118, opened 2017-05-02
raft: ability to remove members


tessr:
At the moment, members of the raft cluster can never leave. (It's like Hotel California in here!)

In order to let members leave the raft cluster, we need to:

- [x] Add an `/remove` (or "evict," but I prefer "remove") endpoint to the raft service 
- [ ] Add the ability to remove a member from the allowed members list
- [ ] As part of removing a member from the allowed members list, also revoke that member's grant, if applicable

This last piece is slightly tricky, because several members may share a single grant; namely, if two members are running on the same host but under different ports (as they often do in development), they will have an identical CN on their x509 certificates and therefore share a single grant. When one of these members is removed from the allowed members list, we need to recalculate the list of grants and recognize that the associated grant cannot be deleted.

PR #1269, opened 2017-06-01
core: add /evict rpc endpoint

jbowens:
Add an evict command to corectl to remove a node from a Chain Core
cluster.

On startup in cmd/cored, call sinkdb.DB.Ping to perform an empty write
to test the raft configuration. This helps handle the scenario where an
evicted node boots up without realizing it's been evicted. The other
nodes won't necessarily replicate the eviction entry to the node if
they've already moved on and forgotten the evicted node's address.

Addresses part of #1118.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -63,6 +63,7 @@ var commands = map[string]*command{
 	"revoke":               {revoke},
 	"join":                 {joinCluster},
 	"init":                 {initCluster},
+	"evict":                {evictNode},
 	"allow-address":        {allowRaftMember},
 	"wait":                 {wait},
 }
@@ -403,6 +404,19 @@ func joinCluster(client *rpc.Client, args []string) {
 
 	req := map[string]string{"boot_address": args[0]}
 	err := client.Call(context.Background(), "/join-cluster", req, nil)
+	dieOnRPCError(err)
+}
+
+// evictNode evicts a Chain Core cored process from the cluster.
+// It does not modify the allowed-member list.
+func evictNode(client *rpc.Client, args []string) {
+	const usage = "usage: corectl evict [node address]"
+	if len(args) != 1 {
+		fatalln(usage)
+	}
+
+	req := map[string]string{"node_address": args[0]}
+	err := client.Call(context.Background(), "/evict", req, nil)
 	dieOnRPCError(err)
 }
 
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -249,6 +249,13 @@ func main() {
 		err := server.Serve(listener)
 		chainlog.Fatalkv(ctx, chainlog.KeyError, errors.Wrap(err, "Serve"))
 	}()
+
+	// Verify that we're connected to the rest of the cluster, if initialized.
+	err = errors.Root(sdb.Ping())
+	if err == context.DeadlineExceeded {
+		chainlog.Fatalkv(ctx, chainlog.KeyError, "Unable to reach rest of raft cluster. Was the node evicted?")
+	}
+
 	resetIfAllowedAndRequested(db, sdb)
 
 	conf, err := config.Load(ctx, db, sdb)
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -254,6 +254,8 @@ func main() {
 	err = errors.Root(sdb.Ping())
 	if err == context.DeadlineExceeded {
 		chainlog.Fatalkv(ctx, chainlog.KeyError, "Unable to reach rest of raft cluster. Was the node evicted?")
+	} else if err != nil && err != raft.ErrUninitialized {
+		chainlog.Fatalkv(ctx, chainlog.KeyError, err)
 	}
 
 	resetIfAllowedAndRequested(db, sdb)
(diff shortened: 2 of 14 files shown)


database/sinkdb/state.go
```diff
@@ -34,25 +34,38 @@ func newState() *state {
...
 
-// SetPeerAddr sets the address for the given peer.
-func (s *state) SetPeerAddr(id uint64, addr string) {
+// Peers returns the current set of peer nodes. The returned
+// map must not be modified.
+func (s *state) Peers() map[uint64]string {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	s.peers[id] = addr
+	return s.peers
```
jbowens: I made this change so that Evict can lookup a node ID by address. Also, looking forward, I think we'll often need to grab the entire list of peers for things like broadcasting to the entire cluster. Since changes to the peer list should be infrequent, I think it's okay to make a new copy on every change.

jbowens:
PTAL

cmd/cored/main.go
```diff
@@ -249,6 +249,13 @@ func main() {
 		err := server.Serve(listener)
 		chainlog.Fatalkv(ctx, chainlog.KeyError, errors.Wrap(err, "Serve"))
 	}()
+
+	// Verify that we're connected to the rest of the cluster, if initialized.
+	err = errors.Root(sdb.Ping())
+	if err == context.DeadlineExceeded {
+		chainlog.Fatalkv(ctx, chainlog.KeyError, "Unable to reach rest of raft cluster. Was the node evicted?")
```
kr: It would be pretty cool if we could get a definitive "you are not a member" message somehow. I guess that would take a bunch of plumbing, including looking at the http response in sendmsg. Maybe a todo for that? Or is there a better idea?

cmd/cored/main.go
```diff
@@ -249,6 +249,13 @@ func main() {
 		err := server.Serve(listener)
 		chainlog.Fatalkv(ctx, chainlog.KeyError, errors.Wrap(err, "Serve"))
 	}()
+
+	// Verify that we're connected to the rest of the cluster, if initialized.
+	err = errors.Root(sdb.Ping())
+	if err == context.DeadlineExceeded {
```
kr: What if there's some other error here?

cmd/corectl/main.go
```diff
@@ -406,6 +407,19 @@ func joinCluster(client *rpc.Client, args []string) {
 	dieOnRPCError(err)
 }
 
+// evictNode evicts a Chain Core cored process from the cluster.
+// It does not modify the allowed-member list.
```
kr: Should it modify the allowed member list? That might be a nice follow up.

jbowens: I was thinking since allow-address and init/join are separate actions, evict and disallow-address should be separate.

They do seem like potentially semantically different things. If for example a host running cored dies, you might evict the node to reconfigure the quorum. You might replace the host and reuse the same address.

cmd/cored/main.go
jbowens: 👍 , I'll add a TODO for now

cmd/cored/main.go
jbowens: I'll add a Fatalkv call for all non ErrUninitialized errors

jbowens:
PTAL

database/sinkdb/sinkdb.go
```diff
@@ -34,6 +35,17 @@ type DB struct {
 	raft  *raft.Service
 }
 
+// Ping peforms an empty write to verify the connection to
+// the rest of the cluster.
+func (db *DB) Ping() error {
+	const timeout = 30 * time.Second
```
tessr: This seems long to me,  but  ¯\\\_(ツ)\_/¯

database/sinkdb/state.go
```diff
@@ -41,25 +41,38 @@ func (s *state) SetAppliedIndex(index uint64) {
...
 }
 
-// GetPeerAddr gets the current address for the given peer, if set.
-func (s *state) GetPeerAddr(id uint64) (addr string) {
+// SetPeerAddr sets the address for the given peer.
+func (s *state) SetPeerAddr(id uint64, addr string) {
+	newPeers := make(map[uint64]string)
```
tessr: why make a whole new map?

database/sinkdb/state.go
```diff
@@ -41,25 +41,38 @@ func (s *state) SetAppliedIndex(index uint64) {
...
+	for nodeID, addr := range s.peers {
+		newPeers[nodeID] = addr
+	}
+	newPeers[id] = addr
+	s.peers = newPeers
 }
 
 // RemovePeerAddr deletes the current address for the given peer if it exists.
 func (s *state) RemovePeerAddr(id uint64) {
+	newPeers := make(map[uint64]string)
```
tessr: same question ^

database/sinkdb/state.go
jbowens: Since Go maps aren't safe for concurrent access, readers would need to keep s.mu locked while accessing the map. [This change](https://github.com/chain/chain/pull/1269/files#r119689333) would make that interface really clunky. By forcing all mutations of the map to create a new map, readers can call Peers and use the returned map without keeping a hold of s.mu. Changes to the peer list should be infrequent, so it seems like a reasonable tradeoff.

tessr: Okay! Cool. ⛄️

PR #1270, opened 2017-06-01
database/sinkdb: use map key existence for Exists

jbowens:
For now use the existence of the key in the state map for
Version.Exists. Once we have a backfill for Chain Core 1.2.x
snapshots that omit versions, we can use the version check
instead.


kr:
LGTM

PR #1271, opened 2017-06-01
sdk/node: create access token specific test

chrisgarvin:


sdk/node/test/accessTokens.js
```diff
@@ -0,0 +1,139 @@
...
+const networkTokenId = `network-${uuid.v4()}`
+
+describe('Access tokens test', () => {
+
+  describe('Promise style', () => {
+
+    describe('Access token creation', () => {
+
+      it('access token creation successful', () => {
+        return client.accessTokens.create({
```
dominic: I can go back and forth on this, but I'd like to lean heavily towards always* having an explicit assertion in test cases? For example, in the [Ruby test for access token creation](https://github.com/chain/chain/blob/main/sdk/ruby/spec/integration/access_tokens_spec.rb#L4-L6), we test that we get a token value back. 

\* Excepting the callback tests, where we've accepted that the implicit assertion of "success" is good enough.

sdk/node/test/accessTokens.js
```diff
@@ -0,0 +1,139 @@
...
+          type: 'client',
+          id: tokenId
+        })
+      })
+
+      it('access token creation rejected due to duplicate ID', () => {
```
dominic: This test, and all the rest of the file, is depending on the order of the tests to be successful. That's not a huge problem, but one of the goals of this refactor is to be able to run our tests cases in a random order. For a little background, running tests in random order is a really great way to catch more intermittent failures in the underlying software, where operations performed in a certain order could cause unexpected results.

However, it's not that hard to change it up. One easy way to accomplish this is to give each `describe` block a `beforeEach` hook that creates any dependency needed for each test independently. For example, for the first few tests, it might look like this:

```
describe('Promise style', () => {
  beforeEach(() => {
    tokenId = `token-${uuid.v4()}`
  }

  describe('Access token creation', () => {
    beforeEach(() => {
      client.accessTokens.create({
        type: 'client',
        id: tokenId
      })
    })

    it('access token creation successful', () => {
      return client.accessTokens.create({
        type: 'client',
        id: `another-${tokenId}`
      })
    })

    it('access token creation rejected due to duplicate ID', () => {
      return expect(client.accessTokens.create({
        type: 'client',
        id: tokenId
      }))
      .to.be.rejectedWith('CH302')
    })

    it('access token returned in list after creation', () => {
      return client.accessTokens.query()
        .then(resp => expect(resp.items.map(item => item.id)).to.contain(tokenId))
    })
  })

  ...
}
```

Note that we re-create `tokenId` per test as well, so that each test is looking for its own unique UUID value. Using two different `beforeEach` blocks, we can make sure every test has its own data.

PR #1275, opened 2017-06-02
net/http/authz: encapsulate grant storage

jbowens:
Hide grant storage details from core, keeping the grant prefix as an
unexported constant in net/http/authz. Always use authz.Store methods
for accessing and modifying grants.

As written, the list grants endpoint will perform a stale read. Not sure
how we feel about that.

### DIFF
--- a/net/http/authz/store.go
+++ b/net/http/authz/store.go
@@ -9,20 +9,21 @@ import (
 	"chain/errors"
 )
 
+const keyPrefix = `/core/grant/`
+
 // Generate code for the Grant and GrantList types.
 //go:generate protoc -I. -I$CHAIN/.. --go_out=. grant.proto
 
 // Store provides persistent storage for grant objects.
 type Store struct {
-	sdb       *sinkdb.DB
-	keyPrefix string
+	sdb *sinkdb.DB
 }
 
 // NewStore returns a new *Store storing grants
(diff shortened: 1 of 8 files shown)


net/http/authz/store.go
```diff
@@ -9,28 +9,29 @@ import (
 	"chain/errors"
 )
 
+const keyPrefix = `/core/grant/`
```
kr: Minor, but I feel like this actual path is specific to chain core, whereas this package could conceivably be used in other products. (Consider if signerd had a database, maybe it would put grants in `/signerd/grant/`.)

This isn't as bad as actually importing a package in `chain/core/...`; it's just a constant string and not hard to change later if necessary. So okay. But maybe we should have a todo about factoring this out?

Ideally imo we would instantiate the grant Store in just one place, and that place would dictate the path prefix. Currently it seems like that would require an annoying amount of plumbing between chain/cmd/cored and chain/core and chain/core/config, so I'm fine with not doing that for now.

jbowens: Ah, I see. Yeah, this is sort of analogous to hard-coding a specific SQL table name into the net/http/authz package, which isn't ideal. I'll add a TODO for now.

jbowens:
Might try another direction

PR #1276, opened 2017-06-02
sdk/node: create authorization grant specific test

chrisgarvin:


jeffomatic:
Couple quick questions/comments:

1. Are you avoiding removing anything from the single-file integration tests before you've moved everything over?
2. Nit about information architecture: I think we could consider removing the "Promise-style" level of nesting, and promoting its children to top level of the suite. I think we're not trying to distinguish between promise- and callback-style anymore. We can treat the fact that the interface supports callbacks as just another feature alongside everything else.

chrisgarvin:
image

sdk/node/test/accessTokens.js
```diff
@@ -12,141 +12,141 @@ const client = new chain.Client()
 
 let tokenId, clientTokenId, networkTokenId
 
-describe('Access tokens test', () => {
+function createToken() {
+  tokenId = `token-${uuid.v4()}`
```
jeffomatic: We should try to figure out an alternative to setting a global variable here. Suggestion: wherever you use this function, get the token ID in the `then` clause and set a variable that is local to the `it` example's callback. I think this may eliminate the need for your global variables.

sdk/node/test/accessTokens.js
```diff
@@ -12,141 +12,141 @@ const client = new chain.Client()
...
+  return client.accessTokens.create({
+    type: 'client',
```
jeffomatic: Setting the token type is deprecated, and shouldn't be used except in the deprecated behavior tests.

sdk/node/test/accessTokens.js
```diff
@@ -10,143 +10,147 @@ const expect = chai.expect
...
-
-      it('creation successful', () => {
-        return client.accessTokens.create({
-          type: 'client',
-          id: `another-${tokenId}`
-        }).then(resp => expect(resp.token).not.to.be.empty)
-      })
+  it('creation successful', () => {
+    return client.accessTokens.create({
+      type: 'client',
```
jeffomatic: The access token type is deprecated, so it shouldn't be used unless we're explicitly testing the deprecated interface.

sdk/node/test/authorizationGrants.js
```diff
@@ -0,0 +1,111 @@
...
+
+const client = new chain.Client()
+
+let tokenName, tokenGrant
+
+describe('Authorization grant', () => {
+
+  before('set up grant data', () => {
+    tokenName = uuid.v4()
+    return client.accessTokens.create({type: 'client', id: tokenName}).then(resp => {
```
jeffomatic: `type` is deprecated

PR #1277, opened 2017-06-02
dashboard: add consistent login test on app start

dominic:
Extract responsibilities from `features/app` to create two new app features:

* `container`: manages high level loading -> login -> config -> app flow.
* `authn`: manages testing what type of authentication is required, and 
    the client token form. 

`authn` now displays the login page at `/login` instead of at whatever path was 
requested, and allows `container` to choose where to redirect once logged in.


jeffomatic:
May or may not be out-of-scope for this PR, but we have three components that could be better named: `Root`, `Container`, and `Main`.

dominic:
@jeffomatic Agreed - any thoughts on those though? They're all kinda high level. `App` > `StateManager` > `Dashboard`?

dashboard/src/features/container/routes.js
```diff
@@ -0,0 +1,20 @@
+import Container from './components/Container'
+
+import { getAuthenticationStatus } from 'features/authn/actions'
+import { fetchCoreInfo } from 'features/core/actions'
+import { fetchTestnetInfo } from 'features/testnet/actions'
+
+const CORE_POLLING_TIME = 2 * 1000
+const TESTNET_INFO_POLLING_TIME = 30 * 1000
+
+export default (store) => ({
```
jeffomatic: We never got to finish the dashboard refactoring, but I wonder if there's a chance to make this new component look like what we want. Here's what I think we discussed in the past:

1. Explicitly name your exported variables, so that it has the same identifier that is likely to be used in the importing file (in this case, `makeRoutes()`).
2. If possible, naming your files to be the same as the exported identifier (in this case, `makeRoutes.js`).
3. Avoiding the use of index files, so there is only path for importing.

I can't remember if there was a stable consensus on those principles, especially the last one, but the above reflect my gut feeling as of today.

dashboard/src/features/shared/components/PageTitle/PageTitle.jsx
```diff
@@ -47,7 +47,7 @@ class PageTitle extends React.Component {
 }
 
 const mapStateToProps = (state) => {
-  const routes = makeRoutes()
+  const routes = makeRoutes().childRoutes.find(route => route.useForBreadcrumbs)
```
jeffomatic: `routes` feels like not quite the right name here, as `find` returns a singular thing.

dashboard/src/features/authn/reducers.js
```diff
@@ -0,0 +1,54 @@
...
+const clientToken = (state = '', action) => {
+  if      (action.type == 'SET_CLIENT_TOKEN') return action.token
+  else if (action.type == 'ERROR' &&
+           action.payload.status == 401)      return ''
+
+  return state
+}
+
+const authenticationReady = (state = false, action) => {
+  if (action.type == 'AUTHN_LOADED') return true
```
jeffomatic: Above there's a property called `authenticationRequired`, which is set by an action called `AUTHENTICATION_REQUIRED`. Here, there's a property called `authenticationReady` that is set by an action called `AUTHN_LOADED`...maybe we can make the action name the same as the property?

dashboard/src/features/authn/reducers.js
```diff
@@ -0,0 +1,54 @@
...
+
+  return state
+}
+
+
+// const validToken = (state = false, action) => {
```
jeffomatic: Do we still need this?

dashboard/src/features/authn/actions.js
```diff
@@ -0,0 +1,34 @@
+import { unauthedClient } from 'utility/environment'
+import { fetchCoreInfo } from 'features/core/actions'
+
+const setClientToken = (token) => ({type: 'SET_CLIENT_TOKEN', token})
```
jeffomatic: Is it worth the extra level of indirection here to make an actual action creator function?

dashboard/src/features/authn/actions.js
```diff
@@ -0,0 +1,34 @@
...
+
+export const clearSession = ({ type: 'USER_LOG_OUT' })
```
jeffomatic: Would it be alright to dispatch `{type: 'USER_LOG_OUT'}` wherever `clearSession` is used? Just want to make sure there is value in creating a whole new concept here.

dashboard/src/features/container/routes.js
```diff
@@ -0,0 +1,20 @@
...
+  path: '/',
+  component: Container,
+  onEnter: () => {
```
jeffomatic: Just to clarify, this only ever gets called once for the lifetime of the whole app, right? If you want this logic to run again, you have to reload the page?

dashboard/src/features/authn/actions.js
dominic: Nope; that's just a holdover from the previous implementation.

dashboard/src/features/container/routes.js
dominic: Correct. After checking the configuration on launch, the state is updated correctly. The most important change is checking for 401s with an unauthenticated client on launch.

jeffomatic:
Code LGTM but I think I want to test this out a bit before I merge, just because it's a relatively large change.

jeffomatic:
Okay I did some testing and it looks good, but the lack of a create-token modal is something to look into next.

PR #1279, opened 2017-06-02
protocol/vm/vmutil: move under vm package

erykwalder:
Just organization to clarify that the vmutil package is for vm1
programs.


kr:
:+1:

PR #1280, opened 2017-06-02
cmd/docsync: new command

dominic:
Add support for zero-downtime doc deploys by uploading files one at a time 
with correct the content type, and removing files not present on the local 
system after the fact.

### DIFF
--- a/cmd/docsync/main.go
+++ b/cmd/docsync/main.go
@@ -14,7 +14,6 @@ import (
 
 func main() {
 	localKeys := mustListContents("/Users/jeff/go/src/chain/cmd/docsync/DO_NOT_COMMIT")
-	fmt.Println(len(localKeys))
 
 	region := "us-east-1"
 	sess := session.Must(session.NewSession(aws.NewConfig().WithRegion(region)))
@@ -34,53 +33,61 @@ func main() {
 		log.Fatalln("s3 list objects error:", err)
 	}
 
-	intersection, localOnly, remoteOnly := setAnalyze(localKeys, remoteKeys)
-	fmt.Println("local size:", len(localKeys))
-	fmt.Println("remote size:", len(remoteKeys))
-	fmt.Println("intersection:", len(intersection))
-	fmt.Println("local only:", len(localOnly))
-	fmt.Println("remote only:", len(remoteOnly))
+	var prefixedLocalKeys []string
+	for _, k := range localKeys {
+		prefixedLocalKeys = append(prefixedLocalKeys, path.Join("docs", k))
+	}
+	fmt.Println("keys to upload:", len(prefixedLocalKeys)) // TEMP
+
+	remoteOnly := setDiff(remoteKeys, prefixedLocalKeys)
+	fmt.Println("keys to delete:", len(remoteOnly)) // TEMP
+
+	// TODO:
+	// 1. upload localKeys to prefixedLocalKeys, using a default content type of
+	//    text/html for extensionless files.
+	// 2. remove remoteOnly keys
+	// 3. Make local directory and bucket configurable.
 }
 
-func setAnalyze(a, b []string) (intersection []string, aOnly []string, bOnly []string) {
-	sort.Strings(append([]string{}, a...))
-	sort.Strings(append([]string{}, b...))
+func setDiff(a, b []string) []string {
+	// don't modify input
+	a = append([]string{}, a...)
+	b = append([]string{}, b...)
+	sort.Strings(a)
+	sort.Strings(b)
 
-	fmt.Println(a[0])
-	fmt.Println(b[0])
-
-	var i, j int
+	var (
+		diff []string
+		i, j int
+	)
 
 	for {
 		if i == len(a) {
-			bOnly = append(bOnly, b[j:]...)
 			break
 		}
 
 		if j == len(b) {
-			aOnly = append(aOnly, a[i:]...)
+			diff = append(diff, a[i:]...)
 			break
 		}
 
 		if a[i] < b[j] {
-			aOnly = append(aOnly, a[i])
+			diff = append(diff, a[i])
 			i++
 			continue
 		}
 
 		if b[j] < a[i] {
-			bOnly = append(bOnly, b[j])
 			j++
 			continue
 		}
 
 		// invariant: a[i] == b[j]
-		intersection = append(intersection, a[i])
 		i++
 		j++
 	}
 
-	return
+	return diff
 }
 
 func mustListContents(parentPath string) []string {
--- a/cmd/docsync/main.go
+++ b/cmd/docsync/main.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"path"
+	"sort"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+func main() {
+	localKeys := mustListContents("/Users/jeff/go/src/chain/cmd/docsync/DO_NOT_COMMIT")
+	fmt.Println(len(localKeys))
+
+	region := "us-east-1"
+	sess := session.Must(session.NewSession(aws.NewConfig().WithRegion(region)))
+	svc := s3.New(sess)
+
+	var remoteKeys []string
+	err := svc.ListObjectsPages(&s3.ListObjectsInput{
+		Bucket: aws.String("chain.com"),
+		Prefix: aws.String("docs/"),
+	}, func(page *s3.ListObjectsOutput, lastPage bool) bool {
+		for _, obj := range page.Contents {
+			remoteKeys = append(remoteKeys, *obj.Key)
+		}
+		return true
+	})
+	if err != nil {
+		log.Fatalln("s3 list objects error:", err)
+	}
+
+	intersection, localOnly, remoteOnly := setAnalyze(localKeys, remoteKeys)
+	fmt.Println("local size:", len(localKeys))
+	fmt.Println("remote size:", len(remoteKeys))
+	fmt.Println("intersection:", len(intersection))
+	fmt.Println("local only:", len(localOnly))
+	fmt.Println("remote only:", len(remoteOnly))
+}
+
+func setAnalyze(a, b []string) (intersection []string, aOnly []string, bOnly []string) {
+	sort.Strings(append([]string{}, a...))
+	sort.Strings(append([]string{}, b...))
+
+	fmt.Println(a[0])
+	fmt.Println(b[0])
+
+	var i, j int
+
+	for {
+		if i == len(a) {
+			bOnly = append(bOnly, b[j:]...)
+			break
+		}
+
+		if j == len(b) {
+			aOnly = append(aOnly, a[i:]...)
+			break
+		}
+
+		if a[i] < b[j] {
+			aOnly = append(aOnly, a[i])
+			i++
+			continue
+		}
+
+		if b[j] < a[i] {
+			bOnly = append(bOnly, b[j])
+			j++
+			continue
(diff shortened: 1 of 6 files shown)


cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+			i++
+			continue
+		}
+
+		if b[j] < a[i] {
+			j++
+			continue
+		}
+
+		// invariant: a[i] == b[j]
```
kr: This isn't a loop invariant, it's just the condition that happens to be true at this point.

Also, nbd but I think it would work just as well to say

```
if a[i] < b[j] {
  diff = append(diff, a[i])
  i++
} else {
  j++
}
```

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	fmt.Println("keys to upload:", len(prefixedLocalKeys)) // TEMP
+
+	remoteOnly := setDiff(remoteKeys, prefixedLocalKeys)
+	fmt.Println("keys to delete:", len(remoteOnly)) // TEMP
+
+	for _, k := range prefixedLocalKeys {
+		var body []byte
+
+		path := strings.Replace(k, "docs", os.Args[2], 1)
```
jeffomatic: Rather than iterate over `prefixedLocalKeys` and remove the `docs` prefix, I think it might be easier to iterate over `localKeys` and just add the prefix back when uploading to the s3 bucket.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+		}
+		return true
+	})
+
+	if err != nil {
+		log.Fatalln("s3 list objects error:", err)
+	}
+
+	var prefixedLocalKeys []string
```
jeffomatic: I think you only need to calculate this when performing the deletes, so you can move this paragraph so it's just above the deletion paragraph.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	remoteOnly := setDiff(remoteKeys, prefixedLocalKeys)
+	fmt.Println("keys to delete:", len(remoteOnly)) // TEMP
+
+	for _, k := range prefixedLocalKeys {
+		var body []byte
+
+		path := strings.Replace(k, "docs", os.Args[2], 1)
+		body, err = ioutil.ReadFile(path)
+		if err != nil {
+			fmt.Println(err.Error())
```
jeffomatic: This will terminate the program with a success status. If you want this to terminate with an error message, you could try `log.Fatalln()`, which will print something to stderr and then terminate the program with a non-zero status.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+		} else {
+			upload.SetContentType(contentType)
+		}
+
+		fmt.Println("uploading ", k, " with type ", contentType)
+
+		_, err = svc.PutObject(upload)
+
+		if err != nil {
+			fmt.Println(err.Error())
```
jeffomatic: See above re: error handling.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+			Bucket: aws.String(os.Args[1]),
+			Key:    aws.String(k),
+		}
+
+		fmt.Println("deleting ", k)
+
+		_, err = svc.DeleteObject(remove)
+
+		if err != nil {
+			fmt.Println(err.Error())
```
jeffomatic: See above re: error handling.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	"sort"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+func main() {
+	localKeys := mustListContents(os.Args[2])
```
jeffomatic: Nit: it might help to bind `os.Args[1]` and `os.Args[2]` to variable names that describe what the params mean.

cmd/docsync/main.go
jeffomatic: Oops, that was my bad. I think maybe we can clean everything up if we do something like this:

```
if a[i] < b[j] {
  diff = append(diff, a[i])
  i++
} else if b[j] < a[i] {
  j++
} else { // a[i] == b[j]
  i++
  j++
}
```

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
+package main
```
jeffomatic: There's a whole style of documenting files that implement commands like this; see `docgenerate` for one such example: https://github.com/chain/chain/blob/main/cmd/docgenerate/main.go

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	"mime"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
```
jeffomatic: Before we land this PR, we're gonna want a separate PR that vendors the AWS s3 package, since (I think?) it hasn't been vendored yet. I haven't gone through the vendoring workflow in its current version, but there are instructures [here](https://github.com/chain/chain#dependencies), and maybe we can work together on that.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if err != nil {
+		log.Fatalln("s3 list objects error:", err)
+	}
+
+	var prefixedLocalKeys []string
+	for _, k := range localKeys {
+		prefixedLocalKeys = append(prefixedLocalKeys, path.Join("docs", k))
+	}
+	fmt.Println("keys to upload:", len(prefixedLocalKeys)) // TEMP
```
jeffomatic: `// TEMP` stuff can be removed, I think that's just my old debugging code.

cmd/docsync/main.go
dominic: The packages we're using all seem to be vendored: https://github.com/chain/chain/tree/main/vendor/github.com/aws/aws-sdk-go

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	localKeys := mustListContents(os.Args[2])
+
+	region := "us-east-1"
+	sess := session.Must(session.NewSession(aws.NewConfig().WithRegion(region)))
+	svc := s3.New(sess)
+
+	var remoteKeys []string
+	err := svc.ListObjectsPages(&s3.ListObjectsInput{
+		Bucket: aws.String(os.Args[1]),
+		Prefix: aws.String("docs/"),
```
jeffomatic: Some admittedly academic food for thought: what if the S3 prefix was also a param to the command? That would make the command more generic.

So it would be:

docsync bucket bucket-prefix local-dir

dominic: Yeah, I like that. We could upload only 1.2 if we really needed to at least.

cmd/docsync/main.go
jeffomatic: Yep I was wrong, thanks!

cmd/docsync/main.go
jeffomatic: Yeah exactly.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+func main() {
+	bucket := os.Args[1]
+	bucketPrefix := os.Args[2]
```
jeffomatic: I think it might be a good footgun prevention device to ensure the presence of a trailing slash here. I don't think it's ever valid for there not to be a trailing slash.

The tool I like for these situations is `strings.TrimPrefix`, e.g.:

    s = strings.TrimPrefix(s, "/") + "/"

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	sess := session.Must(session.NewSession(aws.NewConfig().WithRegion(region)))
+	svc := s3.New(sess)
+
+	var remoteKeys []string
+	err := svc.ListObjectsPages(&s3.ListObjectsInput{
+		Bucket: aws.String(bucket),
+		Prefix: aws.String(bucketPrefix),
+	}, func(page *s3.ListObjectsOutput, lastPage bool) bool {
+		for _, obj := range page.Contents {
+			if !(*obj.Key == "docs/") {
```
jeffomatic: Should this be `bucketPrefix` instead of `"docs/"`? Also, does the parent folder actually exist as its own S3 object?

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	})
+
+	if err != nil {
+		log.Fatalln("s3 list objects error:", err)
+	}
+
+	for _, k := range localKeys {
+		var body []byte
+
+		path := fmt.Sprintf("%s/%s", localDir, k)
```
jeffomatic: I think `path.Join` is the more idiomatic way of doing this.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+		contentType := mime.TypeByExtension(ext)
+
+		if contentType == "" {
+			upload.SetContentType("text/html")
+		} else {
+			upload.SetContentType(contentType)
+		}
+
+		_, err = svc.PutObject(upload)
+
```
jeffomatic: Usually we put error handling conditional blocks right underneath the function call that generates the error value, without a newline.

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+		if err != nil {
+			log.Fatalln(err.Error())
+		}
+	}
+
+	var prefixedLocalKeys []string
+	for _, k := range localKeys {
+		prefixedLocalKeys = append(prefixedLocalKeys, path.Join("docs", k))
```
jeffomatic: I think rather than using `path.Join` here, we should be prefixing with `bucketPrefix`. The setDiff is trying to see what is in the remote directory but not in the local directory, but in order for those things to be comparable, it needs to make the local keys (relative paths) look like remote keys ("absolute" paths with the prefix).

cmd/docsync/main.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, f := range files {
+		n := f.Name()
+
+		if f.IsDir() {
+			descendants := mustListContents(path.Join(parentPath, n))
+			for _, d := range descendants {
+				res = append(res, path.Join(n, d))
+			}
+		} else {
+			if !strings.HasPrefix(n, ".") {
```
jeffomatic: We should probably skip dot directories too, right? Maybe move this out of the `if/else` section here and do this instead:

```
if string.HasPrefix(n) {
  continue
}
```

cmd/docsync/main.go
dominic: assuming this means `TrimSuffix` but :+1:

cmd/docsync/main.go
dominic: It does yeah - it gets reported first in the list.

cmd/docsync/main.go
dominic: Hah. Yes.

jeffomatic:
Would you mind doing an informal benchmark of the new tool vs the old one on an upload to staging?

dominic:
A full upload? Sure.

dominic:
Unscientific coffeeshop benchmark:

```
OLD:
bin/upload-docs  196.55s user 54.15s system 96% cpu 4:18.92 total

NEW:
bin/upload-docs  144.91s user 44.27s system 55% cpu 5:42.38 total
```

PR #1282, opened 2017-06-06
sdk/node: create receiver specific test

chrisgarvin:


chrisgarvin:
image

sdk/node/test/receivers.js
```diff
@@ -0,0 +1,127 @@
...
+      signer.addKey(key, client.mockHsm.signerConnection)
+      return client.accounts.create({
+        alias: `account-${uuid.v4()}`,
+        rootXpubs: [key.xpub],
+        quorum: 1
+      })
+    })
+}
+
+function createAsset() {
```
dominic: I bet that future tests may want to use these too. No need to change it now, but think about extracting these to imported functions if the next set of tests use similar helpers.

sdk/node/test/receivers.js
```diff
@@ -0,0 +1,127 @@
...
+
+  describe('Pay to receiver', () => {
+    let testAccountAlias, testAssetAlias, testAccountReceiver
+
+    before(() => {
+      return createAccount()
+        .then((account) => testAccountAlias = account.alias )
+        .then(() => createAsset())
+        .then((asset) => testAssetAlias = asset.alias )
+        .then(() => client.accounts.createReceiver({ accountAlias: testAccountAlias }))
```
dominic: `createAccount` and `createAsset` could live pretty well in a `Promise.all` here too if you like that style.

PR #1283, opened 2017-06-06
database/raft: always update AppliedIndex

jbowens:
Always update the state's applied index, even if there is no change to
the state. Otherwise, triggerSnapshot will create a snapshot at an index
less than the true applied index. This can result in less than
`nSnapCatchupEntries` entries being available.

This is a fix for the 1.2 branch.

### DIFF
--- a/database/raft/raft.go
+++ b/database/raft/raft.go
@@ -736,8 +736,6 @@ func (sv *Service) applyEntry(ent raftpb.Entry, writers map[string]chan bool) {
 			defer sv.stateMu.Unlock()
 			defer sv.stateCond.Broadcast()
 			sv.state.RemovePeerAddr(cc.NodeID)
-		default:
-			panic(fmt.Errorf("unknown confchange type: %v", cc.Type))
 		}
 	case raftpb.EntryNormal:
 		//raft will send empty request defaulted to EntryNormal on leader election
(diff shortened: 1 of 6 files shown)


kr:
Is there a way to do this change without refactoring the unlock and broadcast calls, or are those all necessary? In other words, is there a version of this fix that's just a couple of lines even if it has some redundancy?

database/raft/raft.go
```diff
@@ -735,11 +736,16 @@ func (sv *Service) applyEntry(ent raftpb.Entry, writers map[string]chan bool) {
 			defer sv.stateMu.Unlock()
 			defer sv.stateCond.Broadcast()
 			sv.state.RemovePeerAddr(cc.NodeID)
+		default:
+			panic(fmt.Errorf("unknown confchange type: %v", cc.Type))
```
kr: This seems like a prudent thing to check for, but it's separate right? Sorry I just want to be extra careful with stable branch fixes.

Of course it should never hit this case, but this line might actually turn a silent failure into a crash, which might be the right thing to do but might actually be worse than silently ignoring the request? Either way I'd rather do it in a separate patch or not at all for this branch.

kr:
LGTM after one request

PR #1284, opened 2017-06-06
net/raft: always update AppliedIndex

jbowens:
Always update the state's applied index, even if there is no change to
the state. Otherwise, triggerSnapshot will create a snapshot at an index
less than the true applied index. This can result in less than
`nSnapCatchupEntries` entries being available.

Also, document that `nSnapCatchupEntries` must be <= `snapCount`.

We might want to consider making snapCount and nSnapCatchupEntries
fields on Service so that we can write snapshot tests without simulating
a 10,000 entries.

### DIFF
--- a/database/sinkdb/state.go
+++ b/database/sinkdb/state.go
@@ -34,6 +34,13 @@ func newState() *state {
 	}
 }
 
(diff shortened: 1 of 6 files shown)


kr:
LGTM

PR #1290, opened 2017-06-07
net/http/authn: provide error details

jbowens:
Provide error details when a request fails to authenticate. If no
credentials were provided, indicate that in the error detail. Otherwise,
provide details on each invalid credential.

Examples:
```
$ CORE_URL=https://mba.local:1999 corectl create-token jackson
RPC error: CH009 Request could not be authenticated
Detail: Invalid credentials: x509: certificate specifies an incompatible key usage
```

```
$ CORE_URL=http://jackattack:almondmilk@mba.local:1999 corectl create-token jackson
RPC error: CH009 Request could not be authenticated
Detail: Invalid credentials: invalid token: "jackattack"
```

```
$ CORE_URL=http://mba.local:1999 corectl create-token jackson
RPC error: CH009 Request could not be authenticated
Detail: No authentication credentials provided.
```

### DIFF
--- a/net/http/authn/authn.go
+++ b/net/http/authn/authn.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/hex"
+	"fmt"
 	"net"
 	"net/http"
 	"strings"
@@ -48,10 +49,17 @@ func NewAPI(tokens *accesstoken.CredentialStore, crosscorePrefix string, rootCAs
 // Authenticate returns the request, with added tokens and/or localhost
 // flags in the context, as appropriate.
 func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
-	ctx := certAuthn(req, a.rootCAs)
+	var authnErrors []string
+
+	ctx, err := certAuthn(req, a.rootCAs)
+	if err != nil {
+		authnErrors = append(authnErrors, err.Error())
+	}
 
 	token, err := a.tokenAuthn(req)
-	if err == nil && token != "" {
+	if err != nil {
+		authnErrors = append(authnErrors, err.Error())
+	} else if token != "" {
 		// if this request was successfully authenticated with a token, pass the token along
 		ctx = newContextWithToken(ctx, token)
 	}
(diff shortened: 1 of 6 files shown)


jbowens:
PTAL

jbowens:
PTAL

kr:
:eyes:

net/http/authn/authn.go
```diff
@@ -72,8 +80,14 @@ func (a *API) Authenticate(req *http.Request) (*http.Request, error) {
 
 	// if there is no authentication at all, we return an "unauthenticated" error,
 	// which may be helpful when debugging
-	if len(X509Certs(ctx)) < 1 && err != nil {
-		return req, errors.New("unauthenticated")
+	if len(X509Certs(ctx)) < 1 && Token(ctx) == "" {
+		err := errors.New("unauthenticated")
+		if len(authnErrors) > 0 {
+			err = errors.WithDetailf(err, "Invalid credentials: %s", strings.Join(authnErrors, "; "))
```
kr: I think this (and two lines down) is a case where https://github.com/golang/go/wiki/CodeReviewComments#error-strings applies, and we don't want to capitalize the messages here.

kr:
LGTM, with one nit.

net/http/authn/authn.go
jbowens: Hrm, these are error details though, so these are presented to the user in the json error response. They also _don't_ appear in the `Error` output. I'm not sure if we've adopted a style for error details or not.

kr: Oh, yeah, I was just going on the same principle as that comment, based on how it ultimately gets output:

```
Detail: Invalid credentials: invalid token: "jackattack"
```

vs

```
Detail: invalid credentials: invalid token: "jackattack"
```

It looks weird to have a capital letter in the middle of a sentence (after the first colon).

jbowens: Ah, I see. The "Detail: " prefix is added by corectl. Maybe we should get rid of those prefixes from corectl? 

```
RPC error: CH009 Request could not be authenticated
Detail: Invalid credentials: invalid token: "jackattack"
```
vs
```
CH009 Request could not be authenticated
Invalid credentials: invalid token: "jackattack"
```

jbowens: Seems like it's a bigger issue of how we want to present errors and what conventions should we use. There are quite a few places where the error details are complete sentences.

This is a little different than the situation described in that code-review-comment because the error detail is a specific value in our already structured error representation. There's no expectation that it will be modified with additional context. Maybe we should look at what we want exposed to the end user and work back from there.

kr: Good points. So it probably shouldn't block this PR. Never mind about my comment here.

PR #1291, opened 2017-06-08
sdk/node: add separate transaction feed test

jeffomatic:


sdk/node/test/transactionFeeds.js
```diff
@@ -0,0 +1,144 @@
+/* eslint-env mocha */
+
+const chain = require('../dist/index.js')
+const uuid = require('uuid')
+const assert = require('assert')
```
chrisgarvin: you can rid of lines 3-5, since they are unused

sdk/node/test/transactionFeeds.js
```diff
@@ -0,0 +1,144 @@
...
+  createAsset,
+  signer
+} = require('../testHelpers/util')
+
+const buildSignSubmit = buildFunc =>
+  client.transactions.build(buildFunc)
+    .then(tpl => signer.sign(tpl))
+    .then(tpl => client.transactions.submit(tpl))
+
+describe('Transaction feeds', () => {
```
chrisgarvin: Maybe switch to singular `Transaction feed` to stay consistent with the other test files

sdk/node/test/transactionFeeds.js
jeffomatic: Nice catch

sdk/node/test/transactionFeeds.js
jeffomatic: Will do, although I notice that the file name itself is plural, so we should pick one noun style for everything and stick with it.

PR #1296, opened 2017-06-09
sdk/node/test: add per-object tests for queryAll

jeffomatic:
Does not include tests for objects that don't have individual test files
yet:

- utxos
- balances

### DIFF
--- a/sdk/node/test/accessTokens.js
+++ b/sdk/node/test/accessTokens.js
@@ -64,6 +64,24 @@ describe('Access token', () => {
       .then(resp => expect(resp.items.map(item => item.id)).to.not.contain(tokenId))
   })
 
+  describe('queryAll', () => {
+    it('success example', () => {
+      let created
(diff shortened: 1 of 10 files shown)


sdk/node/test/accessTokens.js
```diff
@@ -64,6 +64,24 @@ describe('Access token', () => {
...
+  describe('queryAll', () => {
+    it('success example', () => {
+      let created
+      const queried = []
+
+      return createToken().then(token =>
+        created = token
+      ).then(() =>
+        client.accessTokens.queryAll({}, (token, next, done) => {
+          queried.push(token)
```
dominic: Would it be a cleaner test to push `token.id` here, and then have the expectation be:

```
expect(queried).to.include(created.id)
```

jeffomatic: SGTM

PR #1298, opened 2017-06-09
all: set TLS 1.2 as the minimum required version

dominic:


core/core_test.go
```diff
@@ -57,8 +57,8 @@ func TestForwardToLeader(t *testing.T) {
 		httpClient: &http.Client{
 			Transport: &http.Transport{
 				TLSClientConfig: &tls.Config{
-					RootCAs: certpool,
-				},
+					RootCAs:    certpool,
```
jeffomatic: Is this necessary for the tests to pass? Would we expect all clients to set this value in order to interact with Chain Core?

dominic: Per offline conversation, this is a simulated Chain Core API http object, not an external client object connecting to a Chain Core.

net/tls.go
```diff
@@ -0,0 +1,19 @@
...
+func DefaultTLSConfig() *tls.Config {
+	return &tls.Config{
+		// Avoids most of the memorably-named TLS attacks
+		MinVersion: tls.VersionTLS12,
+		// Causes servers to use Go's default ciphersuite preferences,
+		// which are tuned to avoid attacks. Does nothing on clients.
+		PreferServerCipherSuites: true,
+		// Only use curves which have constant-time implementations
+		CurvePreferences: []tls.CurveID{
+			tls.CurveP256,
```
tarcieri: I would suggest adding `tls.X25519` here, before `tls.CurveP256` (this was specifically recommended by @gtank despite what's on cryptopasta. Also Curve25519 rocks)

dominic: Oops sorry! I totally added that, and then replaced it with a copypasta from his `main`. Adding it now.

PR #1299, opened 2017-06-09
sdk/node/test: add buildSignSubmit helper to utils

jeffomatic:
This commit generalizes a test utility function for use in other test
files.

### DIFF
--- a/sdk/node/test/transactionFeeds.js
+++ b/sdk/node/test/transactionFeeds.js
@@ -9,7 +9,6 @@ const {
   client,
   createAccount,
   createAsset,
-  signer,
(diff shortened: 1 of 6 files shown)


sdk/node/test/transactionFeeds.js
```diff
@@ -9,14 +9,10 @@ const {
   client,
   createAccount,
   createAsset,
-  signer
+  signer,
```
dominic: `signer` isn't used in this file anymore.

jeffomatic: nice catch

PR #1300, opened 2017-06-09
sdk/node/test: add UTXO tests

jeffomatic:


sdk/node/test/unspentOutputs.js
```diff
@@ -0,0 +1,74 @@
...
+  let asset, account1, account2, tx1, tx2
+
+  before(() =>
+    Promise.all([
+      createAsset(),
+      createAccount(),
+      createAccount(),
+    ]).then(res =>
+      [asset, account1, account2] = res
+    ).then(() =>
```
dominic: This `then` isn't actually necessary.

PR #1303, opened 2017-06-10
sdk/node/test: remove legacy tests

jeffomatic:
Depends on #1301


jeffomatic:
Also depends on implementing callback versions of query calls.

PR #1307, opened 2017-06-12
sdk/node: move test helpers to compile directory

jeffomatic:
The test helpers are written using ES6 syntax, and should be
compiled alongside the tests. Using tests in CI builds (which use
an ES5 interpreter) will fail otherwise.


jeffomatic:
PTAL

PR #1310, opened 2017-06-12
database/raft: save snapshot position first

jbowens:
Save the snapshot position to the WAL before saving the snapshot data to
disk. On boot, we open the WAL at the most recently saved snapshot's
position. If we don't save the position to the WAL first, we might open
the WAL at a snapshot position that was never saved to the WAL.

This is a fix for the 1.2.x release line.

See coreos/etcd#8082, coreos/etcd#8088.


jbowens:
PTAL. This is the 1.2.x version of #1312.

PR #1311, opened 2017-06-12
net/tls.go: Prefer X25519 over NIST P-256

tonychain:
X25519 should be faster and less likely to experience timing sidechannels.
(See CVE-2016-7056 for a recent practical key recovery attack against P-256)


tonychain:
Would be good to get a 👍 from @gtank that this is the correct ordering

PR #1312, opened 2017-06-12
net/raft: save snapshot position first

jbowens:
Save the snapshot position to the WAL before saving the snapshot data to
disk. On boot, we open the WAL at the most recently saved snapshot's
position. If we don't save the position to the WAL first, we might open
the WAL at a snapshot position that was never saved to the WAL.

See coreos/etcd#8082, coreos/etcd#8088.

### DIFF
--- a/net/raft/raft.go
+++ b/net/raft/raft.go
@@ -98,6 +98,7 @@ type Service struct {
 	// it is ok to read without keeping startMu locked in
 	// code paths where Service is known to be initialized.
 	startMu  sync.Mutex
+	wal      *wal.WAL
 	raftNode raft.Node
 	id       uint64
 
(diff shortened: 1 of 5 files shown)


jbowens:
PTAL

CoreOS confirmed this in coreos/etcd#8088.

net/raft/raft.go
```diff
@@ -734,17 +738,17 @@ func (sv *Service) join(addr, baseURL string) (id uint64, walobj *wal.WAL, err e
...
-			return 0, nil, errors.Wrap(err)
-		}
 		err = walobj.SaveSnapshot(walpb.Snapshot{
 			Index: raftSnap.Metadata.Index,
 			Term:  raftSnap.Metadata.Term,
 		})
 		if err != nil {
 			return 0, nil, errors.Wrap(err)
 		}
+		err := sv.saveSnapshot(&raftSnap)
```
tessr: I wonder if `sv.saveSnapshot` and `wal.SaveSnapshot` could be consolidated somehow, to prevent us from accidentally doing these in the wrong order. (They always get executed together, right?)

PR #1315, opened 2017-06-12
all: set TLS 1.2 as the minimum required version

dominic:
1.1-compatible backport of #1298, #1311


jeffomatic:
Would it be alright if we separated this into 2 PRs? Future Dom troubleshooting something on release day will thank you.

dominic:
Present Jeff is totally correct :+1:

PR #1316, opened 2017-06-12
vendor: add google/snappy and introduce patches

tessr:
Add snappy, which is a build dependency for rocksdb. 

Per conversation with @ryandotsmith,  also create the `vendor/patches.md` file, which we will use to track small changes (i.e., changes not worth creating a fork for) to our vendored dependencies. In the future, we will likely have a more sophisticated patch system.


tessr:
PTAL

jbowens:
lgtm

PR #1317, opened 2017-06-12
wercker.yml: make branch compatible with node pipeline

jeffomatic:
Wercker requires a pipeline specification in `wercker.yml` (however trivial) to publish successful statuses to a PR. We are not backporting Node SDK updates to older branches (yet), but this will allow us to publish Node integration test statuses to PRs that merge into the 1.2-stable branch.


jeffomatic:
WIP

jeffomatic:
PTAL

PR #1325, opened 2017-06-13
bin/ccde-build-tags: new manifest for build tags

jeffomatic:
This file contains a list of build tags that should be used to
build Chain Core Developer Edition under the current branch. The
historical release branches (e.g. 1.2-stable, etc.) may have
different content for this file, depending on how build tags have
evolved.

This file will be used by the `bin/build-cored-release` command to
produce release-ready binaries built with the appropriate tags.


ryandotsmith:
`/bin` seems like an odd place to put this data...

jeffomatic:
`bin/build-cored-release` will be updated as follows:

```
buildTags=$(echo $(cat $newChain/cmd/cored/ccde-build-tags)) # echo removes newlines
go build -tags "$buildTags" ...
```

jeffomatic:
Gonna move the entire build command into its own script, rather than just export the build tags.

PR #1328, opened 2017-06-13
all: tick cored version to 1.2.1

dominic:


docs/core/reference/changelog.md
```diff
@@ -4,12 +4,26 @@ This changelog covers updates to cored, the Chain Core server binary.
...
 - [Java SDK](https://github.com/chain/chain/blob/1.2-stable/sdk/java/CHANGELOG.md)
 - [Node.js SDK](https://github.com/chain/chain/blob/1.2-stable/sdk/node/CHANGELOG.md)
 - [Ruby SDK](https://github.com/chain/chain/blob/1.2-stable/sdk/ruby/CHANGELOG.md)
 
+<a name="1.2.1"></a>
+## 1.2.1 (June 13, 2017)
+
+### Major changes
+
+TLS version 1.2 is now required for all HTTPS connections ([#1314](https://github.com/chain/chain/pull/1314)).
```
jeffomatic: Is there any guidance we can provide here on how to use TLS 1.2 if manual steps are necessary? It seems like at least Java 7 requires some kind of flag to be set for TLS 1.2 to be available on the client side, but I'm not 100% sure.

dominic: Good question. It's largely Java 6/7 that require it. Ruby and Node will either work, or they won't.

I can add more details :+1:

jeffomatic: This is really just a concern for clients. I didn't really look at the Java-related documentation very closely, so I wasn't sure if the settings were about setting TLS for servers, or if they were for enabling support for clients as well.

jeffomatic:
Would you mind starting a changelog in `docker/de`, and backfilling just for 1.2? We've been punting and I'm hoping it will be a small fix to get it in there.

docs/core/reference/changelog.md
```diff
@@ -4,12 +4,41 @@ This changelog covers updates to cored, the Chain Core server binary.
...
+
+For Java 7 applications, TLS 1.2 must be explicitly specified in the SSLContext
+object:
+
+```
+SSLContext context = SSLContext.getInstance("TLSv1.2");
+context.init(null,null,null);
```
jeffomatic: spaces after commas

dominic:
@jeffomatic No problem

PR #1329, opened 2017-06-13
bin: refactor developer edition build process

jeffomatic:
The build process now involves two commands:

**build-ccde**: generates cored and corectl with build config
appropriate for developer edition. Example:

    OUTPUT_DIR=foobar bin/build-ccde

(OUTPUT_DIR is optional.)

**build-historical-ccde**: checks out a user-provided gitref and
calls build-ccde. Uses a tempdir to avoid clobbering the
working tree. This tool is meant to be used in other build
automation scripts, such as the Docker or Mac builds. Example:

    OUTPUT_DIR=foobar bin/build-historical-ccde chain-core-server-1.1.0

Both of these scripts should be backported to the stable release
branches. For each release branch, build-ccde should be adjusted
so it uses the build tags and other build config options
appropriate for each historical release.


jeffomatic:
TL;DR: we need to think about this refactor a bit more.

Here's the gist of an IRL convo I had with @dominic:

1. There is no actual burning fire right now re: the release process. To build a deployable artifact (Docker, Windows, Mac), we check out the appropriate tag and run the appropriate build script on it. As far as we know, this should work for all stable release branches, and for `main`.
2. The only real problem we're dealing with right now is that it is impossible to fully satisfy the interface that `build-cored-release` claims to provide: to build any historical version of `cored` without forcing the user to perform a manual checkout. Since `build-cored-release` hard-codes the build tags, and since our build tags have changed over time, it's possible to produce binaries of historical versions that are compiled with the wrong (anachronistic) build tags. Even if we backport a tool like `build-ccde`, it will not affect gitrefs that cannot be updated, such as release tags.

jeffomatic:
I'm gonna close the PR and keep the branch around for a while, but there doesn't seem to be an immediate need for this.

If @kr is going to be dealing with the release process more formally (there is a larger mission statement for this), and there is no issue with our existing build process for current and historical branches, then we're better off not moving forward here.

Issue #1330, opened 2017-06-13
net/raft: panic joining existing cluster


jbowens:
Steps I followed:

1. Start `LISTEN=localhost:1999 cored`. 
2. Run `corectl init` to initialize the raft cluster.
3. Run `corectl config-generator` to configure Chain Core as a generator.
4. Run `corectl allow-address localhost:1998` to prepare for a second cored instance.
5. Start `LISTEN=localhost:1998 CHAIN_CORE_HOME=~/.chaincore2 cored`
6. Run `CORE_URL=https://localhost:1998 corectl join`
8. Stop the localhost:1999 cored.
9. Restart the localhost:1998 cored.

Stack trace:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x17a6200]

goroutine 23 [running]:
chain/vendor/github.com/coreos/etcd/raft.(*raft).appendEntry(0xc4202980f0, 0xc420075c70, 0x1, 0x1)
    /Users/jackson/src/chain/vendor/github.com/coreos/etcd/raft/raft.go:520 +0x220
chain/vendor/github.com/coreos/etcd/raft.(*raft).becomeLeader(0xc4202980f0)
    /Users/jackson/src/chain/vendor/github.com/coreos/etcd/raft/raft.go:620 +0x431
chain/vendor/github.com/coreos/etcd/raft.(*raft).campaign(0xc4202980f0, 0x1bb1ba7, 0x10)
    /Users/jackson/src/chain/vendor/github.com/coreos/etcd/raft/raft.go:643 +0x959
chain/vendor/github.com/coreos/etcd/raft.(*raft).Step(0xc4202980f0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
    /Users/jackson/src/chain/vendor/github.com/coreos/etcd/raft/raft.go:753 +0x1510
chain/vendor/github.com/coreos/etcd/raft.(*node).run(0xc420156fc0, 0xc4202980f0)
    /Users/jackson/src/chain/vendor/github.com/coreos/etcd/raft/node.go:323 +0x64f
created by chain/vendor/github.com/coreos/etcd/raft.RestartNode
    /Users/jackson/src/chain/vendor/github.com/coreos/etcd/raft/node.go:223 +0x390
```

Here's a zip of the Chain Core data directory. The same panic is hit every restart. 
[...]

PR #1332, opened 2017-06-14
net/raft: wait for ConfChange to be applied

jbowens:
When adding a new node to the cluster, wait for the conf change to be
committed and applied before taking a state snapshot and responding
to the new node.

Previously, Join had a race condition between the application of the
conf change entry and the taking of the snapshot. If the snapshot was
taken before the conf change was committed or applied, the new node
would try to boot its state machine from a state that did not include
itself in the configuration. It could mistakeningly think that it is a
single-node cluster and try to elect itself, panicking when its node
id doesn't exist in the progress list: See issue #1330.

Waiting for the conf change to be applied ensures that:
* The /raft/join endpoint only returns if adding the node to the cluster
  was committed.
* The snapshot returned from the /raft/join endpoint includes the new
  node in its configuration.

An alternative we might want to consider later: Return the peer list
from /raft/join and pass it to StartNode along with the new node ID
appended. This would remove the requirement of booting from a very fresh
snapshot. This is how etcd boots new nodes:
https://github.com/coreos/etcd/blob/750dc7f157c2d13e9b7d229ad749c1a4d62b0def/etcdserver/raft.go#L394-L400

### DIFF
--- a/generated/rev/RevId.java
+++ b/generated/rev/RevId.java
@@ -1,4 +1,4 @@
 
 public final class RevId {
-	public final String Id = "main/rev3242";
+	public final String Id = "main/rev3243";
 }
--- a/generated/rev/revid.go
+++ b/generated/rev/revid.go
@@ -1,3 +1,3 @@
 package rev
 
-const ID string = "main/rev3242"
+const ID string = "main/rev3243"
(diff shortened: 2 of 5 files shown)


tessr:
Wow, nice.

PR #1335, opened 2017-06-14
database/raft: wait for ConfChange to be applied

jbowens:
When adding a new node to the cluster, wait for the conf change to be
committed and applied before taking a state snapshot and responding
to the new node.

Previously, Join had a race condition between the application of the
conf change entry and the taking of the snapshot. If the snapshot was
taken before the conf change was committed or applied, the new node
would try to boot its state machine from a state that did not include
itself in the configuration. It could mistakeningly think that it is a
single-node cluster and try to elect itself, panicking when its node
id doesn't exist in the progress list: See issue #1330.

Waiting for the conf change to be applied ensures that:
* The /raft/join endpoint only returns if adding the node to the cluster
  was committed.
* The snapshot returned from the /raft/join endpoint includes the new
  node in its configuration.

This is a backport fix for the 1.2.x release line.

### DIFF
--- a/database/raft/raft.go
+++ b/database/raft/raft.go
@@ -535,6 +535,16 @@ func (sv *Service) wait(index uint64) bool {
 	return sv.state.AppliedIndex() >= index //if false killed b/c of done signal
 }
 
+// waitForNode waits until the provided nodeID is committed into
+// the cluster peer list.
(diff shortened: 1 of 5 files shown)


jbowens:
PTAL, backport of #1332 onto 1.2-stable

tessr:
LGTM

PR #1336, opened 2017-06-14
net/raft: add unit test for cluster setup

jbowens:
Add a simple go test for creating a new three-node cluster. In Chain
Core, the database/sinkdb package provides the implementation to raft's
State interface. For this internal test (and future tests), this adds a
simple State key-value store implementation.

This new test takes test statement coverage in net/raft from 14.6% to
63.9%.

### DIFF
--- a/net/raft/raft_test.go
+++ b/net/raft/raft_test.go
@@ -2,11 +2,15 @@ package raft
 
 import (
 	"bytes"
+	"chain/net"
 	"context"
+	"crypto/x509"
 	"io/ioutil"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 )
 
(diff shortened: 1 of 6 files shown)


net/raft/raft_test.go
```diff
@@ -116,3 +120,119 @@ func TestStartUninitialized(t *testing.T) {
...
+	}
+
+	// Create the raft service, passing in the server's address
+	node.service, err = Start(node.addr, node.dir, httpClient, node.state)
+	if err != nil {
+		node.cleanup()
+		t.Fatal(err)
+	}
+
+	node.wg.Done()
```
tessr: Apologies in advance for a really foggy review comment, but: 

You're basically just using the wait group to ensure that we don't try to serve http before the node is actually ready, right? Isn't a wait group really for blocking on a whole group of goroutines? Is there something simpler that we could use instead? (Maybe like a ready channel?)

net/raft/raft_test.go
```diff
@@ -116,3 +120,119 @@ func TestStartUninitialized(t *testing.T) {
...
+	service *Service
+}
+
+func (n *testNode) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	n.wg.Wait()
+	n.service.ServeHTTP(rw, req)
+}
+
+func (n *testNode) cleanup() {
+	os.RemoveAll(n.dir)
```
tessr: probably should handle the error somehow

jbowens: I don't think there's anything we can do here if cleaning up the test fails. It's also okay to be best effort because newTestNode uses a temporary directory which the OS will eventually cleanup.

tessr: I think it could be useful to log it, mostly because I think totally silent errors lead to mysteries, but I don't feel strongly about it.

net/raft/raft_test.go
jbowens: Yeah. It's definitely possible to use a channel here. I'm not sure what the correct idiom is here, but I usually think of channels as the right tool when there's data that you want to share between two goroutines or you need to be able to `select { ... }` over multiple asynchronous events, neither is the case here.

I actually copied this use of wait group after the [same thing we do in cored](https://github.com/chain/chain/blob/main/cmd/cored/main.go#L506).

Honestly, maybe I should just drop the synchronization because the node's address hasn't escaped yet, so there shouldn't be any incoming requests. That would change if we support restarting an existing node here, but we can handle it then.

net/raft/raft_test.go
jbowens: FWIW, it [seems](https://github.com/golang/go/blob/2c1bff6e0681ff2cace0f69c28b05f47d4262aa3/src/io/ioutil/tempfile_test.go#L19) [pretty](https://github.com/golang/go/blob/994b7eebc7e704b7132e5a69be72ca885b30b83e/misc/cgo/testcarchive/carchive_test.go#L200) [common](https://github.com/golang/go/blob/79d05e75ca38f0e4b9986bbba4cd56398428f9fb/test/run.go#L526) to defer os.RemoveAll() without an error check in the standard library.

tessr: Ok.

net/raft/raft_test.go
tessr: Yeah. Hmm. Alright, that seems fine. (Again, I'm genuinely not sure what the best pattern is here.)

net/raft/raft_test.go
```diff
@@ -116,3 +119,115 @@ func TestStartUninitialized(t *testing.T) {
...
+	defer nodeB.cleanup()
+	nodeC := newTestNode(t)
+	defer nodeC.cleanup()
+
+	// Initialize A, creating a fresh cluster.
+	must(t, nodeA.service.Init())
+
+	// Update the cluster to allow A, B and C's addresses.
+	var err error
+	_, err = nodeA.service.Exec(ctx, set("/allowed/"+nodeA.addr, "yes"))
```
tessr: are we really using "yes" as the value here? (I guess it doesn't matter)

net/raft/raft_test.go
```diff
@@ -116,3 +119,115 @@ func TestStartUninitialized(t *testing.T) {
...
+}
+
+func (n *testNode) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	n.service.ServeHTTP(rw, req)
+}
+
+func (n *testNode) cleanup() {
+	os.RemoveAll(n.dir)
+	n.server.Close()
+	// TODO(jackson): stop the Service too
```
tessr: This TODO can happen now, right?

net/raft/raft_test.go
jbowens: ¯\\\_(ツ)\_/¯, it's for the test state implementation, and it uses the presence of the key not the value to determine if an address is allowed. So it can be anything.

PR #1342, opened 2017-06-15
database/sinkdb: add Close method

jbowens:
Add Close method to release resources associated with a sinkdb handle
and stop its raft service. Also use this Close method to write a basic unit
test for persistence.

Ideally, we could unit test without gracefully closing the database and the
raft WAL, but the WAL file lock won't be released until the process exits
and the file descriptors are released.

### DIFF
--- a/database/sinkdb/sinkdb.go
+++ b/database/sinkdb/sinkdb.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"net/http"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/golang/protobuf/proto"
@@ -31,6 +32,9 @@ func Open(laddr, dir string, httpClient *http.Client) (*DB, error) {
 
 // DB provides access to an opened kv store.
 type DB struct {
+	mu     sync.Mutex
+	closed bool
+
 	state *state
 	raft  *raft.Service
 }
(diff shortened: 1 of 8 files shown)


database/sinkdb/sinkdb.go
```diff
@@ -46,6 +50,21 @@ func (db *DB) Ping() error {
 	return err
 }
 
+// Close closes the database handle releasing its resources. It is
+// the caller's responsibility to ensure that there are no concurrent
+// database operations in flight. Close is idempotent.
+//
+// All other methods have undefined behavior on a closed DB.
```
jbowens: We could, but that would only get us part of the way there. The DB could be closed after another method is invoked but before it completes. We'd have to either hold on to the mutex during the call (and allow a single operation at a time) or track the number of inflight operations and in Close wait for all operations to cease before closing. Practically, since this is just intended to make it unit testable, IMO it's better to leave it undefined and make it the caller's responsibility to ensure they don't use a closed DB.

tessr: Okay. I'm alright with that.

net/raft/raft.go
```diff
@@ -351,6 +357,28 @@ func (sv *Service) Join(bootURL string) error {
...
+	close(sv.stop)
+
+	// if the Service was never initialized, there are no
+	// goroutines to wait for and no open wal
+	if !sv.initialized() {
+		return nil
+	}
+
+	// wait for the service's goroutines to stop
+	sv.stopwg.Wait()
```
jbowens: If you want Close to wait for it, yep. We only spawn long-running goroutines in one place: sv.startLocked.

tessr: ok, cool - can you add a comment somewhere (maybe on the stopwg field) that explains this?

PR #1344, opened 2017-06-15
create makefile

tessr:
This is a simple proposal for the beginning of a makefile to build Chain Core.

I anticipate this may turn into a real 

but we need a simple way to build the new C dependencies for RocksDB.


tessr:
PTAL 🙈

Makefile
```diff
@@ -0,0 +1,13 @@
+## builds chain core with c dependencies, for development environments
+build-dev: build-lib
+	go install -tags 'localhost_auth http_ok reset' chain/cmd/cored
+
+## builds statically-linked c dependencies
+build-lib:
+	cd $(CHAIN)/vendor/github.com/google/snappy; mkdir build; cd build && cmake ../ && make
```
bobg: Add a definition at the top of the Makefile: `SNAPPY = $(CHAIN)/vendor/github.com/google/snappy`. Then split this up into multiple lines:
```
    -mkdir $(SNAPPY)/build
    cd $(SNAPPY)/build && cmake ..
    cd $(SNAPPY)/build && $(MAKE)
```

(The leading hyphen means "ignore errors." The predefined variable `MAKE` does magic when doing recursive makes, see https://www.gnu.org/software/make/manual/make.html#MAKE-Variable.)

Makefile
```diff
@@ -0,0 +1,13 @@
+## builds chain core with c dependencies, for development environments
+build-dev: build-lib
```
bobg: Do you want `make` (with no args) to invoke this rule? If not, add a default rule before it, something like:
```
null:
    @echo Please specify a make target.
```

tessr:
Thank you for the helpful feedback, @bobg! PTAL 😁

Makefile
```diff
@@ -0,0 +1,21 @@
+SNAPPY = $(CHAIN)/vendor/github.com/google/snappy
+ROCKSDB = $(CHAIN)/vendor/github.com/facebook/rocksdb
+
+null:
+	@echo Please specify a make target.
+
+## builds chain core with c dependencies, for development environments
+build-dev: build-lib
+	go install -tags 'localhost_auth http_ok reset' chain/cmd/cored
```
bobg: Just in case the user has the `chain` repo checked out under some name other than `chain`, it might be better to do
```
cd $(CHAIN) && go install -tags 'localhost_auth http_ok reset' ./cmd/cored
```

Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+## builds statically-linked c dependencies
+build-lib:
+	-mkdir $(SNAPPY)/build
+	cd $(SNAPPY)/build && cmake ..
+	cd $(SNAPPY)/build && $(MAKE)
+	cd $(ROCKSDB); $(MAKE) static_lib
```
bobg: Recommend `&&` instead of `;`, so that if the `cd` fails the `$(MAKE)` is not attempted.

Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+## clean removes all data directories and c libraries
+clean:
+	rm -rf $(SNAPPY)/build
+	rm $(ROCKSDB)/librocksdb.a
```
bobg: Use the `-` prefix here, or `rm -f`. (Either one will cause errors to be ignored.)

Makefile
tessr: That seems like a real edge case, but I don't see any harm in protecting against it!

tessr:
Ready for another pass! ⛄️

Makefile
```diff
@@ -0,0 +1,21 @@
...
+build-lib:
+	-mkdir $(SNAPPY)/build
+	cd $(SNAPPY)/build && cmake ..
+	cd $(SNAPPY)/build && $(MAKE)
+	cd $(ROCKSDB) && $(MAKE) static_lib
+
+## clean removes all data directories and c libraries
+clean:
+	rm -rf $(SNAPPY)/build
+	rm -f $(ROCKSDB)/librocksdb.a
```
bobg: Nit: would like a trailing newline.

Makefile
```diff
@@ -0,0 +1,21 @@
...
+null:
+	@echo Please specify a make target.
+
+## builds chain core with c dependencies, for development environments
+build-dev: build-lib
+	cd $(CHAIN) && go install -tags 'localhost_auth http_ok reset' ./cmd/cored
+
+## builds statically-linked c dependencies
+build-lib:
+	-mkdir $(SNAPPY)/build
```
bobg: Looks like `$(SNAPPY)/build` is part of the github repo. Does the build process require that we use it, or could we choose a more-obviously-temporary name for the build directory? It would be nice not to create git diffs with every run of `make build-lib` or `make clean`.

Maybe the right thing is to remove that `build` subdir from the repo? (Looks like it, because its `CMakeCache.txt` includes directory names under `/Users/tessr`.)

tessr: Erp, I think I checked the build directory in by accident. I'll remove that.

Makefile
tessr: man, for some reason, I thought I explicitly added a newline earlier. I don't know where it went! 

(added.)

tessr:
🔴 -y for 🅰️ nother pass

bobg:
LGTM, but please do update the top-level readme as suggested earlier (whether in this PR or a followup one)

tessr:
Oh yes, will do in this PR. Thanks for the reminder.

PR #1353, opened 2017-06-17
dashboard: poll for core connection before determining authn status

dominic:


dashboard/src/features/authn/actions.js
```diff
@@ -9,22 +9,28 @@ export const logIn = (token) => (dispatch) => {
...
         }
 
         if (err.status == 401) {
           dispatch({type: 'AUTHENTICATION_REQUIRED'})
         }
       })
-      .then(() => dispatch(fetchCoreInfo()))
-      .then(() => dispatch({type: 'AUTHENTICATION_READY'}))
+      .then(() => connectedToCore && dispatch(fetchCoreInfo()))
+      .then(() => connectedToCore && dispatch({type: 'AUTHENTICATION_READY'}))
```
jeffomatic: This seems like a race condition between `fetchCoreInfo` and any other action that can update the flash message. This may or may not be an issue with other flash messages as well, but I just wanted to double-check.

dominic: There's probably some light racing that could happen. Polling to `/info` shouldn't happen before `AUTHENTICATION_READY` is dispatched. Since this really only happens on full page reloads, the race condition I can imagine is:

- Open a new browser tab
- Navigate to a page that will throw an exception e.g. http://localhost:1999/dashboard/transactions?filter=%3D
- While fetching authentication status, the loadPage callback returns with an error from `/list-transactions`
- The authentication status checks finish and the page is displayed
- The error flash message from the third step is destroyed

If I introduce an artificial slowdown into the `getAuthenticationStatus` call I can make this happen.

However, this helps in the other direction: if the client is having trouble connecting to the core or the core is offline, it will poll `/info` until it gets a success. In the meantime, the `onEntry` call to `/list-transactions` has started and failed with an error (and the page doesn't load because the call already failed).

So...the correct fix here is to have child routes wait for parent `onEntry` and remove any potential race. This is a pretty artificial failure though; dashboard is almost always run from within cored, and the race condition is not present if the core is killed and restarted while an existing tab is open.

PR #1357, opened 2017-06-19
docker/ci: Update Rust to nightly-2017-06-15

tonychain:
This includes some pretty major upstream changes that affect us: libcollections was merged into liballoc, and that's where we get all of our heap-allocated data structures.

Also bumps rustfmt to 0.9.0, which includes a number of changes to the rustfmt style guide.

Lastly, bumps clippy to 0.0.140, which is confirmed to build on this nightly release.

### DIFF
--- a/docker/ci/install/install-protobuf
+++ b/docker/ci/install/install-protobuf
@@ -2,11 +2,19 @@
 
 set -e
 
+: ${GOPATH:?must be set}
+
+# Install protobuf
 curl -LO https://s3.amazonaws.com/chain-qa/protobuf-3.1.0.tar.gz
 tar xvzf protobuf-3.1.0.tar.gz
 chmod +x protobuf-3.1.0/protoc
 cp protobuf-3.1.0/protoc /usr/local/bin
 cp -a protobuf-3.1.0/libproto* /usr/local/lib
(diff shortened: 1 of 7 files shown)


docker/ci/install/install-protobuf
```diff
@@ -2,11 +2,19 @@
...
 cp protobuf-3.1.0/protoc /usr/local/bin
 cp -a protobuf-3.1.0/libproto* /usr/local/lib
 rm -rf protobuf-3.1.0 protobuf-3.1.0.tar.gz
-go get -u github.com/golang/protobuf/proto
-go get -u github.com/golang/protobuf/protoc-gen-go
+
+# Install protoc-gen-go
+git clone https://github.com/golang/protobuf $(go env GOPATH)/src/github.com/golang/protobuf/
+cd $(go env GOPATH)/src/github.com/golang/protobuf/
+git reset --hard 18c9bb3261723cd5401db4d0c9fbc5c3b6c70fe8 # pin to this golang/protobuf commit hash
```
tonychain: FYI @tessr, here's what I did to pin protobuf to a specific version.

Perhaps we should be using the vendored versions of these libraries rather than grabbing the code off GitHub (I have a similar problem with gas). But that's a larger yak than I'm presently prepared to shave...

PR #1360, opened 2017-06-19
Makefile: add convenience tasks for development

jeffomatic:
This commit adds convenience Make tasks for rapid iteration:

- `make`/`make run`: Run Chain Core
- `make resetdb`: Delete contents of development databases in both Postgres and the Chain Core data directory
- `make rerun`: Delete contents of development databases, configure the local core as a generator, and run Chain Core
- `make dashserve`: Run the development dashboard at http://localhost:3000
- `make docserve`: Run the development documentation server at http://localhost:8080

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,32 @@
 SNAPPY = $(CHAIN)/vendor/github.com/google/snappy
 ROCKSDB = $(CHAIN)/vendor/github.com/facebook/rocksdb
+DB_DEV = core
+RAFT_DEV = $(HOME)/.chaincore/raft
 
-null:
-	@echo Please specify a make target.
+default: run
+
+## run a development version of Chain Core at http://localhost:1999
+run:
+	cored
+
+## reset the development database
+resetdb:
+	-dropdb $(DB_DEV)
+	createdb $(DB_DEV)
+	rm -rf $(RAFT_DEV)
+
+## delete the development database, re-configure core, and run server
+rerun: build-dev corectl resetdb
+	sh -c "corectl wait; corectl config-generator" &
+	cored
+
+## run development dashboard at http://localhost:3000
+dashserver:
+	npm --prefix dashboard start
+
+## run development documentation server at http://localhost:8080
+docserver: md2html
+	md2html serve
 
 ## builds chain core with c dependencies, for development environments
 build-dev: build-lib
@@ -19,3 +43,9 @@ build-lib:
 clean:
 	rm -rf $(SNAPPY)/build
 	rm -f $(ROCKSDB)/librocksdb.a
+
+corectl:
+	go install chain/cmd/corectl
+
+md2html:
+	go install chain/cmd/md2html
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,11 @@ rerun: build-dev corectl resetdb
 	cored
 
 ## run development dashboard at http://localhost:3000
-dashserver:
+dashserve:
 	npm --prefix dashboard start
 
 ## run development documentation server at http://localhost:8080
-docserver: md2html
+docserve: md2html
 	md2html serve
 
 ## builds chain core with c dependencies, for development environments
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ RAFT_DEV = $(HOME)/.chaincore/raft
 default: run
 
 ## run a development version of Chain Core at http://localhost:1999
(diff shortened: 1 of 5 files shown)


jeffomatic:
PTAL. This is a snapshot of what product team uses for developing against Chain Core, curious to see if others will find this useful.

Makefile
```diff
@@ -1,8 +1,32 @@
...
+	createdb $(DB_DEV)
+	rm -rf $(RAFT_DEV)
+
+## delete the development database, re-configure core, and run server
+rerun: build-dev corectl resetdb
+	sh -c "corectl wait; corectl config-generator" &
+	cored
+
+## run development dashboard at http://localhost:3000
+dashserver:
```
bobg: WDYT of using verbs for targets that run things, and nouns for verbs that only build/install things? E.g. `run-dashserver`. (I'd also argue for `run-cored` over simply `run`, in the interest of explicitness.)

Makefile
```diff
@@ -1,8 +1,32 @@
 SNAPPY = $(CHAIN)/vendor/github.com/google/snappy
 ROCKSDB = $(CHAIN)/vendor/github.com/facebook/rocksdb
+DB_DEV = core
+RAFT_DEV = $(HOME)/.chaincore/raft
 
-null:
-	@echo Please specify a make target.
+default: run
```
bobg: Can we stick with requiring explicit make targets?

Makefile
```diff
@@ -1,8 +1,32 @@
...
+
+## run a development version of Chain Core at http://localhost:1999
+run:
+	cored
+
+## reset the development database
+resetdb:
```
tessr: should this drop the raft directory too?

Makefile
jeffomatic: At risk of opening a bikeshed, what would your opposition to having a sensible default be? Building and running cored is one of the most common things I do in an edit/save/compile/test cycle.

Makefile
jeffomatic: I agree in principle, but I am informally optimizing for less typing. In practice, with the current setup, you can save a few keystrokes, i.e.:

- `make` vs `make run-c<TAB>`
- `make da<TAB>` vs `make run-da<TAB>`
- `make do<TAB>` vs `make run-do<TAB>`

This seems like small potatoes but it really adds up when you're going back and forth between code and the console.

Makefile
jeffomatic: Not sure if you're suggesting that it should or shouldn't, but to clarify, the third line in this task drops the Raft file storage. The intent of this task to is to delete anything that constitutes Chain Core's "database" so you get a clean slate the next time you run Chain Core.

Makefile
bobg: Eh, none really. Withdrawn.

Makefile
jeffomatic: Would it hurt your head if we compromised and did a portmanteau, e.g. `dashserve` and `docserve`?

bobg: Good compromise, I like it!

Makefile
tessr: OOPS. Yes. That's what I wanted. Somehow I missed that last line. 😞

jeffomatic:
PTAL again

Makefile
```diff
@@ -1,8 +1,32 @@
 SNAPPY = $(CHAIN)/vendor/github.com/google/snappy
 ROCKSDB = $(CHAIN)/vendor/github.com/facebook/rocksdb
+DB_DEV = core
+RAFT_DEV = $(HOME)/.chaincore/raft
```
tessr: This will only work if CHAIN_CORE_HOME is unset or matches ~/.chaincore. (See https://github.com/chain/chain/blob/main/core/env.go)

I am not too worried about this kind of thing, but can we add a comment at the top of this file explaining that this is for build/development purposes and/or may not work in all environments, or something like that?

bobg:
LGTM

tessr:
jinx!

Makefile
jeffomatic: @tessr Nice catch, I was a little caught up on that myself. I think I can do one better and read CHAIN_CORE_HOME if available:

```
RAFT_DEV = $(or $(CHAIN_CORE_HOME), $(HOME)/.chaincore)/raft
```

tessr: Oh, nice!

PR #1366, opened 2017-06-21
net/raft: update Makefile to run raft cluster

vickiniu:


net/raft/Makefile
```diff
@@ -1,28 +1,19 @@
...
 # our Raft packages. 
 
+
 build:
-	go install  -tags 'localhost_auth http_ok' chain/cmd/cored
+	go install -tags localhost_auth chain/cmd/cored
+	go install chain/cmd/corectl
 	dropdb --if-exists core0
-	rm -rf $(HOME)/.chaincore{0,1,2}
+	rm -rf $(HOME)/.chaincore?*
```
tessr: Let's have this _just_ delete the raft directory, so that if we have certs in `.chaincore*` they won't get deleted.

tessr:
Awesome, LGTM!

net/raft/dev/run-bootstrapped-node.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Enables job control to allow backgrounding & foregrounding cored process
+set -m
```
kr: I think we also want `set -e` in these scripts. And maybe also `set -o pipefail` for good measure.

PR #1369, opened 2017-06-22
dashboard: surface errors generated by preprocessTransaction step

jeffomatic:


dashboard/src/features/transactions/actions.js
```diff
@@ -70,16 +70,39 @@ function getTemplateXpubs(tpl) {
 }
 
 form.submitForm = (formParams) => function(dispatch) {
-  const buildPromise = chainClient().transactions.build(builder => {
-    const processed = preprocessTransaction(formParams)
+  try {
+    const buildPromise = chainClient().transactions.build(builder => {
+      const processed = preprocessTransaction(formParams)
```
dominic: If `preprocessTransaction` is the issue, we don't need to wrap our existing Promise-returning function in a giant try catch block. Can this just wrap line 74 instead?

jeffomatic: Yes. I could go either way. The larger try/catch lets us catch any other unexpected error that may get thrown by any other code, since it will not be surfaced anywhere unless we perform a deliberate catch.

I think this is a problem with error handling that uses dispatch for resolution--any thrown errors (i.e. not promise rejections) are going to fail silently unless we have some explicit failsafe. If I could find one place to add that failsafe, that would obviously be the best possible solution for this PR.

Separately, I wonder if we should consider having the SDK treat the client's builder function as async, and returning any thrown errors as promise rejections. We're not doing that right now and that's partially why we have this bug.

dominic: I'm liking the last option best - a throw in the builder function should result in a promise rejection. That covers any future issues really nicely.

jeffomatic:
Fixed in #1376

Issue #1334, opened 2017-06-14
net/raft: advance ticks on start


jbowens:
To ensure a quick start of a new cluster, currently we wait for the first Ready synchronously before returning (added in #1242). Instead, it might be cleaner/less error prone to artificially advance the state machine to right before an election. This is what etcd does: https://github.com/coreos/etcd/blob/750dc7f157c2d13e9b7d229ad749c1a4d62b0def/etcdserver/raft.go#L371-L379

jbowens:
Fixed in baa06cc.

PR #1370, opened 2017-06-22
net/raft: advance ticks on cluster init

vickiniu:
Advancing ticks until right before election to speed up tests when initializing a raft cluster (see https://github.com/coreos/etcd/pull/3184) . Changes the approach from #1242 to avoid errors from artificially changing the state machine.

There is some randomness in test timing now, since electionTimeout is randomized in the Raft algorithm (up to +/- 0.4s)

Addresses #1334


jbowens:
lgtm

PR #1373, opened 2017-06-23
cmd/cored: make unconfigured cores poll for config

### LINKED ISSUE #1371: core/config: unconfigured cores don't know when they've been configured
If you run three processes (listening on, say, 1999, 1998 and 1997) on an unconfigured core, and then configure the core via the process listening on 1999, process 1999 will exec itself but processes 1998 and 1997 will not. In fact, they won't realize that they are now part of a configured core until they get restarted, even though this change is reflected in sinkDB. 

We have two options for resolving this:

1. We could have unconfigured core processes poll `sinkDB://core/config`* until something appears at that key, at which point they exec themselves.
2. We could block processes from joining a raft cluster until that cluster contains configuration state.

Polling kind of sucks, but the ux for the first solution seems strictly better to me. Also, polling shouldn't be that big of a deal because, seriously, what else is an unconfigured core going to do other than sit around polling sinkDB? (We may need to be careful about using stale vs linearizable reads when polling sinkDB.) 

_* totally just made this notation up_

vickiniu:
Adds goroutine for unconfigured cores to check for an updated config from sinkDB and exec themselves to update.

Also includes small fix to make sure raft run scripts don't break if chaincore directories exist already

Fixes #1371

### DIFF
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -292,15 +292,9 @@ func main() {
 		h = core.RunUnconfigured(ctx, db, sdb, *listenAddr, opts...)
 
 		go func() {
-			ticker := time.Tick(5 * time.Second)
 			for {
-				select {
-				case <-ticker:
-					if conf != nil {
-						return
-					}
-					core.CheckConfigMaybeExec(ctx, sdb, *listenAddr)
-				}
+				core.CheckConfigMaybeExec(ctx, sdb, *listenAddr)
+				time.Sleep(5 * time.Second)
 			}
 		}()
 	}
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -296,7 +296,6 @@ func main() {
 			for {
 				select {
 				case <-ticker:
-					log.Println("Tick!")
 					if conf != nil {
 						return
 					}
--- a/cmd/cored/main.go
+++ b/cmd/cored/main.go
@@ -290,6 +290,20 @@ func main() {
 		opts = append(opts, enableMockHSM(db)...)
 		chainlog.Printf(ctx, "Launching as unconfigured Core.")
 		h = core.RunUnconfigured(ctx, db, sdb, *listenAddr, opts...)
+
+		go func() {
+			ticker := time.Tick(5 * time.Second)
+			for {
+				select {
+				case <-ticker:
+					log.Println("Tick!")
+					if conf != nil {
+						return
+					}
+					core.CheckConfigMaybeExec(ctx, sdb, *listenAddr)
+				}
+			}
+		}()
 	}
 	coreHandler.Set(h)
 	chainlog.Printf(ctx, "Chain Core online and listening at %s", *listenAddr)
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -162,7 +162,7 @@ func deleteFromPG(ctx context.Context, db pg.DB) error {
 
 // Loads config status from sinkdb to see if other raft nodes have configured
 // so that uncofigured nodes can update.
-func UpdateConfigStatus(ctx context.Context, sdb *sinkdb.DB) (*Config, error) {
+func CheckConfigExists(ctx context.Context, sdb *sinkdb.DB) (*Config, error) {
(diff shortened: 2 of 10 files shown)


tessr:
Two meta-comments:

1. Can you update the title of this PR to be a little more descriptive? Perhaps `cmd/cored: make unconfigured cores poll for config`? 
2. Can you also note in the PR body that you also updated the raft run scripts to pass even if the directory exists already?

core/config/config.go
```diff
@@ -160,6 +160,19 @@ func deleteFromPG(ctx context.Context, db pg.DB) error {
 	return errors.Wrap(err, "deleting config stored in postgres")
 }
 
+// Loads config status from sinkdb to see if other raft nodes have configured
+// so that uncofigured nodes can update.
+func UpdateConfigStatus(ctx context.Context, sdb *sinkdb.DB) (*Config, error) {
```
tessr: This sounds like we're actually updating the config status, which we're not--just polling to see if the config exists. Maybe let's rename it to `CheckConfigExists`? (That's more consistent with the caller's name, too.)

core/config/config.go
```diff
@@ -160,6 +160,19 @@ func deleteFromPG(ctx context.Context, db pg.DB) error {
...
+	c := new(Config)
+	ver, err := sdb.Get(ctx, "/core/config", c)
+	if err != nil {
+		return nil, errors.Wrap(err)
+	} else if ver.Exists() {
```
tessr: @jbowens - can you weigh in on whether this is the right way to check if the key exists?

tessr: Also, can we pass `nil` in as the value (in this case `c`) if we're only checking for the existence of a key?

cmd/cored/main.go
```diff
@@ -290,6 +290,19 @@ func main() {
 		opts = append(opts, enableMockHSM(db)...)
 		chainlog.Printf(ctx, "Launching as unconfigured Core.")
 		h = core.RunUnconfigured(ctx, db, sdb, *listenAddr, opts...)
+
+		go func() {
+			ticker := time.Tick(5 * time.Second)
+			for {
+				select {
+				case <-ticker:
+					if conf != nil {
```
tessr: It's totally possible I'm missing something, but how would this happen?

cmd/cored/main.go
```diff
@@ -290,6 +290,19 @@ func main() {
...
+						return
+					}
+					core.CheckConfigMaybeExec(ctx, sdb, *listenAddr)
```
kr: Since we don't need to worry about trying to keep up with an exact 5-second schedule, it might be a little clearer to use Sleep instead:

```go
for {
	core.CheckConfigMaybeExec(ctx, sdb, *listenAddr)
	time.Sleep(5 * time.Second)
}
```

cmd/cored/main.go
```diff
@@ -290,6 +290,19 @@ func main() {
 		opts = append(opts, enableMockHSM(db)...)
 		chainlog.Printf(ctx, "Launching as unconfigured Core.")
 		h = core.RunUnconfigured(ctx, db, sdb, *listenAddr, opts...)
+
+		go func() {
+			ticker := time.Tick(5 * time.Second)
```
kr: Since we don't need to worry about trying to keep up with an exact 5-second schedule, it might be a little clearer to use Sleep instead:

```go
for {
	core.CheckConfigMaybeExec(ctx, sdb, *listenAddr)
	time.Sleep(5 * time.Second)
}
```

tessr: @kr oh, also, do you have an opinion on the right interval? I think 5 seconds might be a little long. (Maybe 1 second?)

core/config/config.go
vickiniu: It looks like we get a segfault if we pass in `nil` instead of `c`

cmd/cored/main.go
vickiniu: Added this since I thought we would need to exist if config got updated, but you're right — the process will just relaunch so this never executes

PR #1374, opened 2017-06-23
net/raft: add unit tests for node eviction

vickiniu:
Adds two unit tests for node eviction: for evicting a single node, and multiple nodes. Also slightly refactors to add a helper function to create a test Raft cluster.

### DIFF
--- a/net/raft/http.go
+++ b/net/raft/http.go
@@ -77,6 +77,23 @@ func (sv *Service) serveMsg(w http.ResponseWriter, req *http.Request) {
 		errorFormatter.Write(req.Context(), w, err)
 		return
 	}
+
+	// If message is from node not in cluster, tell node to remove itself
+	if sv.state.Peers()[m.From] == "" {
+		cc := raftpb.ConfChange{
+			ID:     atomic.AddUint64(&sv.confChangeID, 1),
+			Type:   raftpb.ConfChangeRemoveNode,
+			NodeID: m.From,
+		}
+		data, err := cc.Marshal()
+		if err != nil {
+			panic(err)
+		}
+		sv.send([]raftpb.Message{{
+			Type:    raftpb.MsgProp,
+			Entries: []raftpb.Entry{{Type: raftpb.EntryConfChange, Data: data}},
+		}})
+	}
 	sv.raftNode.Step(req.Context(), m)
 }
 
--- a/net/raft/http.go
+++ b/net/raft/http.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"os"
 	"strings"
 	"sync/atomic"
 
@@ -85,7 +84,7 @@ func (sv *Service) serveMsg(w http.ResponseWriter, req *http.Request) {
 // It receives notice that the node has been evicted and
 // kills the process.
 func (sv *Service) serveKill(w http.ResponseWriter, req *http.Request) {
-	os.Exit(0)
+	panic("Removed from cluster")
 }
 
 // serveJoin is registered as a handler for the /raft/join rpc.
--- a/net/raft/http.go
+++ b/net/raft/http.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"strings"
 	"sync/atomic"
 
@@ -78,6 +79,13 @@ func (sv *Service) serveMsg(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	sv.raftNode.Step(req.Context(), m)
+}
+
+// serveKill is registered as a handler for the /raft/kill rpc.
+// It receives notice that the node has been evicted and
+// kills the process.
+func (sv *Service) serveKill(w http.ResponseWriter, req *http.Request) {
+	os.Exit(0)
 }
 
 // serveJoin is registered as a handler for the /raft/join rpc.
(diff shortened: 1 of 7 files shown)


net/raft/raft_test.go
```diff
@@ -157,6 +136,53 @@ func TestClusterSetup(t *testing.T) {
 	if got != "bar" {
 		t.Errorf("reading /foo, nodeC got %q want %q", got, "bar")
 	}
+	nodeA.cleanup()
```
kr: It's probably better to `defer` each of these calls just after line 127 where the nodes are created.

net/raft/raft_test.go
```diff
@@ -157,6 +136,53 @@ func TestClusterSetup(t *testing.T) {
...
+
+func TestEvictMultiple(t *testing.T) {
+	ctx := context.Background()
+
+	// Create new test cluster
+	nodeA, nodeB, nodeC := newTestCluster(ctx, t)
+	addrB := nodeB.addr
+	addrC := nodeC.addr
+
+	// Have nodeC evict nodeA and nodeB
```
kr: This comment doesn't seem to match what the test is doing (nodeA is evicting the other two).

kr: But along the lines of this comment, would it be worth testing that a node other than node A can evict the other two (either here on in an additional test)?

vickiniu: If we try to evict node A, this triggers a panic from raft.go:738

```go
if cc.NodeID == sv.id {
	panic(errors.New("removed from cluster"))
}
```

I'm not sure what this piece of code is meant to do, but removing this check means that evicting the leader works. Do we want to keep this check in?

vickiniu:
etcd example for dropping messages w/o matching peer addresses: https://github.com/coreos/etcd/blob/e962b0c8498719c8d14c06572c38907ef4919343/etcdserver/raft.go#L302

net/raft/http.go
```diff
@@ -161,6 +168,17 @@ func sendmsg(addr string, data []byte, client *http.Client) {
...
+// Sends eviction notice to a node to kill the process
+func evictionNotice(addr string, client *http.Client) {
+	url := "https://" + addr + "/raft/kill"
+	resp, err := client.Post(url, contentType, nil)
+	if err != nil {
+		log.Printkv(context.Background(), "warning", err)
+		return
+	}
+	defer resp.Body.Close()
+}
```
jbowens: I wonder if instead of pushing eviction notices to evicted nodes, we should provide facilities for a node to recognize that it's evicted when it performs a `sendmsg` to contact other nodes?

Since these eviction notices are best-effort, it's still possible for an eviction to go unnoticed. For example:
1. Node _A_ is the Chain Core leader process.
2. A network partition causes node _A_ to be unreachable from the other processes.
3. An operator intervenes and evicts node _A_ from the cluster.
4. All the other nodes process the eviction entry and try to deliver the eviction notice to node _A_. Because node _A_ is unreachable, they all fail.
5. The network is restored.
6. Node _A_ is still evicted, but its Chain Core process is still running.

Node _A_ would continue to be Chain Core leader but without a working raft cluster.

I guess this scenario could also happen if the network partition is never restored and the node is never evicted. The leader process could continue despite not being able to participate in raft. Maybe the real solution is just to move Chain Core leadership into raft.

vickiniu: Just updated to have nodes recognize evictions when they get a response that they're no longer in the cluster — thanks!

net/raft/http.go
```diff
@@ -161,6 +185,17 @@ func sendmsg(addr string, data []byte, client *http.Client) {
 	defer resp.Body.Close()
 }
 
+// Sends eviction notice to a node to kill the process
+func evictionNotice(addr string, client *http.Client) {
```
tessr: I think you can remove this function now, right?

net/raft/raft.go
```diff
@@ -735,10 +745,10 @@ func (sv *Service) applyEntry(ent raftpb.Entry, writers map[string]chan bool) {
 		case raftpb.ConfChangeAddNode, raftpb.ConfChangeUpdateNode:
 			sv.state.SetPeerAddr(cc.NodeID, string(cc.Context))
 		case raftpb.ConfChangeRemoveNode:
+			sv.state.RemovePeerAddr(cc.NodeID)
 			if cc.NodeID == sv.id {
-				panic(errors.New("removed from cluster"))
+				os.Exit(0)
```
tessr: why this change?

vickiniu: This was basically for testing: `panic` would cause the test to crash with an error and I couldn't `recover` from the test code since the `panic` was in a different goroutine. Also, since this is expected behavior when a node is removed from the cluster, I'm not sure if a panic makes more sense than just exiting.

net/raft/http.go
vickiniu: Yes, definitely! Thanks for the catch

vickiniu:
PTAL :)

net/raft/raft.go
```diff
@@ -407,6 +407,15 @@ func (sv *Service) runUpdatesReady(rd raft.Ready, wal *wal.WAL, writers map[stri
 	sv.raftNode.Advance()
 }
 
+func (sv *Service) processMessages(msgs []raftpb.Message) []raftpb.Message {
```
tessr: could you add a comment here explaining what is going on, please?

vickiniu: added!

tessr:
LGTM

PR #1375, opened 2017-06-24
docs: remove reference to nonexistent "input block"

kr:


kr:
Unfortunately, renumbering the list items makes it look like the whole list changed, but really only item 1 changed and item 2 was deleted.

bobg:
LGTM

PR #1379, opened 2017-06-27
dashboard: align flash message content with page elements

dominic:


dominic:
Updated with feedback:

screen shot 2017-07-06 at 1 21 50 pm

dominic:
PTAL again @devongundry

devongundry:
1. Feels like the spacing of flash messages should be even at the top and the sides - 30px. like:
screen shot 2017-07-07 at 10 13 42 am

As a side note, `.Flash_main...` has a 15px top margin that doesn't seem to be doing anything. when i remove it, it doesn't change the positioning of the flash message.

2. Should we remove the border radius from the inline error messages in the PR as well?
screen_shot_2017-07-06_at_3 04 46_pm

dominic:
@devongundry This has been updated - PTAL again.

PR #1380, opened 2017-06-27
database/sinkdb: move storage from net/raft

tessr:
Nothing fancy here; just moving the actual writing-to-disk from the raft package to the sinkdb package in preparation for some other refactoring.


tessr:
As an added benefit: Providing a storage interface inside of sinkdb for use by raft and others means that we can actually write two storage engine files--one that uses rocksdb and one that calls out to a non-embedded data store, for cloud-hosting. /cc @dominic

PR #1381, opened 2017-06-27
net/raft: add unit test for node restart

vickiniu:
Adds unit tests and a fix for restarting a node in a cluster. Previously, restarting the initial node in a cluster would segfault since there was no snapshot to restore from. 

Now, we trigger snapshot while initializing a raft cluster to ensure that we can always restart nodes without errors.

### DIFF
--- a/database/sinkdb/state.go
+++ b/database/sinkdb/state.go
@@ -95,7 +95,8 @@ func (s *state) RestoreSnapshot(data []byte, index uint64) error {
 
 	// Chain Core 1.2.x generates snapshots without populating versions,
 	// so we manually initialize here to avoid nil errors after marshaling
-	// and unmarshaling the snapshot
+	// and unmarshaling the snapshot, which turns an initialized-but-empty
+	// map into a nil map
 	if s.version == nil {
 		s.version = make(map[string]uint64)
 	}
--- a/database/sinkdb/state.go
+++ b/database/sinkdb/state.go
@@ -92,6 +92,13 @@ func (s *state) RestoreSnapshot(data []byte, index uint64) error {
 	s.peers = snapshot.Peers
 	s.state = snapshot.State
 	s.version = snapshot.Version
+
+	if s.version == nil {
+		s.version = make(map[string]uint64)
+	}
+	if s.peers == nil {
+		s.peers = make(map[uint64]string)
+	}
 	return errors.Wrap(err)
 }
 
@@ -105,6 +112,7 @@ func (s *state) Snapshot() ([]byte, uint64, error) {
 		State:   s.state,
 		Peers:   s.peers,
 	})
+
 	return data, s.appliedIndex, errors.Wrap(err)
 }
 
--- a/database/sinkdb/state.go
+++ b/database/sinkdb/state.go
@@ -93,11 +93,11 @@ func (s *state) RestoreSnapshot(data []byte, index uint64) error {
 	s.state = snapshot.State
 	s.version = snapshot.Version
 
+	// Chain Core 1.2.x generates snapshots without populating versions,
+	// so we manually initialize here to avoid nil errors after marshaling
+	// and unmarshaling the snapshot
 	if s.version == nil {
 		s.version = make(map[string]uint64)
-	}
-	if s.peers == nil {
-		s.peers = make(map[uint64]string)
 	}
 	return errors.Wrap(err)
 }
--- a/database/sinkdb/state.go
+++ b/database/sinkdb/state.go
@@ -112,7 +112,6 @@ func (s *state) Snapshot() ([]byte, uint64, error) {
 		State:   s.state,
 		Peers:   s.peers,
 	})
-
 	return data, s.appliedIndex, errors.Wrap(err)
 }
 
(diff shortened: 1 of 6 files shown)


vickiniu:
PTAL :)

jbowens:
👀

jbowens:
> Previously, restarting the initial node in a cluster would segfault since there was no snapshot to restore from.

Was this true in general or just in the net/raft tests? We produce snapshots pretty infrequently, so when you run a local cored and restart it, most likely it has no snapshot and is booting solely from the write ahead log. That makes me wonder if the bug might be in the state_test.go State implementation.

net/raft/raft_test.go
```diff
@@ -141,6 +141,63 @@ func TestClusterSetup(t *testing.T) {
...
+	httpClient := nodeA.service.client
+
+	_, err := nodeA.service.Exec(ctx, set("/foo", "bar"))
+	must(t, err)
+
+	// Stop the node
+	nodeA.service.Stop()
+
+	// Restart the node
+	nodeA.service, err = Start(nodeA.addr, nodeA.dir, httpClient, nodeA.state)
```
jbowens: I think we might need to use `newTestState()` here instead of `nodeA.state`. Currently, `Start` expects the state to be the empty state. Then it's the net/raft package's responsibility to call RestoreSnapshot or Apply to bring it up-to-date if there's existing state in persisted in a snapshot or the WAL.

jbowens: When I tried this locally, it appeared to fix the panic on restart.

net/raft/raft_test.go
```diff
@@ -141,6 +141,63 @@ func TestClusterSetup(t *testing.T) {
...
+	nodeA.service.Stop()
+	nodeB.service.Stop()
+
+	// Restart the nodes
+	nodeA.service, err = Start(nodeA.addr, nodeA.dir, httpClientA, nodeA.state)
+	if err != nil {
+		nodeA.cleanup()
+		t.Fatal(err)
+	}
+	nodeB.service, err = Start(nodeB.addr, nodeB.dir, httpClientB, nodeB.state)
```
jbowens: Same here for using `newTestState`.

net/raft/raft_test.go
vickiniu: Awesome — just changed to `newTestState()` and restart is working without needing to trigger a snapshot

database/sinkdb/state.go
```diff
@@ -90,6 +90,13 @@ func (s *state) RestoreSnapshot(data []byte, index uint64) error {
 	s.peers = snapshot.Peers
 	s.state = snapshot.State
 	s.version = snapshot.Version
+
```
vickiniu: Added the map initialization since it was creating errors when we manually triggered snapshots: now that we've gotten rid of that, should we leave this in?

database/sinkdb/state.go
```diff
@@ -90,6 +90,13 @@ func (s *state) RestoreSnapshot(data []byte, index uint64) error {
...
+	if s.version == nil {
+		s.version = make(map[string]uint64)
+	}
+	if s.peers == nil {
+		s.peers = make(map[uint64]string)
+	}
```
vickiniu: Added the map initialization since it was creating errors when we manually triggered snapshots: now that we've gotten rid of that, should we leave this in?

jbowens: I don't think we need them because
1. no snapshot should ever have 0 peers. The manually triggered snapshots could've caused that to happen if the snapshot was taken before the initial conf change entries were applied.
2. a snapshot should always at least have a version for the nextNodeID.

We might be violating that second invariant if you create a brand new cluster and don't use it at all. Maybe instead of doing the version nil check, we should initialize the version map with a version for the nextNodeID in [newState](https://github.com/chain/chain/blob/main/database/sinkdb/state.go#L35).

Something like
```
		version: map[string]uint64{nextNodeID: 0},
```

jbowens: > 2. a snapshot should always at least have a version for the nextNodeID.

Ah sorry, I forgot Chain Core 1.2.x saves snapshots without any version data, so the nil check for s.version is important as long as we support migrating from Chain Core 1.2.x. Maybe keep that check but add a comment explaining that Chain Core 1.2.x generated snapshots without populating versions.

database/sinkdb/state.go
```diff
@@ -105,6 +112,7 @@ func (s *state) Snapshot() ([]byte, uint64, error) {
 		State:   s.state,
 		Peers:   s.peers,
 	})
+
```
tessr: extra newline

database/sinkdb/state.go
```diff
@@ -92,6 +92,13 @@ func (s *state) RestoreSnapshot(data []byte, index uint64) error {
 	s.peers = snapshot.Peers
 	s.state = snapshot.State
 	s.version = snapshot.Version
+
+	// Chain Core 1.2.x generates snapshots without populating versions,
+	// so we manually initialize here to avoid nil errors after marshaling
+	// and unmarshaling the snapshot
```
tessr: nit, but I'd add `, which will turn an initialized-but-empty map into a nil map`  to this comment.

jbowens:
lgtm

PR #1382, opened 2017-06-27
desktop/windows/ChainMgr: show error message for non-Windows

kr:
ChainMgr won't work on a non-Windows OS, so let's show
a helpful error message if someone tries to compile it
for another OS.


kr:
The idea here is to make the compile fail with a nice error message if you try to build ChainMgr on a non-Windows OS. Unfortunately it *also* makes gofmt fail with the same error message, so wercker won't let us land this.

Any ideas how to resolve this? I can think of a couple options:

1. Do nothing. Abandon this PR, keep the status quo, and let non-Windows OSes do what they will.
2. Change our wercker invocation of gofmt to somehow allow for exceptions — files that won't be gofmt'd.
3. (Anything else? Maybe there's a way to create a Go source file that gofmt will be happy with but that still fails compilation with a helpful error message?)

jeffomatic:
D'oh. I think I'd vote option 1 if this is the only time we have to worry about this issue.

PR #1384, opened 2017-06-28
core/config: add Options type

jbowens:
Add a new Options type for storing and accessing incremental
configuration options. For now, this only implements sets because that's
what we have an immediate need for. Scalar values will come later.

DefineSet defines a new configuration option that is set of k-tuples.
The caller must pass in a CleanFunc, used to validate and canonicalize
user input, and a EqualFunc, used to determine equality between tuples.

This PR only adds the Options type. Follow-ups will add the RPC handler
logic, corresponding corectl commands and the first config option.


jbowens:
PTAL

kr:
:eyes:

core/config/config.proto
```diff
@@ -26,3 +26,12 @@ message BlockSigner {
...
 
+message ValueTuple {
+	// values is an ordered list of the values in the tuple
+	repeated string values = 1;
+}
+
+message ValueSet {
+	// tuples is an unordered list of the tuples in the set
+	repeated ValueTuple tuples = 1;
+}
```
jbowens: Maybe these should be in an internal package?

core/config/options.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return tuples, nil
+}
+
+// Add adds the provided tuple to the configuration option set indicated
+// by key.
+func (opts *Options) Add(key string, tup []string) (sinkdb.Op, error) {
+	opt, ok := opts.schema[key]
+	if !ok {
+		return sinkdb.Op{}, errors.WithDetailf(ErrConfigOp, "Configuration option %q is undefined.", key)
```
kr: What do you think about sinkdb.Error here?

core/config/options.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return sinkdb.Op{}, errors.Sub(ErrConfigOp, err)
+	}
+
+	var existing ValueSet
+	ver, err := opts.sdb.GetStale(filepath.Join(sinkdbPrefix, key), &existing)
+	if err != nil {
+		return sinkdb.Op{}, err
+	}
+	idx := findIndex(existing.Tuples, cleaned, opt.equalFunc)
```
kr: Nit: name a pure function after what it returns, rather than what it does (that's for a procedure). So maybe `index` or `tupleIndex`.

core/config/options.go
```diff
@@ -0,0 +1,167 @@
...
+
+	var existing ValueSet
+	ver, err := opts.sdb.GetStale(filepath.Join(sinkdbPrefix, key), &existing)
+	if err != nil {
+		return sinkdb.Op{}, err
+	}
+
+	idx := findIndex(existing.Tuples, cleaned, opt.equalFunc)
+	if idx == -1 {
+		// tuple doesn't exists, so the sinkdb op is a no-op
```
kr: Nit: "exist"

core/config/options.go
jbowens: Forgot about sinkdb.Error. Nice

core/config/options.go
```diff
@@ -0,0 +1,167 @@
...
+	"context"
+	"path/filepath"
+
+	"chain/database/sinkdb"
+	"chain/errors"
+)
+
+const sinkdbPrefix = "/core/config/"
+
+var ErrConfigOp = errors.New("invalid config operation")
```
jbowens: I was planning on referencing it from the core/errors.go error table.

jbowens:
Any thoughts on whether the ValueSet and ValueTuple protos should be in an internal package?

kr:
Good q. I guess I usually lean toward making as much as possible private, so yeah.

core/config/options.go
```diff
@@ -0,0 +1,167 @@
...
+	modified := new(configpb.ValueSet)
+	modified.Tuples = append(modified.Tuples, existing.Tuples[:idx]...)
+	modified.Tuples = append(modified.Tuples, existing.Tuples[idx+1:]...)
+	return sinkdb.All(
+		sinkdb.IfNotModified(ver),
+		sinkdb.Set(filepath.Join(sinkdbPrefix, key), modified),
+	)
+}
+
+func tupleIndex(set []*configpb.ValueTuple, search []string, equal func([]string, []string) bool) int {
```
kr: Nit: `equal func(a, b []string) bool` might read a bit nicer, but nbd.

core/config/options.go
```diff
@@ -0,0 +1,167 @@
...
+	for idx, tup := range set {
```
kr: Nit: any particular reason to use `idx` instead of `i` here?

kr:
LGTM (just some nits)

Issue #1167, opened 2017-05-04
Detect and report out-of-date raft config on startup


bobg:
When starting cored, if the raft config is stale, you are likely to encounter a segfault in `Generator.makeBlock`. The non-obvious remedy is to remove the raft config (in `$HOME/.chaincore`) and restart.

Chain Core should detect this situation and report it to the user.

kr:
This should never happen for an end user. We need to make sure there's no such thing as a "stale" raft config -- if we change the schema, we need to migrate the data just like for any other database.

Do you have a stack trace or other error from when this happened? If it's a bug that could affect someone who's not hacking on cored itself, I would like to track it down.

tessr:
@danrobinson ran into it this morning, but I'm not sure exactly what he did to get there.

bobg:
I've encountered it too. I'm guessing it has something to do with changes to the format of files on disk?

The stack trace looks like:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x157b81c]

goroutine 132 [running]:
chain/protocol.(*Chain).GenerateBlock(0xc4203780c0, 0x23b3340, 0xc4204342c0, 0x0, 0x0, 0xed09d5670, 0x397f00e1, 0x23f9ac0, 0x0, 0x0, ...)
    /Users/danielrobinson/Projects/go/src/chain/protocol/block.go:57 +0x7c
chain/core/generator.(*Generator).makeBlock(0xc420181320, 0x23b3340, 0xc4204342c0, 0x0, 0x0)
    /Users/danielrobinson/Projects/go/src/chain/core/generator/block.go:73 +0x4a9
chain/core/generator.(*Generator).Generate(0xc420181320, 0x23b3340, 0xc4204342c0, 0x3b9aca00, 0xc420400740)
    /Users/danielrobinson/Projects/go/src/chain/core/generator/generator.go:96 +0x12f
created by chain/core.(*API).lead
    /Users/danielrobinson/Projects/go/src/chain/core/run.go:284 +0x2e9
```

kr:
Line 57 in block.go is

```
if timestampMS < prev.TimestampMS {
```

So presumably prev is nil.

danrobinson:
[...]

PR #1385, opened 2017-06-28
core/migrate: add core_id table

vickiniu:
To detect and report out-of-date raft configs (see https://github.com/chain/chain/issues/1167), we'll be writing the core ID to postgres and raft. This migration adds a core ID table to postgres to be able to crosscheck the core ID that we'll write to raft.

### DIFF
--- a/core/migrate/data.go
+++ b/core/migrate/data.go
@@ -72,4 +72,11 @@ var migrations = []migration{
 		DROP INDEX signers_type_id_idx;
 		DROP INDEX assets_sort_id;
 	`},
+	{Name: `2017-06-28.0.core.coreid.sql`, SQL: `
+		CREATE TABLE core_id (
+			singleton boolean DEFAULT true NOT NULL,
+			id text);
+		ALTER TABLE ONLY core_id
+			ADD CONSTRAINT core_id_pkey PRIMARY KEY (singleton);
+	`},
 }
--- a/core/migrate/data.go
+++ b/core/migrate/data.go
@@ -75,7 +75,9 @@ var migrations = []migration{
 	{Name: `2017-06-28.0.core.coreid.sql`, SQL: `
 		CREATE TABLE core_id (
 			singleton boolean DEFAULT true NOT NULL,
-			id text);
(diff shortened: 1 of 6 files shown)


core/schema.sql
```diff
@@ -2,6 +2,7 @@
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
```
tessr: I suspect this was added through a versioning mismatch.

core/migrate/data.go
```diff
@@ -72,4 +72,11 @@ var migrations = []migration{
 		DROP INDEX signers_type_id_idx;
 		DROP INDEX assets_sort_id;
 	`},
+	{Name: `2017-06-28.0.core.coreid.sql`, SQL: `
+		CREATE TABLE core_id (
+			singleton boolean DEFAULT true NOT NULL,
+			id text);
+		ALTER TABLE ONLY core_id
+			ADD CONSTRAINT core_id_pkey PRIMARY KEY (singleton);
+	`},
```
jbowens: To ensure that there's only ever one row, we might want to add a CHECK (singleton) constraint like [this](https://github.com/chain/chain/blob/main/core/schema.sql#L338). Otherwise, you could potentially have two rows in the table, one with singleton as true and one as false.

core/migrate/data.go
```diff
@@ -72,4 +72,13 @@ var migrations = []migration{
 		DROP INDEX signers_type_id_idx;
 		DROP INDEX assets_sort_id;
 	`},
+	{Name: `2017-06-28.0.core.coreid.sql`, SQL: `
+		CREATE TABLE core_id (
+			singleton boolean DEFAULT true NOT NULL,
+			id text,
+			CONSTRAINT core_id_singleton CHECK (singleton)
+		);
```
jbowens: might need to run `go run cmd/dumpschema/main.go` to pick up the constraint in core/schema.sql

jbowens:
lgtm

PR #1386, opened 2017-06-28
net/raft: refactor locking

jbowens:
Fix a data race in TestEvictMultiple detected by `go test -race`.
There were a few places where we locked stateMu before accessing State
and a few places where we didn't.

The State implementation from database/sinkdb does its own locking,
so this wasn't a real race in Chain Core, only in tests using the
net/raft state implementation.

We could remove the net/raft's mutex altogether if it weren't for two
things:
* the ConfState stored on Service that needs to be updated every time
  the raft configuration changes
* the stateCond condition variable used for waiting until applyEntry
  changes the state

This change
* adds a mutex to the net/raft test state
* documents that the State interface must be safe for concurrent access
* adds a new confMu mutex for protecting the ConfState
* removes all unnecessary uses of stateMu outside of the condition
  variable
* renames stateMu/stateCond to applyMu/applyCond to reflect that it's
  now only used for the applyEntry condition variable and not for
  protecting the state

An alternative to this approach could be to _always_ lock stateMu before
accessing State, but that would add unnecessary locking because sinkdb's
State requires its own locking regardless.

### DIFF
--- a/net/raft/raft.go
+++ b/net/raft/raft.go
@@ -855,8 +855,8 @@ func (sv *Service) getSnapshot() *raftpb.Snapshot {
 		panic(err)
 	}
 	sv.confMu.Lock()
-	defer sv.confMu.Unlock()
 	snap, err := sv.raftStorage.CreateSnapshot(index, &sv.confState, data)
+	sv.confMu.Unlock()
 	if err != nil {
 		panic(err)
 	}
--- a/net/raft/raft.go
+++ b/net/raft/raft.go
@@ -104,16 +104,22 @@ type Service struct {
 	// complete current state.
 	raftStorage *raft.MemoryStorage
 
+	// Condition variable for applyEntry's changes to state
+	applyMu   sync.Mutex
+	applyCond sync.Cond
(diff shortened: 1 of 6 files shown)


net/raft/raft.go
```diff
@@ -855,8 +855,8 @@ func (sv *Service) getSnapshot() *raftpb.Snapshot {
 		panic(err)
 	}
 	sv.confMu.Lock()
-	defer sv.confMu.Unlock()
 	snap, err := sv.raftStorage.CreateSnapshot(index, &sv.confState, data)
+	sv.confMu.Unlock()
```
vickiniu: Curious: what is the reason we want to avoid `defer confuMu.Unlock()` here?

jbowens: In applyEntry we lock applyMu before locking confMu, so we need to be careful to always lock them in that order to prevent deadlocks.

It's probably very unlikely, but if in the future we needed to acquire applyMu in this function we might forget that confMu is still locked until the function returns.

PR #1388, opened 2017-06-28
core: add incremental config options

jbowens:
Add support for setting and retrieving incremental configuration
options. There are still no configuration options defined so for now
these commands are useless.

* Add a field to /configure to support modifying incremental
  config options.
* Add a /config endpoint for retrieving configuration values.
* Add three subcommands to corectl: get, add and rm for retrieving and
  manipulating set configuration options.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -65,6 +65,9 @@ var commands = map[string]*command{
 	"init":                 {initCluster},
 	"evict":                {evictNode},
 	"allow-address":        {allowRaftMember},
+	"get":                  {get},
+	"add":                  {add},
+	"rm":                   {rm},
 	"wait":                 {wait},
 }
 
@@ -417,6 +420,69 @@ func evictNode(client *rpc.Client, args []string) {
 
 	req := map[string]string{"node_address": args[0]}
 	err := client.Call(context.Background(), "/evict", req, nil)
+	dieOnRPCError(err)
+}
+
+func get(client *rpc.Client, args []string) {
+	const usage = "usage: corectl get [key]"
+	if len(args) != 1 {
+		fatalln(usage)
+	}
+
+	req := map[string]interface{}{
+		"keys": []interface{}{args[0]},
+	}
+
+	var resp map[string][][]string
+	err := client.Call(context.Background(), "/config", req, &resp)
+	dieOnRPCError(err)
+	for key, tuples := range resp {
+		if len(tuples) == 1 {
+			fmt.Printf("%s: %s\n", key, strings.Join(tuples[0], " "))
+		} else {
+			fmt.Printf("%s:\n", key)
+			for _, tup := range tuples {
+				fmt.Printf(" - %s\n", strings.Join(tup, " "))
+			}
+		}
+	}
+}
+
+func add(client *rpc.Client, args []string) {
+	const usage = "usage: corectl add [key] [value]..."
+	if len(args) < 2 {
+		fatalln(usage)
+	}
+
+	req := map[string]interface{}{
+		"updates": []interface{}{
+			map[string]interface{}{
+				"op":    "add",
+				"key":   args[0],
+				"tuple": args[1:],
+			},
+		},
+	}
+	err := client.Call(context.Background(), "/configure", req, nil)
+	dieOnRPCError(err)
+}
+
+func rm(client *rpc.Client, args []string) {
+	const usage = "usage: corectl rm [key] [value]..."
+	if len(args) < 2 {
+		fatalln(usage)
+	}
+
+	req := map[string]interface{}{
+		"updates": []interface{}{
+			map[string]interface{}{
+				"op":    "rm",
+				"key":   args[0],
+				"tuple": args[1:],
+			},
+		},
+	}
+	err := client.Call(context.Background(), "/configure", req, nil)
 	dieOnRPCError(err)
 }
 
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -436,14 +436,9 @@ func get(client *rpc.Client, args []string) {
 	var resp map[string][][]string
 	err := client.Call(context.Background(), "/config", req, &resp)
 	dieOnRPCError(err)
-	for key, tuples := range resp {
-		if len(tuples) == 1 {
-			fmt.Printf("%s: %s\n", key, strings.Join(tuples[0], " "))
-		} else {
-			fmt.Printf("%s:\n", key)
-			for _, tup := range tuples {
-				fmt.Printf(" - %s\n", strings.Join(tup, " "))
-			}
+	for _, tuples := range resp {
+		for _, tup := range tuples {
+			fmt.Println(strings.Join(tup, " "))
 		}
 	}
 }
--- a/core/api.go
+++ b/core/api.go
@@ -67,6 +67,7 @@ type API struct {
 	accessTokens    *accesstoken.CredentialStore
 	grants          *authz.Store
 	config          *config.Config
+	options         *config.Options
 	submitter       txbuilder.Submitter
 	db              pg.DB
 	sdb             *sinkdb.DB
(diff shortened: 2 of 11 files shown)


jbowens:
PTAL

kr:
:eyes:

cmd/corectl/main.go
```diff
@@ -420,6 +423,69 @@ func evictNode(client *rpc.Client, args []string) {
...
+	req := map[string]interface{}{
+		"keys": []interface{}{args[0]},
+	}
+
+	var resp map[string][][]string
+	err := client.Call(context.Background(), "/config", req, &resp)
+	dieOnRPCError(err)
+	for key, tuples := range resp {
+		if len(tuples) == 1 {
+			fmt.Printf("%s: %s\n", key, strings.Join(tuples[0], " "))
```
kr: Since the key is known to the caller, do we have to echo it here? It would be cool if the output were greppable, sortable, etc. What about

```
$ corectl get foo
a1 a2 a3
b1 b3 b3
...
```

with no decoration? Then things like `corectl get foo | sort -k 3` are meaningful and it plays nicer with all the other tools like `uniq` and `join` and `awk`.

cmd/corectl/main.go
```diff
@@ -65,6 +65,9 @@ var commands = map[string]*command{
 	"init":                 {initCluster},
 	"evict":                {evictNode},
 	"allow-address":        {allowRaftMember},
+	"get":                  {get},
```
kr: @jeffomatic was suggesting using making even the read command different depending on whether the option is a tuple or a set. So like set/add vs get/*something*. Maybe *something* could be "list".

But seeing the implementation now, I'm not sure. A single tuple-value could just produce a single line of output. That seems fine to me.

cmd/corectl/main.go
jbowens: Ah yeah, that makes sense

jbowens:
@jeffomatic wdyt of this interface?

cmd/corectl/main.go
jeffomatic: Agreed, this looks just fine. To clarify, my concern was about the analogous SDK/RPC call. But it looks like at that level, there is an endpoint (`/config`) that simply returns the entire config object, and in that object, all keys map to slices of strings. With that level of generality, my concern is moot.

jeffomatic:
1. I think I might prefer a separate endpoint, say `/update-config`, that performs the add/remove operations. That way, `/configure` is not overloaded with two different input conventions, and really, two different use cases.
2. I could see it being useful having a command, say `clear`, that deletes all values of a key. The way `rm` works now, you have to know what's already in there in order to remove stuff.

kr:
> separate endpoint

+1 from me — the motivation for using the existing path was to avoid having a weird name left over once we've eliminated all the monolithic config data, but we can accomplish that just as well by picking any good name that happens to be different from `/configure`, and I think `/update-config` works pretty well.

> a command, say `clear`, that deletes all values of a key

Do we need it? Read-modify-write is not so bad, and an unconditional "delete all values" seems like a potential footgun.

jbowens:
> I think I might prefer a separate endpoint, say /update-config, that performs the add/remove operations. That way, /configure is not overloaded with two different input conventions, and really, two different use cases.

Yeah, I wavered between a separate endpoint and reusing /configure. I thought that having two endpoints /configure and /update-config is more confusing than a single endpoint with more complicated semantics. Also, if the ultimate goal is to remove the monolithic config, it'd be nice to  be able to use the well-named, verb `/configure` endpoint for incremental updates. As we move fields from the monolithic config to to incremental config, it'd be convenient to maintain backwards compatibility using a single endpoint. For example, we can take the existing `block_hsm_url` and `block_hsm_access_token` properties from the monolithic config and adapt them into corresponding writes to the config options.

jeffomatic:
> Do we need it? Read-modify-write is not so bad, and an unconditional "delete all values" seems like a potential footgun.

In the broad realm of configuring stuff, it's pretty common to replace config values with new ones you prefer, and from that perspective, it seems a bit fussy to force someone to say "rm x old-value; add x new-value". I'd rather say "clear x; add x new-value". Or even just "set x new-value".

I agree in the abstract that it's more dangerous to remove values without explicitly naming them, but that pattern feels pretty routine, and I guess my 2c is that it doesn't quite constitute a footgun.

jbowens:
> In the broad realm of configuring stuff, it's pretty common to replace config values with new ones you prefer, and from that perspective, it seems a bit fussy to force someone to say "rm x old-value; add x new-value". I'd rather say "clear x; add x new-value". Or even just "set x new-value".
>
> I agree in the abstract that it's more dangerous to remove values without explicitly naming them, but that pattern feels pretty routine, and I guess my 2c is that it doesn't quite constitute a footgun.

I think we should lean towards less interface surface area and fewer (even if very unlikely) footguns for now. If we find ourselves wishing for a `clear` command, it'll be an easy addition later.

kr:
It should be pretty rare to want to remove all the old values and have a moment where there are zero values before you add new ones.

For single strings (or single tuples) it would just be "set x new-value" like you said.

For sets I think you'd ~always want to incrementally replace the old values with new ones. e.g. for hsm urls, we probably won't even allow you to remove the last one because then you'd have no hsm urls.

jeffomatic:
> it'd be convenient to maintain backwards compatibility using a single endpoint

Not sure if I agree or disagree...could you explain a little more about this?

jbowens:
> Not sure if I agree or disagree...could you explain a little more about this?

As we move data into the config options store, we can add logic to translate pieces of the monolithic config to the corresponding option updates right here.: https://github.com/chain/chain/blob/4703ce55b1833f1c9d0642553e446e3e66d44eb5/core/core.go#L169-L170

jeffomatic:
> For single strings (or single tuples) it would just be "set x new-value" like you said.

Are we hoping to have that out of the gate? I didn't see it in this PR.

> It should be pretty rare to want to remove all the old values and have a moment where there are zero values before you add new ones.

Agreed. I'm angling more toward a mechanism by which you could easily replace a list with an entirely disjoint list. (As just one example, suppose we had multiple signerd URLs for block signing, and they all got reassigned to a new subdomain.) `clear` was just a minimal way to get that, but it's not the only way. What if we allowed `set` as you described above to be variadic? It would let you replace an existing list with an entirely new list.

jbowens:
> Are we hoping to have that out of the gate? I didn't see it in this PR.
 
We don't yet have a plan of a single tuple config option. Our immediate need is HSM urls, so we're starting with sets of tuples.

> As just one example, suppose we had multiple signerd URLs for block signing, and they all got reassigned to a new subdomain.

In that case, you would want to add the new URLs first and then remove the old URLs, unless you plan on taking downtime.

>  What if we allowed set as you described above to be variadic? It would let you replace an existing list with an entirely new list.

The set, add and rm commands already take multiple arugments to support tuples. For example

```
corectl add hsmurls https://hsm1.mycompany.com hsm1-access-token
```

If we allowed setting multiple, the interface would get a little hairy.
```
corectl set hsmurls https://hsm1.mycompany.com hsm1-access-token https://hsm2.mycompany.com hsm2-access-token
```

kr:
As far as the CLI is concerned, I'd vote for sticking with the incremental add and rm for swapping out a whole set of tuples. That is,

```
corectl add hsmurl https://newhost1/
corectl add hsmurl https://newhost2/
corectl add hsmurl https://newhost3/
corectl rm hsmurl https://oldhost1/
corectl rm hsmurl https://oldhost2/
corectl rm hsmurl https://oldhost3/
```

No matter what, we need some notation to let you specify multiple tuples. Using the command name itself (`corectl add`) as the separator seems pretty good.

Other tools (like dashboard) could give you a UI for providing multiple tuples and apply them all in one batch.

jeffomatic:
Cool, I think the corectl interface makes sense. I actually had a different idea of how this was working, so the examples helped.

kr:
> As we move fields from the monolithic config to to incremental config, it'd be convenient to maintain backwards compatibility using a single endpoint.

Not sure how much this would really matter, but we might also get some benefit from being able to apply granular config vars and the monolithic config object together atomically.

kr:
LGTM

PR #1389, opened 2017-06-29
desktop/windows/ChainMgr: drop Administrator privileges

kr:
Postgres refuses to run as Administrator on some
versions of Windows, so we need to drop privileges
before starting its process.


kr:
This runs successfully on desktop versions of windows (I tested on Windows 7), but those don't seem to exhibit the behavior we're trying to work around, so it still needs to be tested on Windows Server 2016 (or some similar version).

jeffomatic:
This is great. I can take a crack at testing this on EC2 tomorrow.

jeffomatic:
I logged into the EC2 Windows 2016 server I'm using and cherry-picked these commits onto `1.2-stable` (also had to run `go get golang.org/x/sys/windows`). I get the following error when running `ChainMgr.exe`:

```
app=core-manager 2017/07/07 00:05:11 error: The group may not be disabled.
```

With a little digging, I found that this error comes from the call to `AdjustTokenGroups` at the end of `dropAdminPrivs`. Haven't had a chance to investigate further but hopefully this is useful info.

kr:
Aha, I was wondering if maybe we need to call AdjustTokenGroups before AdjustTokenPrivileges. It might work to move the AdjustTokenPrivileges call to the end of the function (but I really don't know).

jeffomatic:
> call AdjustTokenGroups before AdjustTokenPrivileges

Will give it a shot.

jeffomatic:
> > call AdjustTokenGroups before AdjustTokenPrivileges
> Will give it a shot.

I tried moving the call to `AdjustTokenPrivileges` to the end, but unfortunately I still run into an error from `AdjustTokenGroups`.

kr:
Okay since this approach doesn't work, I'm closing this PR. Word on the internet is that to change privileges you have to start a new process, so we'd have to do something much closer to what pg_ctl does. I can work on that next.

jeffomatic:
Just FYI, I have [a branch](https://github.com/chain/chain/commit/433bf8d8bf8fd83d036890eaa60b0759b1736feb) that uses `pg_ctl` literally, and it seems to work on my EC2 server. I've sent an experimental build to some customers we think are experiencing this issue to verify that it works.

PR #1387, opened 2017-06-28
core/config: write core ID to postgres

vickiniu:
To detect and report out-of-date raft configs (see https://github.com/chain/chain/issues/1167), we'll be writing the core ID to postgres and raft. 

This writes the core ID to postgres when we configure.

PR #1390, opened 2017-06-29
core/config: check for stale Raft config

vickiniu:
See https://github.com/chain/chain/issues/1167

In https://github.com/chain/chain/pull/1385 and https://github.com/chain/chain/pull/1387, we write the config coreID to postgres whenever we configure a new cored process. This is to detect stale Raft configs resulting from when you restart a cored process, create a new Postgres db, but do not remove ~/.chaincore, preserving an old Raft config.

This PR checks the coreID stored in Raft against the coreID in Postgres (if any), and panics if there is a mismatch, suggesting that the user clean their Raft directory. Also adds a test to make sure we correctly detect stale configs.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -113,7 +113,7 @@ func idMatchesPG(ctx context.Context, id string, db pg.DB) (bool, error) {
 	const q = `SELECT id FROM core_id`
 	var pgID string
 	err := db.QueryRowContext(ctx, q).Scan(&pgID)
-	if err != sql.ErrNoRows {
+	if err != nil && err != sql.ErrNoRows {
 		return false, errors.Wrap(err)
 	}
 	return err == nil && pgID == id, nil
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -41,6 +41,7 @@ var (
 	ErrBadQuorum       = errors.New("quorum must be greater than 0 if there are signers")
 	ErrNoBlockPub      = errors.New("blockpub cannot be empty in mockhsm disabled build")
 	ErrNoBlockHSMURL   = errors.New("block hsm URL cannot be empty in mockhsm disabled build")
+	ErrStaleRaftConfig = errors.New("raft core ID doesn't match Postgres core ID")
 
 	Version, BuildCommit, BuildDate string
 
@@ -67,7 +68,8 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	} else if ver.Exists() {
 		raftID := c.Id
 		if !idMatchesPG(ctx, raftID, db) {
-			panic("Raft core ID doesn't match Postgres core ID: do you have a stale Raft config? Try: `rm -rf ~/.chaincore/raft`")
+			raftDir := HomeDirFromEnvironment() + "/raft"
+			return nil, errors.Wrap(ErrStaleRaftConfig, "Stale Raft config in "+raftDir)
 		}
 		return c, nil
 	}
@@ -107,13 +109,7 @@ func idMatchesPG(ctx context.Context, id string, db pg.DB) bool {
 	const q = `SELECT id FROM core_id`
 	var pgID string
 	err := db.QueryRowContext(ctx, q).Scan(&pgID)
-	if err != nil {
-		return false
-	}
-	if pgID == id {
-		return true
-	}
-	return false
+	return err == nil && pgID == id
 }
 
 // loadFromPG loads the stored configuration from Postgres.
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -65,6 +65,10 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	if err != nil {
 		return nil, errors.Wrap(err)
 	} else if ver.Exists() {
+		raftID := c.Id
+		if !idMatchesPG(ctx, raftID, db) {
+			panic("Raft core ID doesn't match Postgres core ID: do you have a stale Raft config? Try: `rm -rf ~/.chaincore/raft`")
+		}
 		return c, nil
 	}
 
@@ -97,6 +101,19 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 		panic(err)
 	}
 	return c, nil
+}
+
+func idMatchesPG(ctx context.Context, id string, db pg.DB) bool {
+	const q = `SELECT id FROM core_id`
+	var pgID string
+	err := db.QueryRowContext(ctx, q).Scan(&pgID)
+	if err != nil {
+		return false
+	}
+	if pgID == id {
+		return true
+	}
+	return false
 }
 
 // loadFromPG loads the stored configuration from Postgres.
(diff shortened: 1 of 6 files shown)


core/config/config.go
```diff
@@ -65,6 +65,10 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	if err != nil {
 		return nil, errors.Wrap(err)
 	} else if ver.Exists() {
+		raftID := c.Id
+		if !idMatchesPG(ctx, raftID, db) {
+			panic("Raft core ID doesn't match Postgres core ID: do you have a stale Raft config? Try: `rm -rf ~/.chaincore/raft`")
```
bobg: Don't hardcode `~/.chaincore/raft` here, populate it with the actual value, lest you advise people to remove the wrong directory (e.g. while testing with an alternate config).

Don't use backticks, which aren't meaningful in the console or log file where this message is likely to show up.

I might even consider rewording this message so that it doesn't contain a copy-and-paste command. For most things that would be a useful convenience, but for an `rm -rf` it's not a bad idea to require a little extra user intervention.

core/config/config.go
```diff
@@ -99,6 +103,19 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
...
 }
 
+func idMatchesPG(ctx context.Context, id string, db pg.DB) bool {
+	const q = `SELECT id FROM core_id`
+	var pgID string
+	err := db.QueryRowContext(ctx, q).Scan(&pgID)
+	if err != nil {
+		return false
+	}
+	if pgID == id {
```
bobg: Could end the function here with `return pgID == id`

core/config/config.go
kr: This should return an error, not panic.

core/config/config.go
```diff
@@ -99,6 +103,19 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
...
+		return true
+	}
+	return false
```
kr: The last 7 lines of this function can (and perhaps should) be just one line.

core/config/config_test.go
```diff
@@ -0,0 +1,42 @@
...
+	}()
+
+	ctx := context.Background()
+	sdb := sinkdbtest.NewDB(t)
+	_, db := pgtest.NewDB(t, pgtest.SchemaPath)
+	c := newTestConfig(t)
+
+	// Write config to sinkdb
+	sdb.Exec(ctx,
+		sinkdb.IfNotExists("/core/config"),
```
kr: Nit: this line isn't really necessary since we know it's a fresh sinkdb, but nbd.

vickiniu:
Unsure if we should handle these two cases in the same way:

- Raft has coreID, but Postgres has no coreID
- Raft has coreID, but Postgres has a different coreID

I'm not sure which scenario would land at the latter, but it seems stickier than just a stale Raft config. Right now, we throw the same error for both.

kr:
Personally I think it's fine to treat those to cases the same, as long as the error message is phrased in a way that covers both.

core/config/env.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"os"
+	"path/filepath"
+)
+
+// HomeDirFromEnvironment returns the directory to use
+// for reading config and storing variable data.
+// It returns $CHAIN_CORE_HOME,
+// or, if that is empty, $HOME/.chaincore.
+func HomeDirFromEnvironment() string {
```
kr: It looks like there's no corresponding delete of this function from chain/core?

Also, maybe consider moving this function separately in its own PR first?

core/config/config.go
```diff
@@ -65,6 +66,11 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	if err != nil {
 		return nil, errors.Wrap(err)
 	} else if ver.Exists() {
+		raftID := c.Id
+		if !idMatchesPG(ctx, raftID, db) {
+			raftDir := HomeDirFromEnvironment() + "/raft"
```
jbowens: might want to use [filepath.Join](https://godoc.org/path/filepath#Join) instead to ensure portability

core/config/config_test.go
```diff
@@ -0,0 +1,38 @@
+package config
+
+import (
+	"chain/database/pg/pgtest"
+	"chain/database/sinkdb"
+	"chain/database/sinkdb/sinkdbtest"
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"testing"
```
jbowens: nit: we group our imports with standard lib packages first, then vendored packages, then chain packages. like

```
import (
	"context"
	"crypto/rand"
	"encoding/hex"
	"testing"

	"chain/database/pg/pgtest"
	"chain/database/sinkdb"
	"chain/database/sinkdb/sinkdbtest"
)
```

kr: If you're using `goimports` to automatically insert imports, you can use `$CHAIN/bin/gochainimports` as a drop-in replacement that will do this grouping automatically.

core/config/env.go
vickiniu: Thanks for the suggestion! :) https://github.com/chain/chain/pull/1392

vickiniu:
Looks like the diff is still showing changes against `main` from when I opened the PR, not including recent commits on `main`. 

I rebased this branch to include moving `env.go` from core -> config (https://github.com/chain/chain/pull/1392#issuecomment-312107190) but those changes are all showing up in the diff. Not sure if there's a good way to fix this?

core/config/config.go
```diff
@@ -99,6 +106,13 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	return c, nil
 }
 
+func idMatchesPG(ctx context.Context, id string, db pg.DB) bool {
+	const q = `SELECT id FROM core_id`
+	var pgID string
+	err := db.QueryRowContext(ctx, q).Scan(&pgID)
+	return err == nil && pgID == id
+}
```
jbowens: If there's an error querying the database here, we'll end up returning ErrStaleRaftConfig from Load which is potentially inaccurate. Maybe we should add an error return value to differentiate between trouble connecting to the database and a wiped Postgres database.

Heads up that If there is no Core ID in the table, QueryRowContext will return [sql.ErrNoRows](https://godoc.org/database/sql#pkg-variables). idMatchesPG could translate that specific error into a `(false, nil)` return value and return any other error as-is.

jbowens:
looks great!

core/config/config.go
```diff
@@ -99,6 +109,16 @@ func Load(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*Config, error) {
 	return c, nil
 }
 
+func idMatchesPG(ctx context.Context, id string, db pg.DB) (bool, error) {
+	const q = `SELECT id FROM core_id`
+	var pgID string
+	err := db.QueryRowContext(ctx, q).Scan(&pgID)
+	if err != sql.ErrNoRows {
```
kr: If we get any other error here, shouldn't we treat that as an error, rather than "not a match"? I would have expected

```go
if err != nil {
```

here

kr:
LGTM

PR #1391, opened 2017-06-29
core: fix misreported error status codes

jeffomatic:


jeffomatic:
PTAL, it's a quickie.

jbowens:
lgtm

PR #1392, opened 2017-06-29
core/config: move HomeDirFromEnvironment from core to config

vickiniu:
For https://github.com/chain/chain/pull/1390, we need to call HomeDirFromEnvironment in the config package so we can identify the raft directory. Moving from core into config to avoid circular imports.


kr:
LGTM

PR #1393, opened 2017-06-29
core/config: add ListFunc

jbowens:
Add a ListFunc method to return a closure that returns the current set
of values for a set configuration option. Packages consuming
configuration options can accept closures of type

  func() [][]string

without needing to import core/config. Passing around these closures
should also ease unit testing.

### DIFF
--- a/core/config/options.go
+++ b/core/config/options.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"sync"
+	"sync/atomic"
 
 	"chain/core/config/internal/configpb"
 	"chain/database/sinkdb"
@@ -27,6 +29,9 @@ func New(sdb *sinkdb.DB) *Options {
 type Options struct {
 	sdb    *sinkdb.DB
 	schema map[string]option
+
+	errsMu sync.Mutex
+	errs   map[string]error
 }
 
 // CleanFunc is required when defining a new configuration option.
@@ -42,6 +47,20 @@ type option struct {
 	tupleSize int
 	cleanFunc CleanFunc
 	equalFunc EqualFunc
+}
+
+// Err returns any persistent errors encountered by ListFunc's closures
+// when retrieving configuration values. If there are multiple errors,
+// it'll return an arbitrary one.
+func (opts *Options) Err() error {
+	opts.errsMu.Lock()
+	defer opts.errsMu.Unlock()
+
+	// TODO(jackson): return all of the errors as a composite error?
+	for key, err := range opts.errs {
+		return fmt.Errorf("%q: %s", key, err.Error())
+	}
+	return nil
 }
 
 // DefineSet defines a new configuration option as a set of tuples of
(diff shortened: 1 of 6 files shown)


kr:
:eyes:

core/config/options.go
```diff
@@ -71,6 +72,35 @@ func (opts *Options) List(ctx context.Context, key string) ([][]string, error) {
...
 }
 
+// ListFunc returns a closure that returns the set of tuples for the
+// provided key.
+//
+// The configuration option for key must be a set of tuples.
+// ListFunc will panic if the provided key is undefined in the schema or
+// is defined as a scalar. The returned function will perform a stale
+// read of the configuration value.
+func (opts *Options) ListFunc(key string) func() ([][]string, error) {
```
kr: It's too bad the closure has to return an error… I don't immediately see any way around it tho.

core/config/options.go
```diff
@@ -71,6 +72,35 @@ func (opts *Options) List(ctx context.Context, key string) ([][]string, error) {
...
+	opt, ok := opts.schema[key]
+	if !ok {
+		panic(fmt.Errorf("unknown config option %q", key))
+	} else if opt.equalFunc == nil {
+		panic(fmt.Errorf("config option %q is a scalar, not a set", key))
+	}
+
+	return func() ([][]string, error) {
+		var set configpb.ValueSet
+		_, err := opts.sdb.GetStale(filepath.Join(sinkdbPrefix, key), &set)
```
kr: s/filepath/path/

core/config/options_test.go
```diff
@@ -49,3 +49,25 @@ func TestSetTuples(t *testing.T) {
...
+
+	ctx := context.Background()
+
+	must(t, sdb.Exec(ctx, opts.Add("example", []string{"foo"})))
+	must(t, sdb.Exec(ctx, opts.Add("example", []string{"bar"})))
+
+	// perform a linearizable read since ListFunc won't and we
+	// want a deterministic test case
+	_, err := opts.List(ctx, "example")
+	must(t, err)
```
kr: Nit: this could be `must(t, sdb.raft.WaitRead(ctx))`

core/config/options.go
```diff
@@ -71,6 +91,56 @@ func (opts *Options) List(ctx context.Context, key string) ([][]string, error) {
 	return tuples, nil
 }
 
+// ListFunc returns a closure that returns the set of tuples for the
+// provided key.
+//
+// The configuration option for key must be a set of tuples.
+// ListFunc will panic if the provided key is undefined in the schema or
```
kr: Nit: I think there's a bit of a convention in godoc to describe a function's behavior in the present tense, so "ListFunc panics if…" rather than "ListFunc will panic if…".

I'm sure we're not consistent about this in our codebase and I'm not even sure how strong of a convention it really is. Dunno why I noticed it just now. I think at least the standard library usually uses present tense e.g. https://godoc.org/net/http#ServeMux.Handle.

core/config/options.go
```diff
@@ -71,6 +91,56 @@ func (opts *Options) List(ctx context.Context, key string) ([][]string, error) {
...
+	} else if opt.equalFunc == nil {
+		panic(fmt.Errorf("config option %q is a scalar, not a set", key))
+	}
+
+	// old is the last successfully retrieved tuple set for this
+	// configuration key. The returned closure will update it on
+	// every successful lookup. If an error occurs, the closure
+	// returns this value.
+	var old atomic.Value
+	old.Store([][]string(nil))
```
jbowens: This ensure that `old.Load().([][]string)` won't panic if the first call to the closure fails to load the value. Otherwise `old.Load()` would return an untyped nil which cannot be asserted to [][]string.

Alternatively, we could return an error from ListFunc and perform a synchronous lookup before returning the closure so that we always have a previous value available in the closure. It's hard to tell how onerous returning an error from ListFunc would be since we don't have any call sites yet.

kr: I think returning nil is pretty sensible.

kr:
LGTM (just a couple nits)

core/config/options.go
jbowens: Good catch, I think I switched tenses a few times in there too, lol.

PR #1394, opened 2017-06-29
core/config: use path for forming sinkdb keys

jbowens:
Use the path package instead of path/filepath for forming sinkdb keys.


kr:
LGTM

PR #1395, opened 2017-06-30
core/config: introduce AddOrUpdate

jbowens:
For set configuration options, equality may be defined on a subset of
the fields of the tuple. Previously, we treated an Add of a tuple with
an existing key as a no-op. This splits the Add method into Add and
AddOrUpdate. Add will error if a tuple's key conflicts with an existing
tuple that is not exactly equal to the new tuple. AddOrUpdate will
replace the conflicting tuple.

This commit also updates corectl to allow providing the `-u` flag to
use the AddOrUpdate behavior.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -444,7 +444,13 @@ func get(client *rpc.Client, args []string) {
 }
 
 func add(client *rpc.Client, args []string) {
-	const usage = "usage: corectl add [key] [value]..."
+	const usage = "usage: corectl add [-u] [key] [value]..."
+
+	op := "add"
+	if len(args) > 0 && args[0] == "-u" {
+		op = "add-or-update"
+		args = args[1:]
+	}
 	if len(args) < 2 {
 		fatalln(usage)
 	}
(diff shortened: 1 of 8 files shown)


jbowens:
This makes me wonder if `add` is the write terminology for the corectl command.

Alternatively, we could  error if a tuple with the provided key already exists.

jbowens:
@jeffomatic thoughts on the interface? After writing this I'm leaning towards erroring instead, but I'm curious what your opinion is?

kr:
straw man idea:

```
$ corectl add foo tuplekey tupleval1
$ corectl add foo tuplekey tupleval1
error
$ corectl add -u foo tuplekey tupleval1
ok
```

kr:
oops

```
$ corectl add foo tuplekey tupleval1
$ corectl add foo tuplekey tupleval2
error
$ corectl add -u foo tuplekey tupleval2
ok
```

jeffomatic:
I think Keith's `-u` is a nice compromise!

kr:
Also I think if the tuple matches exactly that should still be ok, a no-op, so that adding a tuple can be idempotent. If it's not too hard to implement.

jbowens:
Refactored into Add and AddOrUpdate, and added the `corectl add -u` flag.

PTAL

PR #1396, opened 2017-06-30
docs: add cloud deploy instructions

jeffomatic:


docs/1.2/core/reference/deploying-to-the-cloud.md
```diff
@@ -0,0 +1,101 @@
...
+
+### AWS
+
+First, if you haven't already, download and install the AWS command-line tools, and then configure them:
+
+```
+aws configure
+```
+
+This will install your AWS credentials to a standard location that will be accessible to the `docker-machine` command. Next, run Docker Machine to create a new ECW instance that will serve as your Docker Host:
```
chrisgarvin: ECW -> EC2

jeffomatic: Nice catch, fixed

PR #1398, opened 2017-07-03
docs/future/core/reference: document new commands

jbowens:
Document the new configuration corectl commands.

### DIFF
--- a/docs/future/core/reference/corectl.md
+++ b/docs/future/core/reference/corectl.md
@@ -247,7 +247,7 @@ Argument:
 
 ### `add`
 
-Adds a tuple to a configuration option's set of values. The configuration option must be defined as a set. The number of value arguments provided must match the configuration option's tuple length. Add will error if a conflicting tuple already exists in the set. To overwrite the existing tuple, provide the `-u` flag.
+Adds a tuple to a configuration option's set of values. The configuration option must be defined as a set. The number of value arguments provided must match the configuration option's tuple length. `add` will error if a conflicting tuple already exists in the set. To overwrite the existing tuple, provide the `-u` flag.
 
 ```
 corectl add [key] [value]...
@@ -255,7 +255,7 @@ corectl add [key] [value]...
 
 Flags:
 
-* **-u**: Overwrite the existing tuple if it conflicts.
+* **-u**: Overwrites the existing tuple if it conflicts.
 
 Arguments:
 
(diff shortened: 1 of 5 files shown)


docs/future/core/reference/corectl.md
```diff
@@ -228,6 +231,52 @@ Argument:
...
+```
+
+Argument:
+
+* **key**: The configuration option to retrieve.
+
+
+### `add`
+
+Adds a tuple to a configuration option's set of values. The configuration option must be defined as a set. The number of value arguments provided must match the configuration option's tuple length. Add will error if a conflicting tuple already exists in the set. To overwrite the existing tuple, provide the `-u` flag.
```
jeffomatic: I like minimal docs but I think this might not quite get the picture across, since our configuration scheme is a bit idiosyncratic.

YMMV, but here's a strawman. Instead of:

> Adds a tuple to a configuration option's set of values. The configuration option must be defined as a set. The number of value arguments provided must match the configuration option's tuple length.

we try something more explanatory, such as:

> Adds a tuple to a configuration option's set of values. 
> 
> A configuration option is identified by a key, and consists of a set of tuples. Tuples for a given configuration option always the contain the same number of elements. For example, each tuple in the `block_signers` configuration option must contain two elements, a URL and a public key.

And then in the docs for `rm` and `get`, we can link to `add` for an explanation:

```
See [`add`](#add) for a description of how configuration options are structured.
```

docs/future/core/reference/corectl.md
```diff
@@ -228,6 +231,52 @@ Argument:
...
+
+```
+corectl add [key] [value]...
+```
+
+Flags:
+
+* **-u**: Overwrite the existing tuple if it conflicts.
```
jeffomatic: Grammar nit (sorry 😞 ): sometimes the verbs are in imperative mood ("Overwrite", "Remove"), and in other cases they're not ("Adds", "Retrieves").

docs/future/core/reference/corectl.md
jbowens: > A configuration option is identified by a key, and consists of a set of tuples.

The tricky part to communicate is that an option consists _either_ of a set of tuples or a single tuple. Some options you can add multiple values to, others only one. For now we only have sets but we should probably be careful with our language so that we don't confuse. I wouldn't be surprised if scalar option values are introduced before 1.3 ships.

jeffomatic: Got it. I think the scalar-ness or vector-ness of a given config option is vague with both the original and the suggestion.

How about:

> A configuration option is identified by a key, and consists of either a single tuple or a set of tuples. If a configuration option can have a set of tuples, each tuple in the set should contain the same number of elements.

PR #1399, opened 2017-07-03
core/blocksigner: add EnclaveClient

jbowens:
Move the Chain Enclave client from main.go into the core/blocksigner
package. Support multiple Chain Enclave URLs in the client to prepare
for retrieving multiple URLs from a core/config configuration option.

A future commit will define a new configuration option and propagate
its ListFunc closure to EnclaveClient.URLs.

### DIFF
--- a/core/blocksigner/enclave.go
+++ b/core/blocksigner/enclave.go
@@ -1,0 +1,57 @@
+package blocksigner
+
+import (
+	"context"
+	"time"
+
+	"chain/core/rpc"
+	"chain/crypto/ed25519"
+	"chain/encoding/json"
+	"chain/protocol/bc/legacy"
+)
+
+// enclaveTimeout is the time to wait for an HSM's response to a
+// sign-block request before moving on to the next HSM. Because
+// blocks are expected to be generated every second, one second
+// retireving a signature would be a very high latency.
+const enclaveTimeout = time.Second
+
+// EnclaveClient implements the Signer interface by calling
+// Chain Enclave to sign blocks.
+type EnclaveClient struct {
+	// URLs is called on every Sign request to retrieve the URLs
+	// and access tokens for Chain Enclave.
+	URLs       func() [][]string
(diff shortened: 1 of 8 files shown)


tonychain:
Cool. @jbowens when @ameets is back we should chat about the future of this code. We'd like to move to protos for talking to Enclave (but will continue to support the JSON-based interface for now)

kr:
:eyes:

core/blocksigner/enclave.go
```diff
@@ -0,0 +1,57 @@
...
+	"chain/crypto/ed25519"
+	"chain/encoding/json"
+	"chain/protocol/bc/legacy"
+)
+
+// enclaveTimeout is the time to wait for an HSM's response to a
+// sign-block request before moving on to the next HSM. Because
+// blocks are expected to be generated every second, one second
+// retireving a signature would be a very high latency.
+const enclaveTimeout = time.Second
```
kr: Maybe instead we should request a signature from all enclaves in parallel, and let the caller of Sign impose whatever overall timeout they want. That would reduce latency when there's a failure, and when one of the enclaves is just slow.

jbowens: Sounds good

jbowens:
Updated to request signatures from enclaves in parallel

core/blocksigner/enclave.go
```diff
@@ -0,0 +1,67 @@
...
+		// can modify it with the URL and access token
+		client := ec.BaseClient
+		client.BaseURL = tup[0]
+		client.AccessToken = tup[1]
+
+		go func(client rpc.Client) {
+			var signature []byte
+			err := client.Call(ctx, "/sign-block", body, &signature)
+			ch <- result{signature: signature, err: err}
+		}(client)
```
kr: Nit: since client is declared inside the loop body, it doesn't need to be a param here.

PR #1400, opened 2017-07-03
core/rpc: rename Client.Username

jbowens:
Rename Client.Username to ProcessID. Client uses basic auth when
AccessToken is not empty, but Username is the not the username used
in basic auth. Username is only used for the user agent. Rename it
prevent confusion.


jbowens:
PTAL, easy pickings

jbowens:
PTAL anyone

PR #1402, opened 2017-07-05
docker/de: remove recommendations to remove container

jeffomatic:
Most of our Docker docs use CLI examples that remove the container
after it has finished running. There doesn't appear to be any
reason for this related to running Chain Core itself, and it is
probably not what most people want when running Docker.

### DIFF
--- a/docker/de/Readme.md
+++ b/docker/de/Readme.md
@@ -17,7 +17,7 @@ $ sh bin/build-docker-de
 >**Note:** The `--name` flag allows you to name the container and refer to that name in subsequent commands.
 
 ```sh
-$ docker run --rm -p 1999:1999 --name chaincore chaincore/developer
+$ docker run -p 1999:1999 --name chaincore chaincore/developer
 ```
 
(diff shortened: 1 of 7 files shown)


docker/de/Readme.md
```diff
@@ -17,7 +17,7 @@ $ sh bin/build-docker-de
 >**Note:** The `--name` flag allows you to name the container and refer to that name in subsequent commands.
 
 ```sh
-$ docker run --rm -p 1999:1999 --name chaincore chaincore/developer
+$ docker run -p 1999:1999 --name chaincore chaincore/developer
 ```
 
 By default, once you stop a running a container, all data is lost. Chain Core stores data in three locations: a data directory, a separate Postgres database, and a log directory. To persist the data, create directories on your development machine and mount them to the container on `docker run`:
```
boymanjor: Technically, the phrase "By default, once you stop a running a container, all data is lost." is incorrect. The data is lost once you remove the container from the Docker host. Would you mind updating that in this PR.

jeffomatic: Okay got it, that's a good change.

PR #1405, opened 2017-07-05
net/http/httperror: log User-Agent with errors

jbowens:
Log request User-Agents along errors. Often we're only given excerpts
of logs in support requests, and it'll be helpful to be able to determine
the client being used (dashboard vs corectl vs various SDKs).

We might also want to improve some of the existing user agents. For
example, corectl currently just uses core/rpc.Client's default user agent.

### DIFF
--- a/net/http/httperror/httperror.go
+++ b/net/http/httperror/httperror.go
@@ -102,6 +102,7 @@ func (f Formatter) Log(ctx context.Context, err error) {
 		"status", resp.HTTPStatus,
 		"chaincode", resp.ChainCode,
 		"path", reqid.PathFromContext(ctx),
+		"useragent", reqid.UserAgentFromContext(ctx),
 		log.KeyError, errorMessage,
 	}
 	if resp.HTTPStatus == 500 {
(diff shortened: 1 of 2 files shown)


kr:
maybe not in this PR, but i think this user agent and path stuff doesn't belong in the reqid package. instead, anything that wants to log http-specific values should take an actual http.Request object

jbowens:
> maybe not in this PR, but i think this user agent and path stuff doesn't belong in the reqid package. instead, anything that wants to log http-specific values should take an actual http.Request object

Yeah, agreed. WDYT about the Core ID stuff? Since that bit is Chain-Core specific, maybe it should be in a chain/core handler.

jbowens:
Maybe I'll do the refactor first, so !PTAL

kr:
> WDYT about the Core ID stuff? Since that bit is Chain-Core specific, maybe it should be in a chain/core handler.

+1 Yeah that makes sense. This is why we have reqids to begin with. We can potentially print separate log messages for http info (user agent and oath), core id, and other things like error messages. The reqid will tie them all together.

jbowens:
Closing in favor of #1408

PR #1406, opened 2017-07-05
database/sinkdb: move file reading from net/raft

dominic:
Following #1380, this moves file reading from the raft package 
into the sinkdb package to prepare for future interfaces.


tessr:
We are not doing this anymore. :toot:

Issue #1345, opened 2017-06-15
net/raft: unit test raft snapshots


jbowens:
Currently, we don't have any unit tests covering raft snapshots. We only trigger snapshots every 10,000 raft entries, so most (all?) of our tests never trigger a snapshot. With the test State implementation from #1336, we can write a net/raft unit test that triggers snapshots more frequently. This kind of unit test could detect a regression of the crash from #1284.

`snapCount` and `nSnapCatchupEntries` are constants right now. We might need to copy those values into the Service struct so that the unit test can configure them with very low values to force frequent snapshots.

PR #1407, opened 2017-07-05
net/raft: add snapshot unit test

vickiniu:
Fixes https://github.com/chain/chain/issues/1345

Add unit test to verify raft snapshot matches node state. Moves `snapCount` and `nSnapCatchupEntries` into `service` so we can trigger snapshots more frequently in tests.

### DIFF
--- a/net/raft/raft_test.go
+++ b/net/raft/raft_test.go
@@ -272,15 +272,16 @@ func TestLeaderEviction(t *testing.T) {
 func TestSnapshot(t *testing.T) {
 	ctx := context.Background()
 
-	// Create new test cluster, triggers snapshot
+	// Create new one-node test cluster, triggers snapshot
 	node := newSnapshotTestNode(ctx, 2, t)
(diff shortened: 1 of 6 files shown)


net/raft/raft_test.go
```diff
@@ -264,6 +269,47 @@ func TestLeaderEviction(t *testing.T) {
...
+
+	// Create new test cluster, triggers snapshot
+	node := newSnapshotTestNode(ctx, 2, t)
+	defer node.cleanup()
+
+	// Wait for snapshot to successfully write, times out after 2 seconds
+loop:
+	for {
+		select {
+		case <-time.After(2 * time.Second):
```
jbowens: I think we need to move the `time.After(2 * time.Second)` above the for loop and save it to a local variable. Otherwise each iteration will invoke time.After and receive a new channel, and we'll always hit the default case.

jbowens:
Looks great!

PR #1408, opened 2017-07-05
net/http/reqid: remove path, coreid

jbowens:
Remove the Path and Core ID context annotation from net/http/reqid.
Replace it with additional middleware in chian/core.

Add logging of request user agents.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -212,6 +212,7 @@ func (a *API) buildHandler() {
 	if a.config != nil && a.config.BlockchainId != nil {
 		handler = blockchainIDHandler(handler, a.config.BlockchainId.String())
 	}
+	handler = loggingHandler(handler)
 	a.handler = handler
 }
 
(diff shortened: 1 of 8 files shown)


core/api.go
```diff
@@ -316,6 +317,22 @@ func blockchainIDHandler(handler http.Handler, blockchainID string) http.Handler
...
 }
 
+// loggingHandler pulls out request data and adds it to the request's
+// logging context.
+func loggingHandler(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		ctx := req.Context()
+		ctx = log.AddPrefixkv(ctx, "path", req.URL.Path)
+		if userAgent := req.UserAgent(); userAgent != "" {
+			ctx = log.AddPrefixkv(ctx, "useragent", userAgent)
```
kr: Instead of repeating this info on every line, do you think we could actually print a log message here and leave the log context alone? That would make subsequent log lines less verbose and maybe easier to read?

core/metrics.go
```diff
@@ -52,7 +51,9 @@ var (
 
 func coreCounter(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		countCore(reqid.CoreIDFromContext(req.Context()))
+		if coreID := req.Header.Get("Chain-Core-ID"); coreID != "" {
```
kr: Do we need the if statement?

core/api.go
jbowens: With our current logging, I think that'd actually be more verbose. Most of our logging is only when an error occurs, but logging here would log the user agent on every request.

core/metrics.go
jbowens: I think we need it somewhere, otherwise the count of Cores seen will be off by 1.

core/api.go
jbowens: If we want to avoid adding to the request context's logging prefix, another alternative is to propagate the request to httperror.Formatter.Write directly. It would be a bit invasive though because there are quite a few call sites, and I don't think all of them currently have access to the request.

core/metrics.go
kr: Okay, I just noticed this because it's a behavior change. (We would count the empty string before.)

core/api.go
kr: Okay yeah this seems good for now. At least if we want to change our output style, we can think about that separately.

PR #1411, opened 2017-07-06
database/localdb: new package

tessr:
This adds a new package, localdb, for writing to RocksDB. 

For now, this package exposes a single write operation, `Put`. We will soon want to replace this with something that supports test-and-sets, perhaps like `Exec` from sinkdb. However, that likely will require more design, so let's start here, for now.


tessr:
Tests should pass after Wercker's Docker image gets updated, which @vickiniu is working on. This could probably use at least a cursory review before then 🙏

jbowens:
:eyes:

database/localdb/localdb.go
```diff
@@ -0,0 +1,59 @@
+package localdb
+
+import (
+	"chain/errors"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/tecbot/gorocksdb"
```
jbowens: nit: import grouping, chain pkgs below vendored packages

database/localdb/localdb.go
```diff
@@ -0,0 +1,59 @@
...
+	// TODO(tessr): tune rocksdb. assess write options
+	wo := gorocksdb.NewDefaultWriteOptions()
+	return db.store.Put(wo, []byte(key), encodedValue)
+}
+
+func (db *DB) Get(key string, v proto.Message) error {
+	// TODO(tessr): tune rocksdb. assess read options
+	ro := gorocksdb.NewDefaultReadOptions()
+	slice, err := db.store.Get(ro, []byte(key))
+	defer slice.Free()
```
jbowens: I think we might want this defer after the err check.

It looks like Get will return a nil slice if err is non-nil: https://github.com/tecbot/gorocksdb/blob/master/db.go#L225-L228
And Slice.Free doesn't handle the case where s is nil: https://github.com/tecbot/gorocksdb/blob/master/slice.go#L30-L35

database/localdb/localdb.go
```diff
@@ -0,0 +1,59 @@
...
+	"chain/errors"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/tecbot/gorocksdb"
+)
+
+// DB provides access to a kv store.
+type DB struct {
+	store  *gorocksdb.DB
+	closed bool
```
jbowens: should closed be protected by a mutex?

tessr: yes. good call!

database/localdb/localdb.go
tessr: ah, yes, thank you! I'll add a test for that case, too.

tessr: just kidding, of course we'll only hit that case if something goes wrong inside rocksdb, which is hard to simulate in a test. so I'll just move that defer and call it good.

tessr:
Thanks to @vickiniu's tenacious debugging, wercker is now cooperating, and this is ready for another pass!

tessr:
@jbowens @kr can I get 👀  from one of you? 🙏

kr:
:eyes:

database/localdb/localdbtest/test.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package localdbtest;
+
+message TestItem {
+  string value = 1;
+}
```
jbowens: Is this proto only for tests within the localdb package? Maybe it should be in a localdb/internal/localdbtest package instead?

database/localdb/localdb.go
```diff
@@ -0,0 +1,65 @@
+package localdb
```
kr: It would be good to have some documentation (even something basic) on all public symbols and at the package level.

database/localdb/localdb.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"sync"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/tecbot/gorocksdb"
+
+	"chain/errors"
+)
+
+// DB provides access to a kv store.
```
kr: I think it is safe to call methods on a *DB concurrently, right? We should document that, since it is an exception to the usual convention (that it is not safe to call methods concurrently on most objects).

database/localdb/localdbtest/test.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package localdbtest;
+
+message TestItem {
```
kr: Is this meant to be used by other packages outside of database/localdb? If not, consider putting it in database/localdb/internal.

database/localdb/localdb.go
```diff
@@ -0,0 +1,65 @@
...
+func Open(rocksDir string) (*DB, error) {
+	// TODO(tessr): tune rocksdb. assess all these options
+	bbto := gorocksdb.NewDefaultBlockBasedTableOptions()
+	bbto.SetBlockCache(gorocksdb.NewLRUCache(3 << 30))
+	opts := gorocksdb.NewDefaultOptions()
+	opts.SetBlockBasedTableFactory(bbto)
+	opts.SetCreateIfMissing(true)
+	rocks, err := gorocksdb.OpenDb(opts, rocksDir)
+	if err != nil {
+		return nil, err
```
kr: Nit: errors.Wrap?

database/localdb/localdb.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return &DB{store: rocks}, nil
+}
+
+func (db *DB) Close() {
+	db.mu.Lock()
+	defer db.mu.Unlock()
+	if !db.closed {
+		db.store.Close()
+		db.closed = true
```
kr: What happens to the other methods after db is closed? Should those check for this flag before calling methods on db.store, and return an error?

database/localdb/localdbtest/test.proto
tessr: Yes. Sure!

database/localdb/localdb.go
tessr: Good call.

database/localdb/localdbtest/test.proto
tessr: No, and yes!

database/localdb/localdb.go
tessr: That seems like it must be true, but I'm not sure how to verify that. http://godoc.org/github.com/tecbot/gorocksdb doesn't say anything explicitly. (And checking both the gorocksdb godoc and some of the RocksDB docs for concurrency-related notes seems to mostly yield references to concurrent log compaction.)

database/localdb/localdb.go
tessr: How meticulous should I be about wrapping errors? For example, should I turn the line 

```go
return db.store.Put(wo, []byte(key), encodedValue)
```
(https://github.com/chain/chain/pull/1411/files/5b9b56740b6a0e968b788bc688a7a0047129a434#diff-8bd93a3248475c7605dc6a2fc602ddbaR29) 

into 

```go
err = db.store.Put(wo, []byte(key), encodedValue)
if err != nil {
        return errors.Wrap(err)
}
return nil
```
?

tessr: Or, I guess

```go
return errors.Wrap(db.store.Put(wo, []byte(key), encodedValue))
```

?

database/localdb/localdb.go
tessr: (Done.)

database/localdb/localdb.go
tessr: (Done.)

database/localdb/localdb.go
```diff
@@ -0,0 +1,86 @@
...
+		return errDBClosed
+	}
+
+	encodedValue, err := proto.Marshal(value)
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	// TODO(tessr): tune rocksdb. assess write options
+	wo := gorocksdb.NewDefaultWriteOptions()
```
jbowens: maybe for now we should set `wo.SetSync(true)`?

> Q: If my machine crashes and rebooted, will RocksDB preserve the data?
>
> A: Data is synced when you issue a sync write (write with WriteOptions.sync=true), call DB::SyncWAL(), or when memtables are flushed.

database/localdb/localdb.go
jbowens: We gucci

> Q: Is basic operations Put(), Write(), Get() and NewIterator() thread safe?
> 
> A: Yes.

https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ

database/localdb/localdb.go
tessr: I'm so glad _someone_ is reading the documentation!

database/localdb/localdb.go
kr: The more the better, until a performance profile tells us we should dial it back.

If you don't like hiding the Put call inside you can split the difference between those two:

```go
err = db.store.Put(wo, []byte(key), encodedValue)
return errors.Wrap(err)
```

database/localdb/localdb.go
tessr: (Done.)

tessr:
This is ready for another pass!

jbowens:
:eyes: 👍  2 👈

tessr:
🙇‍♀️  🎃 👪

PR #1414, opened 2017-07-06
dashboard: adjust tutorial layout

dominic:
Replaces #1378


devongundry:
This is looking really tasty. A few notes:

1. Now that we're not competing with the green create buttons, we can make the tutorial buttons blue background

2. On the form views, the tutorial header (with no body) still has the 1px offset issue

3. On the form views, the tutorial sidebar content would look better with 30px padding (instead of 25px) to align it horizontally with the header title.
screen shot 2017-07-06 at 2 39 46 pm

dominic:
Updated - PTAL again @devongundry

devongundry:
This feels amazing

PR #1415, opened 2017-07-06
dashboard: unify modal designs

dominic:


dominic:
PTAL @devongundry

devongundry:
A few notes:

1. I think we should remove the borders from the modals. they look cleaner without them:
screen shot 2017-07-07 at 9 56 06 am
screen shot 2017-07-07 at 9 55 08 am

2. the `<pre>` elements are getting cuttoff at the bottom on chrome. (see above) fine in safari.

PR #1416, opened 2017-07-06
docs: add search functionality

chrisgarvin:


chrisgarvin:
image

cmd/docgenerate/main.go
```diff
@@ -46,15 +47,19 @@ func main() {
...
+
+	// Generate SDK-specific documentation and search index files
 	for _, v := range versionPaths(srcdir) {
 		wg.Add(1)
 		go makeSdkDocs(wg, v, srcdir, outdir)
+		wg.Add(1)
+		go makeIndexInputFiles(wg, v, srcdir)
 	}
+
+	// Generate guides and reference docs
```
dominic: Any particular reason we moved this below the SDK + search index generation?

cmd/docgenerate/main.go
```diff
@@ -173,3 +199,70 @@ func versionPaths(srcdir string) []string {
...
+
+	for _, f := range files {
+		n := f.Name()
+
+		if strings.HasPrefix(n, ".") {
+			continue
+		}
+
+		if f.IsDir() {
+			if n == "enterprise" {
```
dominic: It might be nice to exclude the `LICENSE` file here as well.

docs/1.2/search-results.partial.html
```diff
@@ -0,0 +1,7 @@
+<div id="search-results">
```
dominic: We could close this div on line 1 instead of using an empty newline.

docs/common/js/search.js
```diff
@@ -0,0 +1,43 @@
...
+      name: 'Body',
+      weight: 0.6
+    }]
+  }
+  var fuse = new Fuse(steps, options)
+  var query = decodeURIComponent(window.location.search.slice(3).split('+').join(' '))
+  var result = fuse.search(query.removeStopWords())
+
+  $('#search-results').empty()
+  $('#search-results').append('<div id="search-result">Search results for: <b>' + query + '</b></div>')
```
dominic: We can't use an `id` more than once on any given page. These should be `class="search-result"` instead (and we'll have to update the CSS accordingly).

cmd/docgenerate/main.go
chrisgarvin: nah, not sure how that happened, switching it back.. good catch

cmd/docgenerate/main.go
```diff
@@ -136,6 +141,27 @@ func makeTempRepo(srcdir string) string {
 	return d
 }
 
+func makeIndexInputFiles(wg *sync.WaitGroup, version, srcdir string) {
+	defer wg.Done()
+
+	jsPath := path.Join(srcdir, "docs", version, "searchIndex.js")
+	os.Create(jsPath)
```
jeffomatic: Even though errors are unlikely here, it's a good idea for us to trap potential errors and `panic` or `log.Fatal` on them. So there are several locations in this function where that might be a good idea:

- `os.Create`
- `f.WriteString`

cmd/docgenerate/main.go
```diff
@@ -173,3 +199,70 @@ func versionPaths(srcdir string) []string {
...
+			contents = append(contents, createIndexFile(path.Join(parentPath, n), srcPath, version))
+		} else {
+			ext := filepath.Ext(n)
+			if ext == ".md" {
+				tempPath := path.Join(parentPath, n)
+				tempFile, err := ioutil.ReadFile(tempPath)
+				if err != nil {
+					panic(err)
+				}
+				tempString := string(tempFile)
```
jeffomatic: This is a nit: it could make sense to move the parsing logic into its own function that takes a byte slice as input, and returns a byte slice to be appended to the index contents. This would be lines 239 to 261, essentially everything after reading the file bytes, up to and including the json.Marshal.

cmd/docgenerate/main.go
dominic: :+1: Ah ha, gotcha. Might be worth adding a comment to that effect. So this also quietly updates your local `searchIndex.js` files in `$CHAIN`, and they get copied out after the fact.

PR #1417, opened 2017-07-06
docker/ci: build RocksDB and Snappy in wercker

vickiniu:
To invoke RocksDB and Snappy in Chain Core, we need to build them in the Docker image that Wercker starts up with. This change installs the dependencies that RocksDB and Snappy need in Docker, and also pulls pre-built static libraries for RocksDB and Snappy into the Docker container.

librocksdb and libsnappy are prebuilt here because they take really long to build, and because we've patched the Snappy project so we can't pull from the git project and build like we do for [protobuf](https://github.com/chain/chain/blob/main/docker/ci/install/install-protobuf#L16).


docker/ci/install/install-rocksdb
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -eo pipefail
+
+# RocksDB dependencies
+# Install gflags
```
tessr: Are you sure we need gflags and zstandard? Especially if we're going to just install snappy and rocksdb from pre-built binaries.

Also, I'm pretty sure I've been building this locally without them.

vickiniu: I'll try running it without them, but it lists gflags and zstandard as dependencies for CentOS (https://github.com/facebook/rocksdb/blob/master/INSTALL.md)

tessr: It's possible you need them to build rocksdb, but not to run it.

tessr: Of course, it's also possible that they're necessary! But let's see.

docker/ci/Dockerfile
```diff
@@ -19,8 +19,9 @@ ENV CHAIN="/go/src/chain" \
 ADD install /usr/bin
 
 RUN yum -y update \
-    && yum install -y autoconf automake bash curl gcc gcc-c++ git libtool make openssl-devel ruby wget unzip \
+    && yum install -y autoconf automake bash bzip2 bzip2-devel curl gcc gcc-c++ git libtool make openssl-devel snappy snappy-devel ruby tar wget unzip \
```
tessr: Ugh, I'm sorry I didn't catch this earlier, but I bet we don't need bzip or snappy here, either.

vickiniu: Without bzip, it ran into a runtime error — but snappy might not be necessary.

tessr: Cool. I bet snappy isn't, since we have the static lib.

tessr:
LGTM!!!!

PR #1418, opened 2017-07-07
protocol: fix ApplyValidBlock unprotected access

jbowens:
The current snapshot (c.state.snapshot) is protected by the mutex
c.state.cond.L. Previously ApplyValidBlock accessed the snapshot
without acquiring the mutex.

Also, fix a few comments in chain/protocol.


bobg:
LGTM

PR #1420, opened 2017-07-07
protocol: make block committing idempotent

jbowens:
Previously, we only called CommitAppliedBlock from a single goroutine:
either the core/generator or the core/fetch goroutine. This meant we
could access c.State() and apply new blocks to it without the possiblity
of another goroutine applying newer blocks first.

When we modify Chain to always have an up-to-date state tree regardless
of leader status, the goroutine listening to txdb.ListenBlocks heights
will need to also apply blocks. In preparation for this change, we need
to make committing & applying a block idempotent.

### DIFF
--- a/core/fetch/fetch.go
+++ b/core/fetch/fetch.go
@@ -161,11 +161,7 @@ func applyBlock(ctx context.Context, c *protocol.Chain, prevSnap *state.Snapshot
 	if err != nil {
 		return errors.Wrap(err, "validating fetched block")
 	}
-	snap, err := c.ApplyValidBlock(block)
-	if err != nil {
-		return errors.Wrap(err, "applying fetched block")
-	}
-	err = c.CommitAppliedBlock(ctx, block, snap)
+	err = c.CommitBlock(ctx, block)
 	return errors.Wrap(err, "committing block")
 }
 
--- a/generated/rev/RevId.java
+++ b/generated/rev/RevId.java
@@ -1,4 +1,4 @@
 
 public final class RevId {
-	public final String Id = "main/rev3314";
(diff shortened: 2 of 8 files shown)


jbowens:
PTAL

bobg:
:eyes:

bobg:
One thing I'd like to see in this PR, that will also help me review the rest of it: a comment (in `protocol/doc.go`, perhaps) describing the various phases of block processing. I once understood the sequence but am now having trouble reconstructing it. The names `ValidateBlock`, `ApplyValidBlock`, and `CommitAppliedBlock` were supposed to help illustrate the correct sequence, but those have changed, and there's also "saving" a block and "finalizing" a block (so the naming scheme was never very complete).

jbowens:
> One thing I'd like to see in this PR, that will also help me review the rest of it: a comment (in protocol/doc.go, perhaps) describing the various phases of block processing. I once understood the sequence but am now having trouble reconstructing it.

Ok, I [added some high-level documentation](https://github.com/chain/chain/pull/1420/commits/3d7ed688f9c3af029e58e83882e6525433797484).

> there's also "saving" a block and "finalizing" a block (so the naming scheme was never very complete).

Heh, yeah. FinalizeBlock should probably be called something like 'NotfyBlock.' I'm not 100% on this, but I think we'll eventually be able to remove FinalizeBlock from the Store interface, replacing it with some additional logic in chain/core. It really has more to do with intra-Core coordination than the protocol.

jbowens:
PTAL

bobg:
:eyes:

bobg:
LGTM

PR #1422, opened 2017-07-07
net/http/authz: use stale read in Delete

jbowens:
The Delete method does a read-modify-write so the read doesn't need to
be linearizable. If it reads a stale value, the IfNotModified condition
will fail the sinkdb operation.

By avoiding a consensus round on every call to Delete we also bring
down deleteGrantsByAccessToken's # of consensus rounds from _n + 1_
to 1, where _n_ is the number of policies.


tessr:
👀

tessr:
LGTM

PR #1423, opened 2017-07-07
Codeowners: add file

dominic:
Experimenting with GitHub's new automatic reviewer assignment
system, and potentially using it for branch protection.


kr:
Minor request, can we steer clear of the ALL CAPS file names? It's needlessly ostentatious.

kr:
If this is consistent with the other GitHub special files, their file name matching is not case sensitive.

dominic:
Good to know; will update.

kr:
We could call it `codeowners` or, if for some reason we want it to show up near the top of `ls` output (but why?) then we can capitalize it as `Codeowners`.

kr:
LGTM

PR #1424, opened 2017-07-10
dashboard: generate licenses during production builds

jeffomatic:
This commit adds `license-webpack-plugin` to the dashboard build pipeline. A new file, `3rdpartylicenses.txt`, will be generated in our production webpack bundle. The primary use case for this file is to as the basis for a more accurate NOTICE file in our dashboard source directory, one that only includes licenses for dependencies that we are actually re-distributing.

### DIFF
--- a/bin/bundle-dashboard
+++ b/bin/bundle-dashboard
@@ -2,6 +2,7 @@
 rm -rf $CHAIN/dashboard/node_modules/chain-sdk
 npm --prefix $CHAIN/dashboard install
 npm --prefix $CHAIN/dashboard run build
+rm -f $CHAIN/dashboard/public/3rdpartylicenses.txt # Don't need this for the bundle itself.
 go install chain/cmd/gobundle
 mkdir -p $CHAIN/generated/dashboard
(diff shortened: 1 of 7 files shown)


jeffomatic:
PTAL @danrobinson @dominic

dominic:
@jeffomatic Do we need to write this to `NOTICE`, or is that just a manual step when releasing?

jeffomatic:
We need to combine the output from this step with a manual cut/paste job that includes the Facebook patent grant, and then commit that to `NOTICE`.

dominic:
Okay cool :+1:

The `3rdpartylicenses.txt` file *will* get bundled into `generated/dashboard.go` here. Is that okay? It only adds 70KB, but is still extra noise.

jeffomatic:
> The 3rdpartylicenses.txt file will get bundled into generated/dashboard.go here. Is that okay? It only adds 70KB, but is still extra noise.

I pushed a commit that removes it from the bundle.

dominic:
@jeffomatic LGTM

danrobinson:
LGTM thanks Jeff!

PR #1429, opened 2017-07-10
docker/ci: fix rocksDB and snappy segfault

vickiniu:
The wercker tests for `database/localdb` failed with a shell segfault. This change installs correct dependencies, including GCC 4.8, to avoid segfaulting on starting up localDB


tessr:
LGTM 🦄

PR #1430, opened 2017-07-11
core: add enclave config option

jbowens:
Allow configuring Enclave URLs and access tokens through a new corectl
configuration option.


core/blocksigner/enclave.go
```diff
@@ -42,6 +43,12 @@ func (ec EnclaveClient) Sign(ctx context.Context, pk ed25519.PublicKey, bh *lega
 		client.BaseURL = tup[0]
 		client.AccessToken = tup[1]
 
+		// If the provided access token is just a password with no
+		// user, use a default basic auth username of 'chaincore'.
+		if !strings.Contains(client.AccessToken, ":") {
+			client.AccessToken = "chaincore:" + client.AccessToken
+		}
```
jbowens: instead of this business here, should rpc.Client prefer populating the basic auth password over the username? https://github.com/chain/chain/blob/main/core/rpc/rpc.go#L101-L111

core/config.go
```diff
@@ -0,0 +1,72 @@
...
+			// port 80 is already the default port for http
+			u.Host = host
+		case port == "443" && u.Scheme == "https":
+			// port 443 is already the default port for https
+			u.Host = host
+		}
+	}
+
+	// Remove trailing slash on path
+	if strings.HasSuffix(u.Path, "/") {
```
kr: Should we perhaps use `path.Clean`?

core/config.go
```diff
@@ -0,0 +1,72 @@
...
+	// for the local block signer. Tuple equality is defined on the
+	// the URL, not the access token.
+	opts.DefineSet("enclaves", 2, cleanEnclaveTuple, equalFirst)
+
+	return opts
+}
+
+// normalizeURL performs some low-hanging best-effort normalization
+// of the provided URL. See RFC3986, Section 6.
+func normalizeURL(urlstr string) (string, error) {
```
jbowens: instead of this DIY normalizing, we could use [this purell library](https://github.com/PuerkitoBio/purell), but I think we'd need to vendor:

github.com/PuerkitoBio/purell
github.com/PuerkitoBio/urlesc
golang.org/x/net/idna
golang.org/x/text/unicode/norm
golang.org/x/text/width

which feels like a bit much for our needs.

core/config_test.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	for u, want := range cases {
+		t.Run(u, func(t *testing.T) {
+			normalized, err := normalizeURL(u)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if normalized != want {
+				t.Errorf("got %s, want %s", normalized, want)
```
kr: This might be a nice time to write

```go
t.Errorf("normalizeURL(%q) = %q, want %s", u, normalized, want)
```

and then maybe we don't even need a subtest, idk.

jbowens:
Left a couple o Q's as review comments.

Also wondering:
* Should the option be called `enclaves`? `enclave_urls`? `block_enclaves`? `signer_enclave_urls`?
* Should config.Options.DefineSet return a type that includes the key name so that other code doesn't need to hard-code `"enclaves"`?
* On main, the presence of a HSM URL is used to determine whether MockHSM or EnclaveClient is used. Is it ok to _always_ use MockHSM if Chain Core is compiled with MockHSM, and _always_ use EnclaveClient if it's not?

core/run.go
```diff
@@ -121,13 +121,13 @@ func RateLimit(keyFn func(*http.Request) string, burst, perSecond int) RunOption
...
 // Chain Core.
-func RunUnconfigured(ctx context.Context, db pg.DB, sdb *sinkdb.DB, routableAddress string, opts ...RunOption) *API {
+func RunUnconfigured(ctx context.Context, confOpts *config.Options, db pg.DB, sdb *sinkdb.DB, routableAddress string, opts ...RunOption) *API {
 	a := &API{
 		db:           db,
 		sdb:          sdb,
 		accessTokens: &accesstoken.CredentialStore{DB: db},
 		grants:       authz.NewStore(sdb, GrantPrefix),
-		options:      config.New(sdb),
+		options:      confOpts,
```
kr: Any reason this shouldn't call `Config(sdb)` directly? I am not sure. (And same q for Run below.)

core/blocksigner/enclave.go
kr: Hrm I wish we could just require it to have a colon and/or use it exactly as presented. Do we have folks in the wild who are expecting a particular behavior?

core/config.go
kr: Agreed, I think what you have in the PR currently is about the right level of effort for now.

core/blocksigner/enclave.go
jbowens: Good question. AFAIK, Chain Core has always expected a user:password access token. If the input is just a string, it would only set the username in the basic auth request.

This change was motivated by my own confusion and [this TODO](https://github.com/chain/chain/blob/main/cmd/cored/main.go#L403-L404). Requiring it to have a colon sounds just as good to me.

kr:
> Should the option be called enclaves? enclave_urls? block_enclaves? signer_enclave_urls?

I have some idiosyncratic preferences here (so maybe take it with a grain of salt). I would lean toward singular, so it reads better in the `add` command (as opposed to the `get` command, where if you like you can read it as if the word "list" is implied at the end "get enclaveurl list"). And I think underscores (and hyphens) are ugly, so I would go for `enclave` or `enclaveurl`.

```
corectl add enclave http://foo/ xyz
or
corectl add enclaveurl http://foo/ xyz
```

I think we should prob not put "block" or "signer" in the name because there's no reason cored couldn't use this config for other enclave functions if necessary in the future. It's really just a set of fungible local Chain Enclave instances to talk to.

> Should config.Options.DefineSet return a type that includes the key name so that other code doesn't need to hard-code "enclaves"?

Ahh interesting. I can think of two reasons that would help. One is if we want to rename a config var, which we should really just not do, and the other is if we have typos in the name that won't be caught until runtime. But there won't be many places these names appear (once where it's defined, and once where we grab the live-updating accessor function) so I feel like the chance of this being a problem is low. WDYT?

> On main, the presence of a HSM URL is used to determine whether MockHSM or EnclaveClient is used. Is it ok to always use MockHSM if Chain Core is compiled with MockHSM, and always use EnclaveClient if it's not?

Great idea. We should totally do that.

core/run.go
jbowens: Currently cmd/cored needs access to it because the block signer is initialized in cored. Eventually, I think we should try to push that initialization into chain/core, but it'll require a bit of refactoring around the mockhsm.

jbowens:
Any thoughts on where to put the migration code? As we add additional config options, we might accumulate quite a bit of migration code.

kr:
Hmm as long as the migration code is confined to either the config package or a function in chain/core with "config" in the name (I think that's where it is now) I am not too worried if there's a lot of it, at least it's nicely contained and won't clutter up the rest of the codebase.

core/config.go
```diff
@@ -0,0 +1,97 @@
...
+// Config provides access to Chain Core configuration options
+// and their values.
+//
+// TODO(jackson): Remove the pg.DB argument when the PostgreSQL
+// url is stored in a configuration option.
+func Config(ctx context.Context, db pg.DB, sdb *sinkdb.DB) (*config.Options, error) {
+	opts := config.New(sdb)
+
+	equalFirst := func(a, b []string) bool { return a[0] == b[0] }
+	cleanEnclaveTuple := func(tup []string) error {
```
kr: Should this clean function also require the scheme to be https?

kr:
LGTM after one possible small thing to add

PR #1432, opened 2017-07-11
core: add warning for using mockhsm with multi-process Chain Core

vickiniu:
As we move mockhsm into localdb, mockhsm will no longer be consistent across a multi-process raft cluster. This shouldn't be a problem, as we expect the only people running mockhsm with multiple processes to be folks internally doing testing.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -25,6 +25,7 @@ var (
 	errAlreadyConfigured = errors.New("core is already configured; must reset first")
 	errUnconfigured      = errors.New("core is not configured")
 	errNoMockHSM         = errors.New("core is not configured with a mockhsm")
+	errMulticoreMockHSM  = errors.New("starting multi-core process with mockhsm")
 	errNoReset           = errors.New("core is not configured with reset capabilities")
 	errBadBlockPub       = errors.New("supplied block pub key is invalid")
 	errNoClientTokens    = errors.New("cannot enable client auth without client access tokens")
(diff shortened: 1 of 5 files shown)


core/core.go
```diff
@@ -25,6 +25,7 @@ var (
 	errAlreadyConfigured = errors.New("core is already configured; must reset first")
 	errUnconfigured      = errors.New("core is not configured")
 	errNoMockHSM         = errors.New("core is not configured with a mockhsm")
+	errMockHSM           = errors.New("starting core with mockhsm: should only be used in development")
```
tessr: A few minor things: 

1. It's overkill to make this an `error`. Since we only ever log it's string representation, we don't need to make it an error. Let's just log the string inline. 
2. Since this message shows not when the core starts, but when the cluster is initialized, what about a message like: "warning: this core uses a mockhsm. mockhsm data does not sync across coreds" (Especially since that really is the issue we're concerned about.)

bobg:
Don't use the term "multi-core Chain Core" to mean a multiple-process core. (Say "multiple-process" instead.) Too easy to confuse with a multiple-Chain-Core network.

bobg:
(Also, didn't we decide to share MockHSM data among the processes of a Chain Core by making it a standalone service with its own RPC URL?)

kr:
When you say standalone service do you mean a separate process? I think we should not do that since it would increase operational complexity. If we need mockhsm data to be synchronized, we should store it in a synchronized database. However, I think we don't need it to be synchronized.

tessr:
No, I'm pretty sure we decided that the enclave team had a variety of changes they wanted to make to mockhsm/mock enclave that would make it mirror real-hsm/enclave more closely, which may include making it a standalone service. But I don't think that's been designed thoroughly, and besides, it's out of scope for this localdb project. This is just a small measure that will help prevent us from getting tripped up over our own dev environments.

tessr:
@vickiniu can you also update the PR title and body to not use the phrase "multi-core Chain Core"? 🙏

PR #1434, opened 2017-07-11
sdk/node: support CA transaction flow

dominic:
This moves the HSM signer object underneath the main
`client` object, and adds a placeholder for the `finalize`
call, that passes through the original object, appropriately
wrapped in a batch response as expected.

### DIFF
--- a/sdk/node/src/api/transactions.js
+++ b/sdk/node/src/api/transactions.js
@@ -415,6 +415,7 @@ const transactionsAPI = (client) => {
      *                         added, as well as errors.
      */
     signBatch: (templates, cb) => finalizeBatch(templates)
+      // TODO: merge batch errors from finalizeBatch
       .then(finalized => client.signer.signBatch(finalized.successes, cb)),
 
     /**
--- a/sdk/node/src/api/transactions.js
+++ b/sdk/node/src/api/transactions.js
@@ -385,6 +385,25 @@ const transactionsAPI = (client) => {
     },
 
     /**
+     * sign - Sign a single transaction.
+     *
+     * @param {Object} template - A single transaction template.
+     * @param {objectCallback} [callback] - Optional callback. Use instead of Promise return value as desired.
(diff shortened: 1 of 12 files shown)


sdk/node/src/api/transactions.js
```diff
@@ -385,6 +397,37 @@ const transactionsAPI = (client) => {
...
+     * @returns {BatchResponse} Tranasaction templates with all possible signatures
+     *                         added, as well as errors.
+     */
+    signBatch: (templates, cb) => finalizeBatch(templates)
+      .then(finalized => client.signer.signBatch(finalized.successes, cb)),
+
+    /**
+     * finalize - Finalize a single transaction
+     */
+    finalize,
```
dominic: I agree - I saw https://github.com/chain/chain/blob/ca/docs/future/protocol/specifications/ca-api.md#js-2 accessing it directly, but I have trouble thinking of why a user would want to call this. I'll take these out.

dominic:
@jeff removed dashboard changes and the public finalize interface - PTAL again

sdk/node/src/api/transactions.js
```diff
@@ -385,6 +397,27 @@ const transactionsAPI = (client) => {
     },
 
     /**
+     * sign - Sign a single transaction.
+     *
+     * @param {Object} template - A single transaction template.
+     * @param {objectCallback} [callback] - Optional callback. Use instead of Promise return value as desired.
+     * @returns {Object} Transaction template with all possible signatures added.
+     */
+    sign: (template, cb) => finalize(template)
```
jeffomatic: Just a heads-up. When `finalize` does come in, we're going to have to decide how to handle batch errors. If we do it like multiple HSMs, we'll need to add code that distinguishes between errors that occur in the finalize step, and the errors that occur in the sign step.

dominic: Good point; let me add a `TODO` here at least so it's not forgotten.

PR #1436, opened 2017-07-11
docs: split instructions for creating and starting docker container

dominic:


docs/1.2/core/get-started/install.partial.html
```diff
@@ -61,11 +61,23 @@ <h3 data-no-anchor="">Linux</h3>
...
-          <h3>Run the latest Docker image</h3>
+          <h3>Create a container from the latest Docker image</h3>
+
+          <pre>
+            <code class="language-bash">docker create -t -p 1999:1999 chaincore/developer</code>
+          </pre>
+
+          <h3>Run your container</h3>
+
+          <p>Copy the conainter ID returned from the last command. You can use this to start
```
boymanjor: s/conainter/container/

jeffomatic:
Sorry to make more work, but there are several places where we give instructions about the Docker container (search for `docker run`). Would you mind making analogous updates?

docs/1.2/core/get-started/install.partial.html
```diff
@@ -92,12 +104,23 @@ <h3 data-no-anchor="">Linux</h3>
...
+          <h3>Create a container from the latest Docker image</h3>
 
           <pre>
-            <code class="language-bash">docker run -it -p 1999:1999 chaincore/developer</code>
+            <code class="language-bash">docker create -t -p 1999:1999 chaincore/developer</code>
           </pre>
 
+          <h3>Run your container</h3>
+
+          <p>Copy the conainter ID returned from the last command. You can use this to start
```
boymanjor: Same (sp) error as above.

dominic:
Sure - the only one that looks like a clear change is the Ivy instructions. Do you think we should update the cloud deploy instructions? My impression is that local containers are the real pain point for users.

jeffomatic:
We've got past versions of the Docker install doc, `docker/de/Readme.md` (which I have seen people in community support link to), and I guess we should think about updating the cloud deploy doc, IF the problem is that most people don't know how to use Docker and need the step-by-step instructions as education.

dominic:

docker/de/Readme.md
```diff
@@ -46,6 +40,22 @@ $ docker stop chaincore
...
 
+#### Persist data outside of Docker
+
+Chain Core stores data in three locations: a data directory, a separate Postgres database, and a log directory. To persist the data outside the container, create directories on your development machine and mount them to the container on `docker run`:
+
+```sh
+$ mkdir -p /path/to/store/datadir
+$ mkdir -p /path/to/store/postgres
+$ mkdir -p path/to/store/logs
+$ docker run -p 1999:1999 \
```
jeffomatic: Do you not want to do `docker create` here?

docker/de/Readme.md
```diff
@@ -16,22 +16,16 @@ $ sh bin/build-docker-de
 
 >**Note:** The `--name` flag allows you to name the container and refer to that name in subsequent commands.
 
+First create a new container with the name `chaincore`:
+
 ```sh
-$ docker run -p 1999:1999 --name chaincore chaincore/developer
+$ docker create -p 1999:1999 --name chaincore chaincore/developer
```
jeffomatic: Do we not want `-t` here, as with the other examples? Do we know what it does?

dominic: I left it as I found it without `-t` (though the missing `-i` is an indicator that it was just out-of-date). 

I think without `-t`, you can't close your session with `ctrl-c`. I'll add it here to be consistent.

docker/de/Readme.md
dominic: I think it matters less here. Using `docker run` with the volumes persisted externally, it doesn't really matter whether you start the same container or not. Should we remove the `--name` flag from this command?

boymanjor: `--name` is a nice addition because the caller does not have to reference the container id. That said, I don't mind whether we advertise this flag or not. I only vote for consistency across the docs.

boymanjor: So I guess my vote would be to keep it here.

dominic: @boymanjor Would you prefer to see name in the getting started docs too? We've never included it in our main guide.

boymanjor: Now that I think about it, if users have reference this doc we should probably get rid of the `--name` flag here. My reason being we probably shouldn't abstract away the idea of a container id from non-power users of Docker. The power users will add the flags they so choose. Wdyt @dominic @jeff.

boymanjor: By "here" I mean the entire document.

dominic: I'm okay with it, but that's a really big change to this doc. Maybe a deeper question is "who is this document for?" We've linked to it in support before, but it's not on our docs site.

boymanjor: Discussed out of band: we're going to leave `--name` here and adapt this document into an official Docker primer for our docs site.

PR #1439, opened 2017-07-12
cmd/cored: handle uninitialized raft cluster

jbowens:
core.Config can return raft.ErrUninitialized if the raft cluster is
uninitialized. Like in the config.Load call above, it should be ignored.


kr:
LGTM

PR #1440, opened 2017-07-12
dashboard: add newly generated keys to existing key list

chrisgarvin:


dashboard/src/features/shared/actions/create.js
```diff
@@ -41,6 +43,9 @@ export default function(type, options = {}) {
                 return chainClient().mockHsm.keys.create({alias})
               }).then(newKey => {
                 data.rootXpubs.push(newKey.xpub)
+                return chainClient().mockHsm.keys.query({})
```
dominic: Do we need this `query` here? Could `keyList` instead just be `[newKey]`?

dashboard/src/features/shared/actions/create.js
```diff
@@ -7,6 +7,8 @@ export default function(type, options = {}) {
   const listPath = options.listPath || `/${type}s`
   const createPath = options.createPath || `${listPath}/create`
   const created = (param) => ({ type: `CREATED_${type.toUpperCase()}`, param })
+  const receivedKeys = (param) => ({ type: 'RECEIVED_MOCKHSM_ITEMS', param })
```
jeffomatic: I think this strategy will work and it is the smallest fix in terms of lines of code, but it may be an abuse of the reducer abstraction to use `RECEIVED_MOCKHSM_ITEMS` here. In other words, this may put the solution to the problem in the wrong part of the code. There's no other reason to make this file aware of the `RECEIVED_MOCKHSM_ITEMS` other than to fix a state inconsistency, but fixing state issues should happen in the reducers.

One other route to consider is expanding the reducer for the relevant item in the state tree ([this one?](https://github.com/chain/chain/blob/main/dashboard/src/features/mockhsm/reducers.js)) so that it handles both `RECEIVED_MOCKHSM_ITEMS` and `CREATED_MOCKHSM`.

dominic: This is a particularly tricky spot because the path we're in is `CREATED_ACCOUNT`, not `CREATED_MOCKHSM`. Creating an account has the occasional side effect of generating new Mock HSM keys.

Jeff, is your suggestion to dispatch `CREATED_MOCKHSM` here instead, and make sure the reducer handles that action appropriately? (i.e. a singular case of `RECEIVED_MOCKHSM_ITEMS`). As it stands, we're calling out to the vanilla Chain SDK here, so there's nothing for the reducer to do at this ponit.

An alternative would be dispatching `actions.mockhsm.create` as a thunk action here, which could then appropriately dispatch actions in its own flow, and return the Promise result after completion.

PR #1441, opened 2017-07-13
vendor: add legalese for vendored go dependencies

jeffomatic:
This commit ensures we're complying with open source licenses. For
safe keeping, also include PATENTS and NOTICE (copyright) files.

A subsequent commit will update vendoring instructions to ensure these
files are captured in the future.


jeffomatic:
PTAL @danrobinson et al

danrobinson:
LGTM, thanks!

PR #1442, opened 2017-07-13
Readme.md: add copyright info to vendoring instructions

jeffomatic:


jeffomatic:
PTAL

Readme.md
```diff
@@ -144,6 +144,18 @@ $ rm -rf $CHAIN/vendor/github.com/kr/pretty/.git
 It can change the behavior of cp and put the files
 in the wrong place.)
 
+Since this is a publicy-available open source project, there are some copyright
```
kr: sp: public_l_y

> Since this is a publicy-available

Is this the reason for excluding GPL-family licenses? Maybe better to just list the rules without even a token justification?

Readme.md
```diff
@@ -144,6 +144,18 @@ $ rm -rf $CHAIN/vendor/github.com/kr/pretty/.git
...
+issues to keep in mind:
+
+1. Don't vendor packages that are licensed under any flavor of the GPL.
```
kr: Like you suggested offline, maybe better to explicitly list the licenses we're concerned about (while still broadly including any other license with similar terms).

Like, "Don't vendor packages that are licensed under GPL, LGPL, AGPL, or any other license with similar terms. When in doubt, leave it out (or ask)." ?

jeffomatic: @danrobinson just suggested a whitelist, so maybe we can get him to help here.

Readme.md
jeffomatic: Yep, sounds good to me.

Readme.md
danrobinson: How about:

> Please ask before vendoring dependencies that are not licensed under one (or more) of the following open-source licenses: MIT, ISC, BSD 3-clause, BSD 2-clause, Apache 2.0.

Readme.md
danrobinson: I agree that the explanation is not necessary.

also: publicy -> publicly

PR #1443, opened 2017-07-13
dashboard: add account receiver creation to integration tests

chrisgarvin:


dashboard/test/conf/wdio.local.js
```diff
@@ -43,7 +43,7 @@ exports.config = {
   // and 30 processes will get spawned. The property handles how many capabilities
   // from the same test should run tests.
   //
-  maxInstances: extended ? 1 : 10,
+  maxInstances: extended ? 1 : 1,
```
dominic: This can just be `maxInstances: 1`

dashboard/test/helpers.js
```diff
@@ -84,10 +85,25 @@ const issueTransaction = (client) => expect(
   .then(tpl => client.transactions.submit(tpl))
 ).to.be.fulfilled
 
+const createAccount = (account = 'account') => {
```
jeffomatic: maybe rename `account` to `name`?

chrisgarvin: yeah, good call

dashboard/test/specs/accounts.js
```diff
@@ -31,13 +31,41 @@ describe('accounts', () => {
...
+  describe('creating receivers', () => {
+    let account
+
+    before(() => expect(testHelpers.ensureConfigured()).to.be.fulfilled
+      .then(() => {
+        return testHelpers.createAccount('receiver-test')
+      })
+      .then((acc) => {
+        account = acc
+      })
```
dominic: This doesn't need to be a second `then`.

dashboard/test/specs/accounts.js
```diff
@@ -31,13 +31,41 @@ describe('accounts', () => {
...
       browser.getText('.AccountShow').should.contain(alias)
     })
   })
+
+  describe('creating receivers', () => {
+    let account
+
+    before(() => expect(testHelpers.ensureConfigured()).to.be.fulfilled
+      .then(() => {
+        return testHelpers.createAccount('receiver-test')
```
dominic: You use implicit return at the end of this `then` chain; let's stick with that for one liners.

dashboard/test/specs/accounts.js
```diff
@@ -31,13 +31,32 @@ describe('accounts', () => {
...
       browser.click('.FormContainer button')
       browser.waitForVisible('.AccountShow')
       browser.getText('.AccountShow').should.contain('Created account. Create another?')
       browser.getText('.AccountShow').should.contain(alias)
     })
   })
+
+  describe('creating receivers', () => {
+    before(() => expect(testHelpers.ensureConfigured()).to.be.fulfilled
+      .then(() => testHelpers.createAccount('receiver-test'))
```
dominic: This will fail on a second run - can we add a UUID?

PR #1445, opened 2017-07-17
cmd/corectl: remove hsm-url, hsm-token flags

jbowens:
Remove the hsm-url and hsm-token flags. These flags have been replaced
by a new configuration option that is modified through:

```
corectl add enclave <url> <token>
corectl rm enclave <url> <token>
```


kr:
LGTM

PR #1446, opened 2017-07-17
core/config: add single-tuple config options

jbowens:
Add corectl set command.

As-is, GetFunc provides no distinction between values that are not set and values that are set to the empty byte slice. I'm unsure if we have any need for that distinction or how we should expose it if we do.

Also, I stored the tuples in the same format as sets (a ValueSet protobuf), so that all config options can be accessed through List regardless of their type.

### DIFF
--- a/cmd/corectl/main.go
+++ b/cmd/corectl/main.go
@@ -68,6 +68,8 @@ var commands = map[string]*command{
 	"get":                  {get},
 	"add":                  {add},
 	"rm":                   {rm},
+	"set":                  {set},
+	"unset":                {unset},
 	"wait":                 {wait},
 }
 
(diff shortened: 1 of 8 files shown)


cmd/corectl/main.go
```diff
@@ -68,6 +68,8 @@ var commands = map[string]*command{
 	"get":                  {get},
 	"add":                  {add},
 	"rm":                   {rm},
+	"set":                  {set},
+	"unset":                {unset},
```
kr: Would it be possible (and would it be a good idea) to use `rm` for this as well? Maybe even all the way in to the Remove handler?

core/config/options.go
```diff
@@ -142,6 +152,56 @@ func (opts *Options) ListFunc(key string) func() [][]string {
...
 
+// GetFunc returns a closure that returns the tuple value
+// for the provided key.
+//
+// The configuration option for key must be defined as a
+// single tuple. GetFunc panics if the provided key is
+// undefined or is defined as a set of tuples.
+//
+// The returned function performs a stale read of the configuration
+// value. If an error occurs while reading the value the old
```
kr: Nit: I think a comma before "the old" would help readability.

cmd/corectl/main.go
jbowens: Good q. Would it take the current value as an argument or just be `corectl rm <key>`? I don't really have an opinion either way in unset vs rm.

kr: Oh hm, I guess for sets it needs to take the tuple value to remove. We could start by requiring it no matter what and if it becomes annoying, we can relax that for single-tuple options.

core/config/options.go
```diff
@@ -307,17 +306,25 @@ func (opts *Options) Set(key string, tup []string) sinkdb.Op {
...
 	}
-	if opt.tupleSize != len(tup) {
+	if opt.set && opt.tupleSize != len(tup) {
 		return sinkdb.Error(errors.WithDetailf(ErrConfigOp, "Configuration option %q expects %d arguments.", key, opt.tupleSize))
 	}
 
+	// If opt is defined to take a single value, just remove
+	// the key altogether.
+	if !opt.set {
+		return sinkdb.Delete(path.Join(sinkdbPrefix, key))
```
kr: Should this include an IfNotModified to protect against a concurrent update? I am not sure.

PR #1447, opened 2017-07-17
core/config: make migrateAccessTokens atomic in raft

vickiniu:
Currently, we migrate each access token from postgres to raft one-by-one. This change migrates all access tokens from postgres to raft atomically, so we don't fail halfway through a postgres->raft migration and lose the second half of the migration

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -368,8 +368,10 @@ func migrateAccessTokens(ctx context.Context, db pg.DB, sdb *sinkdb.DB) error {
 			networkGrants = append(networkGrants, &grant)
 		}
 	}
-	err = sdb.Exec(ctx, store.SaveAll(ctx, clientGrants, "client-readwrite"))
-	err = sdb.Exec(ctx, store.SaveAll(ctx, networkGrants, "crosscore"))
+	err = sdb.Exec(ctx,
+		store.Save(ctx, clientGrants...),
+		store.Save(ctx, networkGrants...),
+	)
 	if err != nil {
 		return errors.Wrap(err)
 	}
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -344,6 +344,8 @@ func migrateAccessTokens(ctx context.Context, db pg.DB, sdb *sinkdb.DB) error {
 		tokens = append(tokens, t)
 	})
 
+	var stores []sinkdb.Op
+
 	for _, token := range tokens {
 		data := map[string]interface{}{
 			"id": token.ID,
@@ -364,10 +366,11 @@ func migrateAccessTokens(ctx context.Context, db pg.DB, sdb *sinkdb.DB) error {
 		case "network":
 			grant.Policy = "crosscore"
 		}
-		err = sdb.Exec(ctx, store.Save(ctx, &grant))
-		if err != nil {
-			return errors.Wrap(err)
-		}
+		stores = append(stores, store.Save(ctx, &grant))
+	}
+	err = sdb.Exec(ctx, stores...)
+	if err != nil {
+		return errors.Wrap(err)
 	}
 	return err
 }
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -329,8 +329,6 @@ func tryGenerator(ctx context.Context, url, accessToken, blockchainID string, ht
 	return nil
 }
 
-// TODO(tessr): make all of this atomic in raft, so we don't get halfway through
-// a postgres->raft migration and fail, losing the second half of the migration
 func migrateAccessTokens(ctx context.Context, db pg.DB, sdb *sinkdb.DB) error {
 	store := authz.NewStore(sdb, GrantPrefix)
 	const q = `SELECT id, type, created FROM access_tokens`
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -342,7 +342,8 @@ func migrateAccessTokens(ctx context.Context, db pg.DB, sdb *sinkdb.DB) error {
 		tokens = append(tokens, t)
 	})
 
-	var stores []sinkdb.Op
+	var clientGrants []*authz.Grant
+	var networkGrants []*authz.Grant
 
 	for _, token := range tokens {
 		data := map[string]interface{}{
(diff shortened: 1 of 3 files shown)


core/config/config.go
```diff
@@ -364,10 +364,11 @@ func migrateAccessTokens(ctx context.Context, db pg.DB, sdb *sinkdb.DB) error {
 		case "network":
 			grant.Policy = "crosscore"
 		}
-		err = sdb.Exec(ctx, store.Save(ctx, &grant))
-		if err != nil {
-			return errors.Wrap(err)
-		}
+		stores = append(stores, store.Save(ctx, &grant))
```
kr: Could this produce multiple writes to the same key in this list? If so, that would be an [error](https://github.com/chain/chain/blob/ce5df096c4416342a5f5a31e9a3d2d4ad43cbaec/database/sinkdb/sinkdb.go#L76-L87). Should prob add a test case for that scenario.

vickiniu: Looks like this does produce multiple writes to the same key if we have multiple access tokens with the same `Policy` (since our key is `/core/grant` + `Policy`), and right now that throws an error (if, for example, we try to migrate two different client access tokens)

vickiniu: I'm not sure how we want to handle that: in the non-atomic migration world, it seems like all but one access token with the same key would've just been overwritten. We can do that overwrite inside `migrateAccessTokens` before we `Exec`, or we can keep the error as-is? Otherwise, if actually want to be able to store multiple `client` or `network` tokens in sinkDB, we may need to change the key that we store them under now

net/http/authz/store.go
```diff
@@ -75,6 +75,48 @@ func (s *Store) Save(ctx context.Context, g *Grant) sinkdb.Op {
 	)
 }
 
+// SaveAll returns an Op to store all the grants
+// Duplicates are ignored
+// It also sets field CreatedAt to the time the grants
+// are stored, or the time the original grant was stored
+func (s *Store) SaveAll(ctx context.Context, grants []*Grant, policy string) sinkdb.Op {
```
kr: Would it be reasonable to extend Save to accept multiple grants?

```go
func (s *Store) Save(ctx context.Context, g ...*Grant) sinkdb.Op
```

I think the implementation would be nearly identical to this SaveAll method (including requiring all provided grants to be in the same policy), but then we wouldn't also need the old Save method in addition. What do you think?

net/http/authz/store.go
```diff
@@ -75,6 +75,48 @@ func (s *Store) Save(ctx context.Context, g *Grant) sinkdb.Op {
...
+	key := s.keyPrefix + policy
+	var grantList GrantList
+	ver, err := s.sdb.Get(ctx, key, &grantList)
+	if err != nil {
+		return sinkdb.Error(errors.Wrap(err))
+	}
+	var newGrants []*Grant
+	for _, g := range grants {
+		if g.Policy != policy {
+			continue
```
kr: I think this should be treated as an error instead of silently discarding the grant.

net/http/authz/store.go
```diff
@@ -75,6 +75,48 @@ func (s *Store) Save(ctx context.Context, g *Grant) sinkdb.Op {
...
+// SaveAll returns an Op to store all the grants
+// Duplicates are ignored
+// It also sets field CreatedAt to the time the grants
+// are stored, or the time the original grant was stored
+func (s *Store) SaveAll(ctx context.Context, grants []*Grant, policy string) sinkdb.Op {
+	if len(grants) == 0 {
+		return sinkdb.Op{}
+	}
+	key := s.keyPrefix + policy
+	var grantList GrantList
```
kr: Maybe consider naming this something like `existing`? Then some of the comparisons below would be a little easier to read.

net/http/authz/store.go
```diff
@@ -75,6 +75,48 @@ func (s *Store) Save(ctx context.Context, g *Grant) sinkdb.Op {
...
+			if EqualGrants(*existing, *g) {
+				include = false
+			}
+		}
+		if include {
+			newGrants = append(newGrants, g)
+		}
+	}
+
+	existingGrants := grantList.Grants
```
kr: If `grantList` is renamed to `existing` as suggested above, maybe this variable becomes unnecessary?

core/config/config.go
```diff
@@ -361,12 +362,14 @@ func migrateAccessTokens(ctx context.Context, db pg.DB, sdb *sinkdb.DB) error {
...
 			grant.Policy = "client-readwrite"
+			clientGrants = append(clientGrants, &grant)
 		case "network":
 			grant.Policy = "crosscore"
+			networkGrants = append(networkGrants, &grant)
 		}
-		stores = append(stores, store.Save(ctx, &grant))
 	}
-	err = sdb.Exec(ctx, stores...)
+	err = sdb.Exec(ctx, store.SaveAll(ctx, clientGrants, "client-readwrite"))
```
kr: It looks the error is being ignored here, but the fix is easy:

```go
err = sdb.Exec(ctx,
	store.SaveAll(ctx, clientGrants, "client-readwrite"),
	store.SaveAll(ctx, networkGrants, "crosscore"),
)
```

kr: (That fix also has a small benefit of using only one raft operation instead of two. Each raft operation uses a couple of network round trips.)

net/http/authz/store.go
```diff
@@ -40,38 +40,45 @@ func (s *Store) Load(ctx context.Context, policy []string) ([]*Grant, error) {
...
-	grants := grantList.Grants
-	for _, existing := range grants {
-		if EqualGrants(*existing, *g) {
-			// this grant already exists, do nothing
-			g.CreatedAt = existing.CreatedAt
-			return sinkdb.Op{}
+	var newGrants []*Grant
+	for _, grant := range g {
+		if grant.Policy != policy {
+			return sinkdb.Error(errors.New("Grants have mismatching policies"))
```
kr: Nit: [error strings should not be capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

kr:
LGTM (after one small nitpick)

net/http/authz/store.go
vickiniu: Thanks! Is there a preference for using `fmt.Errorf` over `errors.New` ?

kr: No preference, just, if you need to format the string then `fmt.Errorf(...)` is more convenient than `errors.New(fmt.Sprintf(...))`.

PR #1450, opened 2017-07-18
core/fetch: encapsulate state

jbowens:
Remove package-level globals, encapsulate all state in a Replicator
struct.

### DIFF
--- a/core/fetch/fetch.go
+++ b/core/fetch/fetch.go
@@ -91,9 +91,9 @@ func (rep *Replicator) Fetch(ctx context.Context, c *protocol.Chain, health func
 	}
 }
 
-// PollGeneratorHeight periodically polls the configured peer for
+// PollRemoteHeight periodically polls the configured peer for
 // its blockchain height. It blocks until the ctx is canceled.
-func (rep *Replicator) PollGeneratorHeight(ctx context.Context) {
+func (rep *Replicator) PollRemoteHeight(ctx context.Context) {
 	rep.updateGeneratorHeight(ctx)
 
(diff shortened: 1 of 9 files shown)


core/fetch/fetch.go
```diff
@@ -19,24 +19,37 @@ import (
...
+	rep := &Replicator{peer: peer}
 
-func GeneratorHeight() (uint64, time.Time) {
-	generatorLock.Lock()
-	h := generatorHeight
-	t := generatorHeightFetchedAt
-	generatorLock.Unlock()
-	return h, t
+	// Fetch the generator height periodically.
+	go rep.pollGeneratorHeight(ctx)
```
kr: Maybe export pollGeneratorHeight so core.Run can call it directly? Then we wouldn't need the weird core.API plumbing to get ctx in here.

Or maybe just use context.TODO on this line and deal with it later?

core/run.go
```diff
@@ -177,6 +178,7 @@ func Run(
 	indexer := query.NewIndexer(db, c, pinStore)
 
 	a := &API{
+		ctx:          ctx,
```
kr: I'd really rather avoid this. I see it's just for plumbing to get the context in to fetch.New, but it would be too easy to reuse this field elsewhere. There's a suggestion above for how we might avoid this, but if none of those are viable, at least can we set it to nil after applying the options?

core/fetch/fetch.go
jbowens: Yeah, that's totally reasonable. I was also considering passing the ctx into all RunOptions?

kr:
LGTM

PR #1453, opened 2017-07-18
dashboard: add test for filtering assets and accounts

jeffomatic:
Also fixes list-item tests to make them less brittle.

### DIFF
--- a/dashboard/test/specs/accounts.js
+++ b/dashboard/test/specs/accounts.js
@@ -1,4 +1,5 @@
 const uuid = require('uuid')
+    , assert = require('assert')
 
 describe('accounts', () => {
   describe('list view', () => {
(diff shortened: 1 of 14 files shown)


dashboard/test/specs/accounts.js
```diff
@@ -74,4 +74,31 @@ describe('accounts', () => {
...
+        , id1 = uuid.v4()
+        , id2 = uuid.v4()
+        , id3 = uuid.v4()
+
+    before(() => {
+      expect(testHelpers.ensureConfigured()).to.be.fulfilled
+      .then(() => testHelpers.createAccount(id1, {tags: {x: tag1}}))
+      .then(() => testHelpers.createAccount(id2, {tags: {x: tag2}}))
+      .then(() => testHelpers.createAccount(id3, {tags: {x: tag1}}))
+      .then((account) => browser.url('/accounts'))
```
chrisgarvin: do we need the `account` param here?

jeffomatic: nope nice catch!

PR #1454, opened 2017-07-18
cmd/mdfmt: add command mdfmt, for formatting Markdown files

bobg:


bobg:
PTAL

ryandotsmith:
Nice!

cmd/mdfmt/mdfmt.go
```diff
@@ -0,0 +1,51 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+	"unicode"
+)
+
+func main() {
+	heading := 0 // current heading level
+	blank := 0   // number of pending blank lines
```
kr: Nit: looks like this could be a bool if you like.

PR #1455, opened 2017-07-18
dashboard: add tutorial integration test

chrisgarvin:


dashboard/src/features/tutorial/components/TutorialHeader/TutorialHeader.jsx
```diff
@@ -14,7 +15,7 @@ class TutorialHeader extends React.Component {
     } else {
       const collapsed = !this.props.showTutorial || this.props.currentStep.component == 'TutorialForm'
       return (
-        <div className={styles.container}>
+        <div className={componentClassNames(this, styles.container)}>
```
jeffomatic: I think we should try to find a way to add this to both side of the `if`/`else` clause. Or, we treat the `else` clause as a logically separate component.

PR #1456, opened 2017-07-19
all: make unnecessary vars const

jbowens:
Move some global vars to global consts. I found these
while auditing the code base for global state that might
prevent us from running multiple core.APIs in the same
process.


kr:
LGTM

PR #1457, opened 2017-07-19
errors: lazily load stack trace info

kr:
The standard library now provides a more convenient and robust way to get stack frame information, that correctly handles panic frames and non-Go (e.g. cgo) frames. At the same time, we can save some allocations and CPU time while capturing a stack trace by storing just the "callers" slice of PC values as returned by runtime.Callers.


kr:
PTAL

jbowens:
👀

jbowens:
lgtm

PR #1458, opened 2017-07-19
dashboard: test access control and transaction forms

dominic:


dashboard/test/specs/accessControl.js
```diff
@@ -12,7 +17,14 @@ describe('access control', () => {
     it('displays the correct page', () => {
       browser.getText('.PageTitle').should.contain('Access control')
       browser.getText('thead').should.contain('TOKEN NAME')
+    })
 
+    it('shows a new token form ', () => {
+      browser.waitForVisible('.AccessControlList .btn-primary')
```
jeffomatic: There's kind of an informal rule I have where we use style classnames for style and logical classnames for logic, and not mix the two. Is there a viable alternative to using `btn-primary` here? Like just `button` or something?

dominic: Let me see what I can change.

PR #1463, opened 2017-07-29
Update package.json to add history as a dependency.

religion-counter:
Update package.json to add history as a dependency.

Currently if you don't add history as a dependency the project is not working.


religion-counter:
When will I get a review for the change?

jeffomatic:
Hi, thanks for the PR. Can you describe what problem this solves? On our end, we don't have any issues running the dashboard project that are related to the `history` module`. Thanks.

PR #1471, opened 2017-11-28
Allow overriding AWS S3 Region with env var

harlow:
Previously the S3 connection was initialized as the global variable before `env.Parse()` was called. 

By moving the initialization into the main func we can leverage the env helper functions.

* Add s3 connection to blockCache
* Initialize s3 connection after env has been parsed

Addresses:

```
region = "us-east-1" // TODO(kr): figure out how to not hard code this
```

From commit: https://github.com/chain/chain/commit/aca1da719fc31f34df9a96c20ef3715a9e3493d4


harlow:
cc @kr lemme know your thoughts

PR #1472, opened 2017-11-29
core/account: don't count reserved-but-spent utxos

bobg:


bobg:
PTAL @jbowens

bobg:
Superseded by https://github.com/chain/chain/pull/1473

PR #1474, opened 2017-11-30
docs: remove Windows installer

croaky:
We're a small team, focused on making [Sequence](https://seq.com)
a great product over the last few months. We think Sequence is a going
to be a great upgrade path for many Chain Core Developer Edition users.

In the meantime, we're continuing to support Chain Core DE in the
[Community Slack account](https://chain.slack.com/) and elsewhere.

However, we feel we can no longer provide great support specifically for
the Windows installer for Chain Core DE.

This change removes the instructions for using that installer, leaving
Docker as the preferred installation method on Windows.

### DIFF
--- a/docs/1.0/core/get-started/install.partial.html
+++ b/docs/1.0/core/get-started/install.partial.html
@@ -44,19 +44,6 @@
 
 
         <div id="windows" class="tabcontent">
-          <h2>Download Windows installer</h2>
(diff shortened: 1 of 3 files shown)


tessr:
We should probably also remove the `desktop/windows` directory. I'm happy to do the honors there.

PR #1476, opened 2017-11-30
desktop/windows: remove windows installer

tessr:
This removes the Windows installer and a reference to it in the Github issue template.
It does not remove references to it in other documentation, because people may
still be running the app, which links back to that documentation (e.g. license files).


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,19 +1,19 @@
 Thank you for reporting an issue! Please read the following before submitting:
 
-Github issues are for reporting specific bugs with Chain Core Developer Edition. If you have general questions about Chain Core or questions regarding building an application on Chain Core, please visit our developer support center: 
+Github issues are for reporting specific bugs with Chain Core Developer Edition. If you have general questions about Chain Core or questions regarding building an application on Chain Core, please visit our developer support center:
 https://support.chain.com
 
-Enterprise customers can reach out to their support contact directly.
```
tessr: My editor automatically stripped out these whitespaces. Is that ok? I don't think it makes any difference

tessr:
I don't think we want to modify the change logs? But maybe we should?

croaky:
I think those two lines are currently broken links b/c this PR deletes the referenced `desktop/windows/CHANGELOG.md` files. https://github.com/chain/chain/blob/main/docs/1.2/core/reference/changelog.md

tessr:
Ah! Good catch. I'll delete them, then.

tessr:
Actually - those both point to the `1.2-stable` branch, and I was planning on landing this on `main`. But maybe I should land it on `1.2-stable`?

croaky:
Ah, another good catch. Reviewing how we deploy the docs https://github.com/chain/chain/tree/main/docs#uploading-the-docs it looks like we deploy from each of the `-stable` branches. So we should make these changes on `main`, `1.0-stable`, `1.1-stable`, and `1.2-stable`?

PR #1481, opened 2018-04-17
core: allow block period to be configured

croaky:
Some applications may want to change the block frequency
based on their traffic patterns.

For example, a value of 250 milliseconds (a factor of four)
improved performance in one application we observed.

This variable can set to any value that can be parsed by
https://golang.org/pkg/time/#ParseDuration such as `250ms`.


croaky:
Same change as https://github.com/chain/chain/pull/1480 (which was against `1.2-stable` but against `main`.

